PR #262, opened 2021-11-11
make bulk editing work with cross-database reference. 

ted-the-coder:
(from my question https://forum.jmix.io/t/bulkedit-doesnt-work-with-cross-datastore-reference/704/3)

there is an open issue: https://github.com/Haulmont/jmix-data/


Flaurite:
Close PR as it does not have changes.

PR #879, opened 2022-07-25
Update el translation

pbaris:
Add new keys in old files
Add translation for notifications and notificationsui


knstvk:
Thank you!

Issue #992, opened 2022-09-06
Cannot set UUID parameter into native PostgreSQL query


longirun:
Code below (calling line 63 `setParameter(3, progressId)`) results into exception.
```
    @Transactional
    public void updateSignedCount(UUID progressId, int updatedCount, int skippedCount) {
        entityManager.createNativeQuery("update master_processing_task_progress " +
                        "set version = version + 1, " +
                        "updated_date = now(), " +
                        "published_count = published_count + ?1, " +
                        "signing_skip_count = signing_skip_count + ?2 " +
                        "where ID = ?3")
                .setParameter(1, updatedCount)
                .setParameter(2, skippedCount)
                .setParameter(3, progressId)
                .executeUpdate();
    }

java.lang.RuntimeException: java.lang.ClassNotFoundException: com.haulmont.cuba.core.sys.persistence.PostgresUUID
	at io.jmix.core.common.util.ReflectionHelper.getClass(ReflectionHelper.java:82) ~[jmix-core-1.3.0.jar:na]
	at io.jmix.eclipselink.impl.JmixEclipseLinkQuery.internalSetParameter(JmixEclipseLinkQuery.java:872) ~[jmix-eclipselink-1.3.0.jar:na]
	at io.jmix.eclipselink.impl.JmixEclipseLinkQuery.setParameter(JmixEclipseLinkQuery.java:248) ~[jmix-eclipselink-1.3.0.jar:na]
	at io.jmix.eclipselink.impl.JmixEclipseLinkQuery.setParameter(JmixEclipseLinkQuery.java:61) ~[jmix-eclipselink-1.3.0.jar:na]
	at com.haulmont.ats.service.task.ProcessingTaskServiceBean.updateSignedCount(ProcessingTaskServiceBean.java:63) ~[main/:na]
```

PR #993, opened 2022-09-06
Cannot set UUID parameter into native postgres sql query/jmix#992

longirun:
Closes issue `Cannot set UUID parameter into native postgres sql query` https://github.com/jmix-framework/jmix/issues/992


knstvk:
Thanks!

PR #1143, opened 2022-10-28
Translate to Kurdish language (Sorani) ckb

swarakaka:


swarakaka:
Hi @knstvk 
Do you mean sql  Data Types, Statements and Operators? For example (INTEGER, DOUBLE)

knstvk:
Not exactly. I mean common elements like Create/Edit/Remove actions, Filter elements and so on. They are located in the UI module.
Because now when I enter a simple application with your localization, I still see most of the elements in English.

knstvk:
Ping me when the translation is ready. Please check that a new app with the translation applied looks good.

swarakaka:
Hello @knstvk 
i have completed the translations, please let me know if you notice any problems.

knstvk:
Thank you!
We'll publish an add-on with your translation after the next patch with updated BOM.

PR #1192, opened 2022-11-11
make multitenancy core module compatible with flowui

klaus7:
removed deprecated TenantManagementRole and dependency to security-ui to make multitenancy addon (without the UI module) compatible with flowui


KremnevDmitry:
The issue FlowUI for Multitenancy add-on #1207 has been fixed by commit e8582b9

PR #1251, opened 2022-11-30
Dutch translation

EzSoftware2015:
www.EZ-Software.be


knstvk:
Thank you!
We'll publish an add-on with your translation after the next patch with updated BOM.

PR #1278, opened 2022-12-11
Optimization of FetchPlanBuilder creation

daring2:
Related forum topic: https://forum.jmix.ru/t/medlennaya-zagruzka-sushhnostej-v-ekranah/734

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/FetchPlanBuilder.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/FetchPlanBuilder.java
@@ -32,7 +32,7 @@ import java.util.stream.Collectors;
  */
 public class FetchPlanBuilder {
 
-    protected FetchPlans fetchPlansFactory;
+    protected FetchPlans fetchPlans;
     protected Metadata metadata;
     protected MetadataTools metadataTools;
     protected FetchPlanRepository fetchPlanRepository;
@@ -41,7 +41,7 @@ public class FetchPlanBuilder {
     protected MetaClass metaClass;
     protected Set<String> properties = new LinkedHashSet<>();
     protected Map<String, FetchPlanBuilder> builders = new HashMap<>();
-    protected Map<String, FetchPlan> fetchPlans = new HashMap<>();
+    protected Map<String, FetchPlan> propertiesToFetchPlans = new HashMap<>();
     protected Map<String, FetchMode> fetchModes = new HashMap<>();
     protected boolean systemProperties = false;
     protected boolean loadPartialEntities = false;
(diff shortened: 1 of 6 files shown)


jmix-core/core/src/main/java/io/jmix/core/FetchPlans.java
```diff
@@ -27,13 +26,17 @@
...
+    protected MetadataTools metadataTools;
+    @Autowired
+    protected FetchPlanRepository fetchPlanRepository;
 
     /**
      * Returns {@link FetchPlan} builder for the given entity class.
      */
     public FetchPlanBuilder builder(Class<?> entityClass) {
-        return fetchPlanBuilderProvider.getObject(entityClass);
+        return new FetchPlanBuilder(this, entityClass);
```
knstvk, line 39: I would prefer passing all dependencies separately. Is it possible?

daring2: Sure, I updatred PR.

jmix-core/core/src/main/java/io/jmix/core/FetchPlanBuilder.java
```diff
@@ -36,17 +30,11 @@
...
  * Use {@link FetchPlans} factory to get the builder.
  */
-@Component("core_FetchPlanBuilder")
-@Scope(BeanDefinition.SCOPE_PROTOTYPE)
 public class FetchPlanBuilder {
 
-    @Autowired
-    protected ApplicationContext applicationContext;
-    @Autowired
+    protected FetchPlans fetchPlansFactory;
```
gorbunkov, line 35: Let's make the variable name just `fetchPlans`. In the framework code we try to give a variable the same name the type has unless there is a specific reason not to do it. And anyway, it is a factory for the FetchPlanBuilder, not for the FetchPlan

gorbunkov (changes requested):

jmix-core/core/src/main/java/io/jmix/core/FetchPlanBuilder.java
daring2: I agree with these considerations. I used a non-standard name because the `FetchPlanBuilder` class already contains `fetchPlans` variable that is used in many places. Should I rename the variable (to `fetchPlansMap`, for example) and use `fetchPlans` name for the dependency?

gorbunkov: Okay, I see.. Well, in that case I would go with renaming the collection. Let's make it `propertiesToFetchPlans`.

daring2: Done.

gorbunkov:
Thank you!

PR #1281, opened 2022-12-15
Dutch Translation ApplicationSettings Add-on

EzSoftware2015:


knstvk:
Please rollback changes in `gradle-properties`. The version should not be modified.

Issue #1325, opened 2023-01-06
Permit to configure CORS exposedHeaders from application.properties


Flamesson:
I'm developing a pet project right now. 

Backend - Jmix (v.1.4.1), frontend - React.

I've tested REST requests via Axis from frontend to backend and faced a problem - header responsible for total count (when &returnCount=true in the request url) is missed by default.

I think it should be mentioned that I have used an url with permitted anonymous access without authorization.

Request url was http://192.168.1.84:8080/rest/entities/Product?offset=0&limit=4&returnCount=true. I got such response:
``` json
{
  "data": [
    {
      "_entityName": "Product",
      "_instanceName": "",
      "id": "4c2b1da2-35ce-4948-b49a-bf1dec7fbed3",
      "leftover": 7,
      "version": 1,
      "price": 120,
      "name": " Шу пуэр №2 50 гр."
    },
    {
      "_entityName": "Product",
      "_instanceName": "",
      "id": "9d742227-c6fe-4c39-8a1e-0c0055e27c32",
      "leftover": 4,
      "version": 1,
      "price": 60,
      "name": " Чай зеленый кленовый лист 25 шт."
    },
    {
      "_entityName": "Product",
      "_instanceName": "",
      "id": "a242b9ef-1729-4298-8a46-7f54eafdba96",
      "leftover": 10,
      "version": 1,
      "price": 40,
      "name": " Чай черный 25 шт."
    },
    {
      "_entityName": "Product",
      "_instanceName": "",
      "id": "a72ce45c-2c84-4e27-99a5-290ebb2a594e",
      "leftover": 2,
      "version": 1,
      "price": 50,
      "name": " Чай черный с грушей 25 шт."
    }
  ],
  "status": 200,
  "statusText": "",
  "headers": {
    "cache-control": "no-cache, no-store, max-age=0, must-revalidate",
    "content-length": "714",
    "content-type": "application/json;charset=UTF-8",
    "expires": "0",
    "pragma": "no-cache"
  },
  "config": {
    "transitional": {
      "silentJSONParsing": true,
      "forcedJSONParsing": true,
      "clarifyTimeoutError": false
    },
    "adapter": [
      "xhr",
      "http"
    ],
    "transformRequest": [
      null
    ],
    "transformResponse": [
      null
    ],
    "timeout": 0,
    "xsrfCookieName": "XSRF-TOKEN",
    "xsrfHeaderName": "X-XSRF-TOKEN",
    "maxContentLength": -1,
[...]

PR #1326, opened 2023-01-06
#1325 Permit to configure CORS exposedHeaders from application.properties

Flamesson:
Add exposed headers to cors configuring mechanism

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/CorsProperties.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/CorsProperties.java
@@ -56,6 +56,18 @@ public class CorsProperties {
     List<String> allowedMethods;
 
     /**
+     * <p>A list of response headers other than simple headers (i.e. Cache-Control, Content-Language, Content-Type,
+     *     Expires, Last-Modified, or Pragma) that an actual response might have and can be exposed.</p>
+     *
+     * <p>The special value "*" allows all headers to be exposed for non-credentialed requests.</p>
+     *
+     * <p>By default this is not set.</p>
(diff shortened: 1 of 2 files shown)


jmix-core/core/src/main/java/io/jmix/core/CorsProperties.java
```diff
@@ -55,6 +55,18 @@ public class CorsProperties {
      */
     List<String> allowedMethods;
 
+    /**
+     * <p>A list of response headers other than simple headers (i.e. Cache-Control, Content-Language, Content-Type,
```
gorbunkov, line 59: This javadoc comment will go to Spring boot [configuration metadata](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html) and will be displayed in IntelliJ Idea Ultimate as a hint, when you start entering the application property name in your applicaition. Please remove the leading <p> tag to make this hint look good.

gorbunkov (changes requested):

jmix-core/core/src/main/java/io/jmix/core/CorsProperties.java
Flamesson: As far as I understand, you meant this

Screenshot 2023-12-03 at 13 57 31

Updated javadoc. Now:

Screenshot 2023-12-03 at 14 00 25

Screenshot 2023-12-03 at 13 58 54

gorbunkov: Exactly) Thank you

PR #1329, opened 2023-01-08
Add russian translation for Quartz add-on

daring2:


knstvk:
Thanks for your contribution!
We've refactored the Quartz add-on structure recently, and the message keys have been changed, see https://github.com/jmix-framework/jmix/commit/fbc9293577ea212aeacaa442aff5e6704b23125f
Please update the PR accordingly.

daring2:
I updated message keys. 

The mentioned commit contains a Russian translation for the add-on screen messages. It uses the caption "Задача" for the Job entity that differs from my variant ("Задание"). I've updated the translation to use my version because it is compatible with the CUBA platform translation (https://doc.cuba-platform.com/manual-latest-ru/scheduled_tasks_cuba.html) and users are probably used to it.

daring2:
ОК, I'll create the PR for release_1_5 a bit later.

Issue #1460, opened 2023-02-27
Support for JUnit5 @Nested test classes for JmixUiTestExtension 


mariodavid:
Currently, when annotating a UI integration test with `@UiTest` on a test class, when using `@Nested` inner classes to group together different test cases, the annotation is not correctly interpreted. E.g. the authenticated user is "system" in the inner class although the outer class defines "admin".

Example:

```
@UiTest(authenticatedUser = "admin")
@SpringBootTest
public class AuthenticatedUserTest  {

    @Autowired
    CurrentAuthentication currentAuthentication;

    @Test
    void currentAuthentication_isAdmin() {

        User user = (User) currentAuthentication.getUser();

        assertThat(user.getLogin())
                .isEqualTo("admin");
    }

    @Nested
    class NestedJunit5Class {

        @Test
        void currentAuthentication_isAdmin() {

            User user = (User) currentAuthentication.getUser();

            assertThat(user.getLogin())
                    .isEqualTo("admin");
        }
    }
}
```

The first test case succeeds, the second one fails.

One option is to annotate all inner classes with the `@UiTest` annotation, but this is quite cumbersome, when the inner test are just used as groups of tests that don't have a different setting for the annotation.

Flaurite:
Changes are made from PR https://github.com/jmix-framework/jmix/pull/1463 with some additions.

PR #1463, opened 2023-02-27
Support for nested Junit5 test classes for UiTest annotation

mariodavid:


Flaurite:
Hello @mariodavid!

Just to clarify, have you considered cases when you need to override annotation values on a base class from a nested class?

Flaurite:
Hello @mariodavid,
since master doesn't have classic UI modules PR cannot be merged to master. And unfortunately this branch has some changes (as it based on old master branch) that should not go to release 1.5/1.6.

Almost all changes from this PR and some additions I'll push in this issue: https://github.com/jmix-framework/jmix/issues/1460

Issue #1389, opened 2023-01-25
Bulk Editor cannot save entities when using with multitenancy add-on


mariodavid:
Currently it is not possible to use the bulk editor functionality when the multitenancy add-on is used. Here is an example case:

In the Bookstore example on the product-browse screen I added a bulkEdit button. The data loading of the browser contains the following attributes:
 
```
<data readOnly="true">
    <collection id="productsDc"
                class="io.jmix.bookstore.product.Product">
        <fetchPlan extends="_base">
            <property name="category" fetchPlan="_base"/>
            <property name="supplier" fetchPlan="_base"/>
            <property name="tenant"/>
        </fetchPlan>
        <loader id="productsDl">
            <query>
                <![CDATA[select e from bookstore_Product e]]>
            </query>
        </loader>
    </collection>
</data>
```

The tenant attribute is actually included in "_base", but for testing I explicitly added it. The tenant attribute is defined in the StandardTenantEntity: 

```
@JmixEntity
@MappedSuperclass
public class StandardTenantEntity extends StandardEntity {
    @SystemLevel
    @TenantId
    @Column(name = "TENANT")
    private String tenant;
}
```

```
<action id="bulkEdit" type="bulkEdit">
    <properties>
        <property name="includeProperties" value="unitsInStock,unitsOnOrder"/>
    </properties>
</action>
```

When trying to save the bulk editor, the following error occurs:

```
Caused by: java.lang.IllegalStateException: Cannot get unfetched attribute [tenant] from detached object io.jmix.bookstore.product.Product-075aa739-3324-66a5-d105-e15301124d6b [detached].
```

<img width="2032" alt="CleanShot 2023-01-24 at 09 53 54@2x" src="https://user-images.githubusercontent.com/817400/214495934-7aac5242-2809-457f-a789-8fad7e6ba23c.png">

PR #1469, opened 2023-02-28
Bulk Editor cannot save entities when using with multitenancy add-on …

KremnevDmitry:
Source: https://github.com/jmix-framework/jmix/issues/1389

### DIFF
--- a/jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
+++ b/jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
@@ -16,6 +16,9 @@
 
 package io.jmix.multitenancy.core.impl;
 
+import io.jmix.core.DataManager;
+import io.jmix.core.FetchPlans;
+import io.jmix.core.Id;
 import io.jmix.core.Metadata;
 import io.jmix.core.annotation.TenantId;
 import io.jmix.core.entity.EntityValues;
@@ -40,9 +43,11 @@ public class TenantEntityOperationImpl implements TenantEntityOperation {
     private static final Logger log = LoggerFactory.getLogger(TenantEntityOperationImpl.class);
 
     private final Metadata metadata;
+    private final DataManager dataManager;
 
-    public TenantEntityOperationImpl(Metadata metadata) {
+    public TenantEntityOperationImpl(Metadata metadata, DataManager dataManager, FetchPlans fetchPlans) {
         this.metadata = metadata;
+        this.dataManager = dataManager;
     }
 
     /**
@@ -79,9 +84,11 @@ public class TenantEntityOperationImpl implements TenantEntityOperation {
     public void setTenant(Object entity, String tenantId) {
         MetaProperty property = findTenantProperty(entity.getClass());
         if (property != null) {
-            EntityValues.setValue(entity, property.getName(), tenantId);
+            Object reloadedEntity = dataManager.load(Id.of(entity))
+                    .fetchPlanProperties(property.getName())
+                    .one();
+
+            EntityValues.setValue(reloadedEntity, property.getName(), tenantId);
         }
     }
-
-
 }


jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
```diff
@@ -79,9 +84,11 @@ private Field findTenantField(Class<?> entityClass) {
     public void setTenant(Object entity, String tenantId) {
         MetaProperty property = findTenantProperty(entity.getClass());
         if (property != null) {
-            EntityValues.setValue(entity, property.getName(), tenantId);
+            Object reloadedEntity = dataManager.load(Id.of(entity))
```
gorbunkov, line 87: What idea is behind this? Why do we need to reload an instance, modify some field in this **reloaded** instance and then do nothing with this reloaded instance?

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
```diff
@@ -40,9 +43,11 @@ public class TenantEntityOperationImpl implements TenantEntityOperation {
     private static final Logger log = LoggerFactory.getLogger(TenantEntityOperationImpl.class);
 
     private final Metadata metadata;
+    private final DataManager dataManager;
 
-    public TenantEntityOperationImpl(Metadata metadata) {
+    public TenantEntityOperationImpl(Metadata metadata, DataManager dataManager, FetchPlans fetchPlans) {
```
gorbunkov, line 48: You don't use FetchPlans bean anywhere.

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
```diff
@@ -79,9 +84,11 @@ private Field findTenantField(Class<?> entityClass) {
     public void setTenant(Object entity, String tenantId) {
         MetaProperty property = findTenantProperty(entity.getClass());
         if (property != null) {
-            EntityValues.setValue(entity, property.getName(), tenantId);
+            Object reloadedEntity = dataManager.load(Id.of(entity))
+                    .fetchPlanProperties(property.getName())
+                    .one();
```
gorbunkov, line 89: Anyway, this code will probably fail when you try to create a new tenant instance in a regular way (as it is described in the documentation)

gorbunkov (changes requested):

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantEntityOperationImpl.java
KremnevDmitry: I chose the wrong way to solve this problem. 
Now, the tenant property will be added to the FetchPlan of the persisted object, if possible.

jmix-ui/ui/src/main/java/io/jmix/ui/app/bulk/BulkEditorWindow.java
```diff
@@ -412,7 +413,8 @@ protected FetchPlan createFetchPlan(MetaClass metaClass) {
             String fqn = generateFqn(metaProperty, null);
 
             if (!managedFields.containsKey(fqn)
-                    && !managedEmbeddedProperties.contains(fqn)) {
+                    && !managedEmbeddedProperties.contains(fqn)
+                    && isNotTenantMetaProperty(metaProperty)) {
```
gorbunkov, line 417: For my taste, it's better when methods don't contain negation in themselves (until this is explicitly required). Maybe we should modify this to `!isTenantMetaProperty(...)`?

jmix-ui/ui/src/main/java/io/jmix/ui/app/bulk/BulkEditorWindow.java
```diff
@@ -777,6 +779,10 @@ protected void applyModelValidators(ValidationErrors errors) {
         }
     }
 
+    public boolean isNotTenantMetaProperty(MetaProperty metaProperty) {
```
gorbunkov, line 782: There is no need to make this method public

gorbunkov (changes requested):

gorbunkov (approved):
Please, DON'T merge this into master and release_1_5 until release 1.5.0 is built. It's planned for Friday. This issue will be included into 1.5.1

Issue #968, opened 2022-08-23
BigDecimal field values must be rounded according to field format


gorbunkov:
Currently if you enter the `123.4567` value to the text field with a decimal type, the displayed value will become `123.46`, i.e. it will be rounded. However the model (edited entity attribute) will contain the value that you actually entered (123.4567). So there is a case when you see one value in the UI component and another value is actually stored in the model.

Probably this "rounding" behavior should be implemented inside the datatype and this behavior should be switched off by the feature toggle if necessary.

PR #1516, opened 2023-03-20
BigDecimal field values must be rounded according to field format #968

Gavrilov-Ivan:
https://github.com/jmix-framework/jmix/issues/968
Pay close attention to property - is it ok to put it in 'jmix.core'?

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
@@ -99,9 +99,10 @@ public class CoreProperties {
     PessimisticLock pessimisticLock;
 
     /**
-     * Whether BigDecimalDatatype should round the actual parsed value according to format settings.
+     * Whether BigDecimalDatatype, DoubleDatatype and FloatDatatype should round the actual parsed value
+     * according to format settings.
      */
-    boolean bigDecimalValueRoundByFormat;
+    boolean decimalValueRoundByFormat;
 
     public CoreProperties(
             String webHostName,
@@ -123,7 +124,7 @@ public class CoreProperties {
             @DefaultValue("true") boolean triggerFilesEnabled,
             @DefaultValue("5000") Duration triggerFilesProcessInterval,
             @DefaultValue PessimisticLock pessimisticLock,
-            @DefaultValue("true") boolean bigDecimalValueRoundByFormat) {
+            @DefaultValue("true") boolean decimalValueRoundByFormat) {
         this.webHostName = webHostName;
         this.webPort = webPort;
         this.confDir = confDir;
@@ -152,7 +153,7 @@ public class CoreProperties {
         this.triggerFilesEnabled = triggerFilesEnabled;
         this.triggerFilesProcessInterval = triggerFilesProcessInterval;
         this.pessimisticLock = pessimisticLock;
-        this.bigDecimalValueRoundByFormat = bigDecimalValueRoundByFormat;
+        this.decimalValueRoundByFormat = decimalValueRoundByFormat;
     }
 
     public String getWebHostName() {
@@ -288,9 +289,9 @@ public class CoreProperties {
     }
 
     /**
-     * @see #bigDecimalValueRoundByFormat
+     * @see #decimalValueRoundByFormat
      */
-    public boolean isBigDecimalValueRoundByFormat() {
-        return bigDecimalValueRoundByFormat;
+    public boolean isDecimalValueRoundByFormat() {
+        return decimalValueRoundByFormat;
     }
 }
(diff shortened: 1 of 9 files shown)


gorbunkov:
@Gavrilov-Ivan Did you check whether this problem is also relevant for Double or Float datatypes?

Gavrilov-Ivan:
Yes, it's relevant for Double and Float datatypes also.
Modified parse process of them the similar way:

1. Parse raw string value as BigDecimal
2. Perform rounding if necessary
3. Check if parsed value fits value-range of target datatype.
4. Get result value as a target datatype

Added tests.

jmix-core/core/src/test/groovy/properties/TestCoreProperties.java
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+package properties;
+
+import io.jmix.core.CoreProperties;
+
+import java.time.Duration;
+import java.util.List;
+
+public class TestCoreProperties extends CoreProperties {
```
gorbunkov, line 24: It seems that "groovy" directory is a wrong place for java class. `java/test_support/datatypes` seems to be a better choice.
Also, what is the purpose of the class? Is it for setting the `decimalValueRoundByFormat` property only? If so, then maybe we could add the `@TestPropertySource` annotations to test classes with explicit property value instead?

jmix-core/core/src/main/java/io/jmix/core/metamodel/datatype/impl/DoubleDatatype.java
```diff
@@ -82,10 +90,39 @@ public Double parse(String value, Locale locale) throws ParseException {
...
+    protected Number parse(String value, NumberFormat format) throws ParseException {
+        BigDecimal result = (BigDecimal) super.parse(value, format);
+        if (coreProperties.isDecimalValueRoundByFormat()) {
+            int maximumFractionDigits = format.getMaximumFractionDigits();
+            RoundingMode roundingMode = format.getRoundingMode();
+            result = result.setScale(maximumFractionDigits, roundingMode);
+        }
+
+        if (!isInDoubleRange(result)) {
+            throw new ParseException(String.format("Double range exceeded: \"%s\"", value), 0);
```
gorbunkov, line 107: I'm not sure that "exceeded" is a proper word if we speak about negative values. Maybe we can say something like "The value is out of Double datatype range"? The same for Float.

jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
```diff
@@ -100,6 +100,12 @@ public class CoreProperties {
      */
     PessimisticLock pessimisticLock;
 
+    /**
+     * Whether BigDecimalDatatype, DoubleDatatype and FloatDatatype should round the actual parsed value
+     * according to format settings.
+     */
+    boolean decimalValueRoundByFormat;
```
gorbunkov, line 107: Maybe rename the property to `roundDecimalValueByFormat`? WDYT?

gorbunkov (changes requested):

jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
Gavrilov-Ivan: Looks a little better - renamed

jmix-core/core/src/main/java/io/jmix/core/metamodel/datatype/impl/DoubleDatatype.java
Gavrilov-Ivan: Done

jmix-core/core/src/test/groovy/properties/TestCoreProperties.java
Gavrilov-Ivan: Moved class to `java/test_support`, without `datatypes` sub-package - this class can be used not only for `datatype` scope of tests.
This class helps to change property values in different tests within single test class. Adding a 'static' property source for entire class looks less suitable for this case. Don't want to split it into multiple test classes based on possible values of involved properties.

jmix-core/core/src/main/java/io/jmix/core/metamodel/datatype/impl/BigDecimalDatatype.java
```diff
@@ -94,7 +98,13 @@ public BigDecimal parse(String value, Locale locale) throws ParseException {
         DecimalFormatSymbols formatSymbols = formatStrings.getFormatSymbols();
         DecimalFormat format = new DecimalFormat(formatStrings.getDecimalFormat(), formatSymbols);
         format.setParseBigDecimal(true);
-        return (BigDecimal) parse(value, format);
+        BigDecimal result = (BigDecimal) parse(value, format);
+        if(coreProperties.isBigDecimalValueRoundByFormat()) {
```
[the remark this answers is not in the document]
Gavrilov-Ivan: Had been done earlier

gorbunkov (approved):

glebfox (approved):

Issue #1439, opened 2023-02-17
Exception when email template parameter is an entity with non-UUID identifier


gorbunkov:
Jmix version: 1.4.4

* Create an entity with `Long` id, e.g. `Customer`.
* Create an email template with the parameter of this entity type.
* Select this email template in the list. Click **Send** button. Select the `Customer` entity parameter and then click the **Preview** button.

AR: you'll get an exception:

```
Caused by: java.lang.IllegalArgumentException: Invalid UUID string: 1
	at java.base/java.util.UUID.fromString(UUID.java:215) ~[na:na]
	at io.jmix.reports.libintegration.JmixObjectToStringConverter.convertFromString(JmixObjectToStringConverter.java:69) ~[jmix-reports-1.4.4.jar:na]
	at io.jmix.emailtemplates.TemplateParametersExtractor.convertFromString(TemplateParametersExtractor.java:192) ~[jmix-emailtemplates-1.4.4.jar:na]
	at io.jmix.emailtemplates.TemplateParametersExtractor.getReportDefaultValues(TemplateParametersExtractor.java:115) ~[jmix-emailtemplates-1.4.4.jar:na]
	at io.jmix.emailtemplates.impl.EmailTemplatesImpl.getReportWithParams(EmailTemplatesImpl.java:122) ~[jmix-emailtemplates-1.4.4.jar:na]
	at io.jmix.emailtemplates.impl.EmailTemplatesImpl.generateEmail(EmailTemplatesImpl.java:94) ~[jmix-emailtemplates-1.4.4.jar:na]
	at io.jmix.emailtemplatesui.screen.emailtemplate.send.EmailTemplateSendScreen.getEmailInfo(EmailTemplateSendScreen.java:306) ~[jmix-emailtemplates-ui-1.4.4.jar:na]
	at io.jmix.emailtemplatesui.screen.emailtemplate.send.EmailTemplateSendScreen.onPreviewButtonClick(EmailTemplateSendScreen.java:273) ~[jmix-emailtemplates-ui-1.4.4.jar:na]
	at io.jmix.core.common.event.EventHub.publish(EventHub.java:170) ~[jmix-core-1.4.4.jar:na]
	at io.jmix.ui.component.impl.AbstractComponent.publish(AbstractComponent.java:85) ~[jmix-ui-1.4.4.jar:na]
[...]

PR #1521, opened 2023-03-21
Exception when email template parameter is an entity with non-UUID id entifier jmix-framework/jmix#1439

KremnevDmitry:


jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/JmixObjectToStringConverter.java
```diff
@@ -66,12 +68,28 @@ public Object convertFromString(Class parameterClass, @Nullable String paramValu
...
-            UUID id = UUID.fromString(paramValueStr);
-            Optional entityOpt = dataManager.load(parameterClass)
-                    .id(id)
-                    .fetchPlan(FetchPlan.BASE)
-                    .optional();
-            return entityOpt.isPresent() ? entityOpt.get() : null;
+            MetaProperty idProperty = metadataTools.getPrimaryKeyProperty(parameterClass);
+            Object idValue = null;
+
+            if (idProperty != null && idProperty.getRange().isDatatype()) {
```
gorbunkov, line 74: Please check what happens if primary key is not a Datatype, e.g. the entity has a composite PK. It seems that the method will just return null. We should either handle composite primary keys or raise an exception saying that primary key type is not supported.

gorbunkov (changes requested):

jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/JmixObjectToStringConverter.java
KremnevDmitry: Now an exception will be thrown.

jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/JmixObjectToStringConverter.java
```diff
@@ -66,12 +68,36 @@ public Object convertFromString(Class parameterClass, @Nullable String paramValu
...
+                if (idValue != null) {
+                    return dataManager.load(parameterClass)
+                            .id(Id.of(idValue, parameterClass))
+                            .fetchPlan(FetchPlan.BASE)
+                            .optional()
+                            .orElse(null);
+                }
+            } else if (idProperty.getRange().isClass()) {
+                throw new ReportingException(
+                        String.format("Unsupported composite id from [%s] with value [%s]",
```
gorbunkov, line 96: Let's rephrase: "composite id from" -> "composite primary key in"

gorbunkov (approved):

jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/JmixObjectToStringConverter.java
KremnevDmitry: Fixed

Issue #1456, opened 2023-02-22
The validation message does not disappear after the correct data is entered


rusiaikinat:
env
Jmix version: 1.5.999-SNAPSHOT
Jmix Studio plugin version: 1.5.SNAPSHOT4957-223
IntelliJ version: IntelliJ IDEA 2022.3.1 (Ultimate Edition)

TC
Enter string data in a non-string field
Change the entered data to the correct one

AR
The message about incorrectly entered data remains
![image](https://user-images.githubusercontent.com/114649487/220533775-c33b34d0-3ca5-4882-b02d-341685e3b858.png)

**Field type**: Float, BigInteger, BigDecimal, Short

ER
The message should disappear after entering the correct data

PR #1527, opened 2023-03-22
The validation message does not disappear after the correct data is entered

glebfox:
Fixes jmix-framework/jmix#1456


glebfox:
> Is it possible to write tests for this case?

Can't set unparseable value from server-side

Issue #1523, opened 2023-03-21
Jsessionid should change after login


glebfox:
Currently `JSESSIONID` is same before and after login.

PR #1528, opened 2023-03-23
Session fixation protection

glebfox:
Fixes: jmix-framework/jmix#1523

### DIFF
--- a/jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
+++ b/jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
@@ -188,7 +188,7 @@ public class LoginScreenSupport {
 
             HttpSession httpSession = session instanceof WrappedHttpSession ?
                     ((WrappedHttpSession) session).getHttpSession() : null;
-            log.info("Session reinitialized: HttpSession={}, timeout={}sec",
+            log.debug("Session reinitialized: HttpSession={}, timeout={}sec",
                     httpSession, timeout);
         }
     }
--- a/jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
+++ b/jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
@@ -47,7 +47,6 @@ import org.springframework.security.web.authentication.session.SessionAuthentica
 import org.springframework.stereotype.Component;
 
 import javax.annotation.Nullable;
-import javax.servlet.http.HttpSession;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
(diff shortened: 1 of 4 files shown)


jmix-ui/ui/src/main/java/io/jmix/ui/UiProperties.java
```diff
@@ -135,6 +135,11 @@ public class UiProperties {
      */
     int mainTabCaptionLength;
 
+    /**
+     * Reinitialize session after login to protect from Session Fixation attacks.
```
gorbunkov, line 139: "Whether to reinitialize a session..." - like for other boolean attributes. Also, why Session Fixation start with capital letters?

jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
```diff
@@ -169,11 +169,30 @@ public Authentication authenticate(AuthDetails authDetails,
...
+                && uiProperties.isUseSessionFixationProtection()) {
+            VaadinService.reinitializeSession(VaadinRequest.getCurrent());
+
+            WrappedSession session = VaadinSession.getCurrent().getSession();
+            int timeout = uiProperties.getHttpSessionExpirationTimeoutSec();
+            session.setMaxInactiveInterval(timeout);
+
+            HttpSession httpSession = session instanceof WrappedHttpSession ?
+                    ((WrappedHttpSession) session).getHttpSession() : null;
+            log.info("Session reinitialized: HttpSession={}, timeout={}sec",
```
gorbunkov, line 191: First, do we need this log message with "info" level? Right now we don't log the fact the user has been logged in, but log some "Session reinitialzed" message.
Another thing - if the session is not WrappedHttpSession we will log "Session reinitialized: HttpSession=null" - is this a desired behavior?

gorbunkov (changes requested):

jmix-ui/ui/src/main/java/io/jmix/ui/UiProperties.java
glebfox: Got inspired from CUBA implementation, will fix that

jmix-security/security-ui/src/main/java/io/jmix/securityui/authentication/LoginScreenSupport.java
glebfox: Regarding `HttpSession`, it appears that `WrappedSession`  has 2 implementations, both have different methods to obtain original session, and in case of `LoginScreenSupport` its unlikely that we get `WrappedPortletSession`, but `WrappedSession` has no redeable `toString()` implementation.

gorbunkov: As I see, `WrappedSession` has the `getId()` method. Isn't it what we want to see in logs?

Issue #1535, opened 2023-03-27
Tomcat cannot delete temp file after uploading it by REST add-on using multipart request with default multipart configuration


Flaurite:
### Description

Demo project: [upload.zip](https://github.com/jmix-framework/jmix/files/11108373/upload.zip)

Project contains REST add-on and uses configuration that excludes adding `CommonsMultipartResolver` provided by REST add-on.
If you try to upload file using multipart request (e.g. using Postman) you'll see an exception:
```
java.io.UncheckedIOException: Cannot delete C:\Users\some_user\AppData\Local\Temp\tomcat.8080.3286764258145899829\work\Tomcat\localhost\ROOT\upload_860f2115_9ae1_4fef_a7d5_1b3c5c804eac_00000000.tmp
	at org.apache.tomcat.util.http.fileupload.disk.DiskFileItem.delete(DiskFileItem.java:431) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.catalina.core.ApplicationPart.delete(ApplicationPart.java:54) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.catalina.connector.Request.recycle(Request.java:483) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.catalina.connector.CoyoteAdapter.log(CoyoteAdapter.java:515) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.catalina.connector.CoyoteAdapter.checkRecycled(CoyoteAdapter.java:539) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.coyote.http11.Http11Processor.recycle(Http11Processor.java:1439) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.coyote.AbstractProtocol$ConnectionHandler.release(AbstractProtocol.java:1098) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:1056) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1784) ~[tomcat-embed-core-9.0.71.jar:9.0.71]
[...]

PR #1538, opened 2023-03-27
Tomcat cannot delete temp file after uploading it by REST add-on using multipart request with default multipart configuration

Flaurite:
Fix for https://github.com/jmix-framework/jmix/issues/1535


Flaurite:
Merged in this commit: https://github.com/jmix-framework/jmix/commit/55faa3f62af4aaf7563ceb4f395d76122e34214f

PR #1542, opened 2023-03-27
Default test in add-on fails because of invalid Liquibase configuration 1533

gorbunkov:
The `test-app.properties` file with Liquibase config has been added to add-on templates for Java and Kotlin.


gorbunkov:
@knstvk are you ok with the `test-app.properties` file name in the template?

Issue #1277, opened 2022-12-09
Export/import of audit configuration


Flaurite:

rusiaikinat:
Allows you to create several settings with a zero value #1627

There is no exportation of more than one parameter #1629

PR #1567, opened 2023-04-04
Add feature audit entity config now can be imported/exported #1277

KartnDev:


jmix-audit/audit-flowui/src/main/java/io/jmix/auditflowui/view/entitylog/EntityLogView.java
```diff
@@ -166,17 +131,13 @@ public class EntityLogView extends StandardListView<EntityLogItem> {
     @ViewComponent
     protected Button cancelBtn;
     @ViewComponent
+    private FileUploadField importField;
```
gorbunkov, line 134: If all injected fields are protected,  let's make this one protected as well

gorbunkov (approved):

Issue #1612, opened 2023-04-19
Support Vaadin 24 for dev-server


Fedoseew:
Support Vaadin 24 for module jmix-flowui-devserver

Issue #1309, opened 2022-12-23
Add NPM installers in Flow UI dev server


Fedoseew:
### Description:
Add installers that will download and install all needed things for working with Flow UI from Jmix Studio.

For QA: 
1. Uninstall NodeJs and NPM from your machine, clear all caches.
2. Open Flow UI project in Jmix Studio.

ER: View Designer must be available.

<hr></hr>

### Versions info:
1. Default URL to download Node: https://nodejs.org/dist/.
> setting custom default download URL: io.jmix.flowui.devserver.frontend.Options#withNodeDownloadRoot
4. Default Node version: v18.14.1 (includes NPM). Will be installed if Node with supported version not found.
> see also: com.vaadin.flow.server.frontend.FrontendTools#DEFAULT_NODE_VERSION
5. Default NPM version: 9.3.1. Will be installed if NPM with supported version not found.
> see also: com.vaadin.flow.server.frontend.FrontendTools#DEFAULT_NPM_VERSION
6. Default PNPM version: 5.18.10. Will be installed if PNPM with supported version not found.
> see also: com.vaadin.flow.server.frontend.FrontendTools#DEFAULT_PNPM_VERSION
7. Supported Node version (min = 18.0):
> see also: io.jmix.flowui.devserver.frontend.FrontendTools#SUPPORTED_NODE_VERSION
8. Supported NPM versions (min = 8.6):
> see also: io.jmix.flowui.devserver.frontend.FrontendTools#SUPPORTED_NPM_VERSION
9. Supported PNPM versions (min = 5.0):
> see also com.vaadin.flow.server.frontend.FrontendTools#SUPPORTED_PNPM_VERSION

<hr></hr>

### Current mechanism for installing and build frontend dependencies:
1. Generate/Validate 'package.json' and other needed configuration files.
> see also: io.jmix.flowui.devserver.frontend.NodeTasks#execute
2. io.jmix.flowui.devserver.frontend.TaskRunNpmInstal task steps:
- Validating Node and NPM: 
> see also: io.jmix.flowui.devserver.frontend.FrontendTools#validateNodeAndNpmVersion
- Download and Install Node if needed (if cannot find Node executable or Node version  not supported):
> see also: io.jmix.flowui.devserver.frontend.FrontendTools#installNode
6. Run npm (pnpm) install command.
> see also: io.jmix.flowui.devserver.frontend.TaskRunNpmInstall#runNpmInstall

<hr></hr>

### More info:
[...]

PR #1635, opened 2023-04-25
DevServer Vaadin 24

Fedoseew:
Support Vaadin 24 for dev-server #1612
Add NPM installers in Flow UI dev server #1309


jmix-flowui/flowui-kit/flowui-kit.gradle
```diff
@@ -26,4 +26,4 @@ dependencies {
     api("com.vaadin:vaadin-core:$vaadinFlowVersion") {
```
gorbunkov, line 26: Why do we need an explicit version here? Won't it be taken from Jmix BOM?

Maybe just `api("com.vaadin:vaadin-core")` can work and that we can remove vaadinFlowVersion from settings gradle?

jmix-flowui/flowui-devserver/flowui-devserver.gradle
```diff
@@ -29,13 +29,11 @@ archivesBaseName = 'jmix-flowui-devserver'
 dependencies {
     api(project(":flowui-kit"))
 
-    api("org.eclipse.jetty:jetty-continuation:$jettyVersion")
     api("org.eclipse.jetty:jetty-webapp:$jettyVersion")
-    api("org.eclipse.jetty.websocket:websocket-server:$jettyVersion")
-    api("org.eclipse.jetty.websocket:javax-websocket-server-impl:$jettyVersion")
-
+    api("org.eclipse.jetty.websocket:websocket-jakarta-server:$jettyVersion")
```
gorbunkov, line 33: Do we need a jettyVersion? Won't it be taken from Spring Boot BOM? The same is for jakarta.servlet:jakarta.servlet-api and commons-logging - will these versions be taken from BOM?

Issue #1630, opened 2023-04-24
Some fields do not update required state in frontend while resetting property


Flaurite:
### Description
Fields do not update required state:
![required-fields](https://user-images.githubusercontent.com/22093943/234007527-78fd53bc-8d46-41d2-9271-cc7fc0ba52b7.gif)

### Steps to reproduce
1) Download project: [demo.zip](https://github.com/jmix-framework/jmix/files/11310829/demo.zip)

2) Open blank-view
3) Click on "Reset button" button then "Set required" button.

**ER**
All fields are required

**AR**
1) All fields except "Select" field are not required
2) "Select" field **always** required.

PR #1644, opened 2023-04-26
Some fields do not update required state in frontend while resetting property #1630

Flaurite:


glebfox (approved):

glebfox:
Check new components when merging to 2.0

PR #1646, opened 2023-04-26
Fixed some German translations that are incorrect or incomplete

timlehmkuhl:
I have noticed a few incorrect german translations, which I have fixed here.


mariodavid (approved):
looks good to me 👍 thanks for the contribution!

Issue #1571, opened 2023-04-06
Long Class-Path may break the frontend 


Flaurite:
### Description
See forum: [discussion](https://forum.jmix.io/t/getting-white-and-blank-screen-with-0-selectlocale-error/2780/3)

Most probably the problem occurs on Windows OS. Vaadin for dev server collects classes with specific annotations to get JS and CSS dependencies.
If classpath is very long, Gradle creates temporary jar file that contains all classpath. However Spring consider this jar without classpath that it contains. So we can see the broken UI.

### Steps to reproduce
1. Create project (1.5+)
2. Add dependencies:
```
implementation 'com.haulmont.yarg:yarg-api:2.2.20'
implementation 'com.haulmont.yarg:yarg:2.2.20'
```
3) Run the application

**AR**
Blank screen and errors in the console.

Flaurite:
As "half" workaround you can try to define application property:
```
vaadin.productionMode=true
```
It allows you to start the application, however some components does not work correctly (at least DatePicker, DateTimePicker, TimePicker, DrawerToggle). Not sure the cause of problem for now. Probably need to create another issue.

Flaurite:
Vaadin suggests the following work arounds:
https://vaadin.com/docs/latest/guide/start/gradle#known-issues

Flaurite:
Fixed only in Release 1.5

PR #1659, opened 2023-04-28
Long Class-Path may break the frontend #1571

Flaurite:
Fix https://github.com/jmix-framework/jmix/issues/1571

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixVaadinServletContextInitializer.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixVaadinServletContextInitializer.java
@@ -24,6 +24,14 @@ import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 import javax.servlet.ServletException;
 
+/**
+ * Workaround on Windows OS if you build project using gradle with long class path.
+ * This class extends {@link VaadinServletContextInitializer} and overrides it using
+ * {@code jmix.core.exclude-beans} property in module properties.
+ * <p>
+ * Also see issue and comments:
+ * <a href="https://github.com/jmix-framework/jmix/issues/1571">jmix-framework/jmix#1571</a>
+ */
 public class JmixVaadinServletContextInitializer extends VaadinServletContextInitializer {
 
     protected ApplicationContext applicationContext;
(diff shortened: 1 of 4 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixVaadinServletContextInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.vaadin.flow.spring.VaadinServletContextInitializer;
+import org.springframework.context.ApplicationContext;
+
+import javax.servlet.ServletContext;
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import javax.servlet.ServletException;
+
+public class JmixVaadinServletContextInitializer extends VaadinServletContextInitializer {
```
gorbunkov, line 27: Let's add javadoc that explains that this is a workaround for the issue, like you did in JmixTempJarManifestUtils class

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixVaadinServletContextInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+            return;
+        }
+
+        super.onStartup(servletContext);
+    }
+
+    /**
+     * Listener sets provided class path to "java.class.path" property.
+     */
+    protected class SetupTempClassPathContextListener implements ServletContextListener {
```
gorbunkov, line 64: Do we need two separate classes SetupTempClassPathContextListener and RestoreClassPathContextListener? It seems that we can have one class and create two instances of the same class passing different arguments to its constructor

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/module.properties
```diff
@@ -14,4 +14,6 @@
 # limitations under the License.
 #
 
-jmix.flowui.menu-config = io/jmix/flowui/menu.xml
\ No newline at end of file
+jmix.flowui.menu-config = io/jmix/flowui/menu.xml
+
+jmix.core.exclude-beans=contextInitializer
```
gorbunkov, line 19: Let's add a comment that explains why we add this

gorbunkov (approved):

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/module.properties
Flaurite: Fixed

Issue #1529, opened 2023-03-24
Actions created by class cannot be overridden


knstvk:
Actions created using `io.jmix.ui.Actions#create(java.lang.Class<T>)` method cannot be overridden in the application. For example, `io.jmix.ui.component.filter.FilterSupport#createFilterAction` creates actions by their classes, so overriding of any filter action is impossible.

Perhaps we should only create actions by ID and deprecate `Actions#create(java.lang.Class<T>)` methods, or when creating by class search registered action types with the same id lower in the hierarchy.

Originates from [forum topic](https://forum.jmix.io/t/unique-save-filter-name-configuration/2676).
See also #420.

gorbunkov:
Remove `Actions#create(java.lang.Class<T>)` in 2.0 and deprecate in 1.6

PR #1771, opened 2023-05-31
Actions created by class cannot be overridden jmix-framework/jmix#1529

KremnevDmitry:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/Actions.java
```diff
@@ -30,7 +30,9 @@ public interface Actions {
 
     Action create(String actionTypeId, String id);
 
+    @Deprecated
```
glebfox, line 33: JavaDoc: Use instead...

jmix-ui/ui/src/main/java/io/jmix/ui/Actions.java
```diff
@@ -30,7 +30,9 @@ public interface Actions {
 
     Action create(String actionTypeId, String id);
 
+    @Deprecated
```
glebfox, line 33: JavaDoc: Use instead...

glebfox (changes requested):

Issue #1368, opened 2023-01-19
Bean validation on persistence layer


gorbunkov:
Currently bean validation is performed in different ways when you store entities from the editor screen and using REST API.

Editor screen only validates entity attributes that have corresponding fields in the screen. If in the editor screen you update some entity attribute programmatically and this attribute doesn't have corresponding UI field, then the entity will be saved with invalid value.

In REST API full entity validation is performed before save.

It seems that we may try to do the validation on the persistence layer. The first option is to do the validation using ORM. We explicitly turn the EclipseLink bean validation off in the `JmixEclipseLinkJpaVendorAdapter` class:

```java
map.put("javax.persistence.validation.mode", "NONE");
```

Maybe we should try yo turn it on.

The second option may be to invoke bean validation somewhere at DataManager after all entity listeners are invoked (because some entity attributes may be filled in listeners).

Gavrilov-Ivan:
- Automatical bean validation on persistence layer is enabled
- Can be disabled by setting javax.persistence.validation.mode app property to NONE
- Validator groups can be configured by the following application properties (full class name required as value): jakarta.persistence.validation.group.pre-persist, jakarta.persistence.validation.group.pre-update, jakarta.persistence.validation.group.pre-remove. Default group by default.
- JmixLocalValidatorFactoryBean with Jmix-specific message interpolator is used as validation factory for auto validation (the same as for the manual validation on UI\REST side)

lovtsovaik:
verified
- [x] bean validation enabled
- [x] jakarta.persistence.validation.mode=none
- [x] jakarta.persistence.validation.group.pre-persist, pre-update, pre-remove
- [x] Jmix-specific messages

PR #1776, opened 2023-06-01
[PR] Bean validation on persistence layer #1368

Gavrilov-Ivan:
- Automatical bean validation on persistence layer is enabled
- Can be disabled by setting `javax.persistence.validation.mode` app property to `NONE`
- Validator groups can be configured by the following application properties (full class name required as value): `jakarta.persistence.validation.group.pre-persist`, `jakarta.persistence.validation.group.pre-update`, `jakarta.persistence.validation.group.pre-remove`. `Default` group by default.
- JmixLocalValidatorFactoryBean with Jmix-specific message interpolator is used as validation factory for auto validation (the same as for the manual validation on UI\REST side)

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/exception/ConstraintViolationExceptionHandler.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/exception/ConstraintViolationExceptionHandler.java
@@ -27,6 +27,10 @@ import org.springframework.stereotype.Component;
 
 import java.util.Set;
 
+/**
+ * Handles ConstraintViolationException that can be thrown by bean validation on persistence layer.
+ * Displays violation messages as screen notifications.
+ */
 @Component
 public class ConstraintViolationExceptionHandler extends AbstractUiExceptionHandler {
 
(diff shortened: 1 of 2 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/exception/ConstraintViolationExceptionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+import io.jmix.flowui.view.ViewValidation;
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.ConstraintViolationException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.Set;
+
+@Component
+public class ConstraintViolationExceptionHandler extends AbstractUiExceptionHandler {
```
gorbunkov, line 31: Please add javadoc explaining when this class is used and what it does.

gorbunkov (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/exception/ConstraintViolationExceptionHandler.java
Gavrilov-Ivan: Done

gorbunkov (approved):

gorbunkov:
@Gavrilov-Ivan Could you please copy the information from the PR description to the issue comment? It may be useful for QA and others who will examine the issue.

Issue #1493, opened 2023-03-09
Kotlin templates for new project and views


knstvk:

KremnevDmitry:
Blocked by https://github.com/jmix-framework/jmix/issues/1357

KremnevDmitry:
Fixed: 

- BlankView creation
- BPM Form creation

Embedable and Mapped Entity templates will be implemented later

PR #1789, opened 2023-06-06
Kotlin templates for new project and views jmix-framework/jmix#1493

KremnevDmitry:


jmix-templates/content/flowui/detail-dto/descriptor.xml
```diff
@@ -0,0 +1,34 @@
...
+<view xmlns="http://jmix.io/schema/flowui/view"
+      title="${messageKeys['detailTitle']}"
+      focusComponent="form">
+    <data>
+        <instance id="${dcId}"
+                  class="${entity.fqn}">
+            <loader id="${dlId}"/>
+        </instance>
+    </data>
+    <facets>
```
glebfox, line 20: At the time it handle loading, there won't be any entity in the container, test that it doesn't produce any exceptions

jmix-templates/content/flowui/detail-dto/controller.java
```diff
@@ -0,0 +1,22 @@
...
+${classComment}
+<%}%>@Route(value = "${detailRoute}", layout = <%if (!api.jmixProjectModule.isApplication()) {%> DefaultMainViewParent.class <%} else {%>MainView.class<%}%>)
+@ViewController("${detailId}")
+@ViewDescriptor("${detailDescriptorName}.xml")
+@EditedEntityContainer("${dcId}")
+public class ${detailControllerName} extends StandardDetailView<${entity.className}> {
+
+    @Override
+    protected void findEntityId(BeforeEnterEvent event) {
+        // do nothing
```
glebfox, line 20: Because DTO entity cannot be loaded by Id, we need to prevent Id parsing from route parameters

glebfox (changes requested):

glebfox (approved):

PR #1852, opened 2023-06-16
Pluggable styles from addons

glebfox:
Fix for https://github.com/jmix-framework/jmix/issues/1851

**Solution**

An Add-on may contain a file with custom styles under `META-INF/resources/` directory. In order to add these styles to the resulting app the `jmix.ui.import-styles` property must be added to the `module.properties` file. The value of the property is a relative path from `META-INF/resources/`, e.g. `jmix.ui.import-styles = addon-styles/test.css`

Custom styles are added as `<style type="text/css">` to the `head` element, e.g.:

```
<style type="text/css">
.test {
    color: red;
}

.test-2 {
    color: green;
}

.test-3 {
    color: blue;
}
</style>
```

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixServiceInitListener.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/JmixServiceInitListener.java
@@ -58,7 +58,7 @@ public class JmixServiceInitListener implements VaadinServiceInitListener, Appli
 
     private static final Logger log = LoggerFactory.getLogger(JmixServiceInitListener.class);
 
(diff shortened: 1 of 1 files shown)


knstvk:
Shouldn't the property be called `jmix.ui.export-styles`?

glebfox:
> Shouldn't the property be called `jmix.ui.export-styles`?

We import them in the result app, but export from add-on. From add-on point of view, yeah, `export-styles` looks good.

knstvk:
> From add-on point of view, yeah, export-styles looks good.`

We put this property only to the add-on's module.properties, right? Then `export` seems more correct.

gorbunkov (approved):

PR #1911, opened 2023-06-22
Reports Add-on: Change added style in add-on (1901)

NikitaShchienko:


jmix-reports/reports-flowui/src/main/resources/META-INF/resources/styles/report-style.css
```diff
@@ -14,87 +14,80 @@
  * limitations under the License.
```
glebfox, line 14: Let's use plural form for file name: `report-styles.css`

glebfox (changes requested):

glebfox (approved):

Issue #1926, opened 2023-06-23
Add localization for Quartz entity names


glebfox:
Non of quartz entities have localization for their names

* JobModel - `io.jmix.quartz.model/JobModel`
* TriggerModel - `io.jmix.quartz.model/TriggerModel`

## Solution 

Added en and ru localizations

For QA: check also #1667

konyashkina:
Tested on
Jmix version: 2.0.999-SNAPSHOT
Jmix Studio plugin version: 2.0.SNAPSHOT5366-231
IntelliJ version: IntelliJ IDEA 2023.1.4 (Ultimate Edition)

PR #1929, opened 2023-06-23
Add localization for Quartz entity names

glebfox:
Fix for #1926


jmix-translations/content/io/jmix/quartz/messages_ru.properties
```diff
@@ -47,6 +48,7 @@ io.jmix.quartz.model/ScheduleType=Тип расписания
 io.jmix.quartz.model/ScheduleType.CRON_EXPRESSION=Выражение Cron
 io.jmix.quartz.model/ScheduleType.SIMPLE=Простое
 
+io.jmix.quartz.model/TriggerModel=Условие
```
gorbunkov, line 51: maybe call it just "триггер"? Or have you seen the usage of word "условие" for this?

gorbunkov: I would name it "Триггер" and "Описание триггера". WDYT?

PR #1938, opened 2023-06-27
Reports Add-on: Add translations for views and models (1912)

NikitaShchienko:


jmix-translations/content/io/jmix/reportsflowui/messages_en.properties
gorbunkov, line 1: It seems that the filename with English translation should be `messages.properties`

gorbunkov (changes requested):

gorbunkov (approved):

PR #1945, opened 2023-06-27
Allow using CaseConversion in composed annotations (same as BeanValidation annotations)

LVN-hm:
The change has been taken from here https://github.com/ikuchmin/jmix/tree/case-conversion-1.4.4

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/entity/annotation/CaseConversion.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/entity/annotation/CaseConversion.java
@@ -16,16 +16,22 @@
 
 package io.jmix.core.entity.annotation;
 
-import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+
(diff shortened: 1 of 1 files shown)


jmix-core/core/src/main/java/io/jmix/core/entity/annotation/CaseConversion.java
```diff
@@ -16,16 +16,22 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+
 /**
  * Indicates that auto case conversion should be used for text input fields bound with marked entity field.
  */
 
-@Target({ElementType.FIELD})
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
```
gorbunkov, line 34: @LVN-hm Why do you need all these targets? If the plan is to use the `CaseConversion` as a meta-annotation then `ElementType.ANNOTATION_TYPE` should be enough. Do you really plan to use this annotation on constructors, methods, types, etc.?

Issue #1774, opened 2023-06-01
MetadataTools.hasUuid returns false if entity extends MappedSuperclass


glebfox:
Case:

1. Create a MappedSuperclass entity with IdType:  UUID (e.g. Person)
2. Create a new entity that extends it, e.g. Client
3. Invoke `MetadataTools.hasUuid` for the second entity (Client)

```
metadataTools.hasUuid(metadata.getClass(Client.class));
```

AR:
`hasUuid` returns `false`

ER:
`hasUuid` returns `true`

See [forum topic](https://forum.jmix.ru/t/nekorrektnyj-rezultat-metoda-metadatatools-hasuuid/1106).

PR #1976, opened 2023-07-04
MetadataTools.hasUuid returns false if entity extends MappedSuperclass

Gavrilov-Ivan:
See #1774

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
@@ -869,12 +869,44 @@ public class MetadataTools {
      * @return true if entity's {@link EntityEntry} implements {@link EntityEntryHasUuid}
      */
     public boolean hasUuid(MetaClass metaClass) {
-        return metaClass.getAnnotations().containsKey(UUID_KEY_ANN_NAME);
+        String uuidPropertyName = getUuidPropertyName(metaClass);
+        return uuidPropertyName != null;
     }
 
+    /**
+     * Gets name of uuid key property.
+     *
+     * @param clazz entity java class
+     * @return uuid property name or null if it doesn't exist
+     */
     @Nullable
     public String getUuidPropertyName(Class<?> clazz) {
-        return (String) metadata.getClass(clazz).getAnnotations().get(UUID_KEY_ANN_NAME);
+        MetaClass metaClass = metadata.getClass(clazz);
+        return getUuidPropertyName(metaClass);
+    }
+
+    /**
+     * Gets name of uuid key property.
+     *
+     * @param metaClass entity MetaClass
+     * @return uuid property name or null if it doesn't exist
+     */
+    @Nullable
+    public String getUuidPropertyName(MetaClass metaClass) {
+        String uuidProperty = (String) metaClass.getAnnotations().get(UUID_KEY_ANN_NAME);
+        if(uuidProperty != null) {
+            return uuidProperty;
+        } else {
+            MetaClass ancestor = metaClass.getAncestor();
+            while (ancestor != null) {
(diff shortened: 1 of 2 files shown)


jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -869,12 +869,44 @@ public List<String> getSoftDeleteProperties(Class<?> clazz) {
...
+     * @return uuid property name or null if it doesn't exist
+     */
+    @Nullable
+    public String getUuidPropertyName(MetaClass metaClass) {
+        String uuidProperty = (String) metaClass.getAnnotations().get(UUID_KEY_ANN_NAME);
+        if(uuidProperty != null) {
+            return uuidProperty;
+        } else {
+            MetaClass ancestor = metaClass.getAncestor();
+            while (ancestor != null) {
```
gorbunkov, line 901: As I see there is a `io.jmix.core.metamodel.model.MetaClass#getAncestors` method that returns a full list of metaclass ancestors. No need to do a while loop here if you use this method.

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -869,12 +869,44 @@ public List<String> getSoftDeleteProperties(Class<?> clazz) {
...
+    }
+
+    /**
+     * Gets name of uuid key property.
+     *
+     * @param metaClass entity MetaClass
+     * @return uuid property name or null if it doesn't exist
+     */
+    @Nullable
+    public String getUuidPropertyName(MetaClass metaClass) {
```
gorbunkov, line 895: Do we really need to introduce a new `getUuidPropertyName` method? Why not to implement the behavior in the existing `getUuidPropertyName(Class<?> clazz)`?

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -869,12 +869,44 @@ public List<String> getSoftDeleteProperties(Class<?> clazz) {
...
-        return metaClass.getAnnotations().containsKey(UUID_KEY_ANN_NAME);
+        String uuidPropertyName = getUuidPropertyName(metaClass);
+        return uuidPropertyName != null;
     }
 
+    /**
+     * Gets name of uuid key property.
+     *
+     * @param clazz entity java class
+     * @return uuid property name or null if it doesn't exist
```
gorbunkov, line 880: After your changes the method will throw an exception if there is no metaClass for the given java class. It is worth mentioning in the javadoc.

gorbunkov (changes requested):
It would be nice to have a test for the fixed issue

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
Gavrilov-Ivan: To iliminate double conversion `MetaClass` -> `JavaClass` -> `MetaClass`:
`hasUuid` accepts `MetaClass` and uses result of `getUuidPropertyName` - original one accepts `Class<?>` and convert it to `MetaClass`. Not a critical point, we can keep this conversion and remove extra method.

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
Gavrilov-Ivan: Done

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
Gavrilov-Ivan: Originally done the same way as `getPrimaryKeyName`.
Rewritten with `getAncestors`.

Gavrilov-Ivan:
Added test

gorbunkov (approved):

PR #2032, opened 2023-07-17
The changed report name is reset to the old one after making changes in the template (1924)

NikitaShchienko:


jmix-reports/reports-flowui/src/main/resources/io/jmix/reportsflowui/view/template/report-template-detail-view.xml
gorbunkov, line 1: What do you think about adding the "is default" checkbox inside the report template editor? Users may expect that they can do anything with the template from the editor. Currently we upload files from the detail view, but select a default template from the list view

gorbunkov (changes requested):

gorbunkov (approved):

PR #2033, opened 2023-07-17
New band not saved after adding on the Report template (1921)

NikitaShchienko:


jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/view/template/ReportTemplateDetailView.java
```diff
@@ -189,6 +189,13 @@ public void onBeforeSave(BeforeSaveEvent event) {
                 reportTemplate.setContent(bytes);
             }
         }
+
+        // When changing data in an entity with a composition relationship, you must manually update the changes
```
gorbunkov, line 193: Let's make the comment more clear by adding  "The same ReportTemplate instance may be stored both in Report.templates collection and in the Report.defaultTemplate reference"

gorbunkov (approved):

PR #2056, opened 2023-07-19
Implement a mechanism to prevent access to the screen when opened in a dialog box (2017)

NikitaShchienko:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/UiAccessChecker.java
```diff
@@ -78,6 +78,19 @@ public boolean isViewPermitted(Class<?> target) {
...
     }
 
+    /**
+     * Checks for access to the view.
+     * Thrown exception if no access
+     *
+     * @param target class to check
+     * @throws AccessDeniedException if the user does not have access to the view
+     */
+    public void checkAccessView(Class<?> target) {
```
glebfox, line 88: `checkViewPermitted` for naming similarity.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/UiAccessChecker.java
```diff
@@ -78,6 +78,19 @@ public boolean isViewPermitted(Class<?> target) {
         return hasAccess;
     }
 
+    /**
+     * Checks for access to the view.
+     * Thrown exception if no access
```
glebfox, line 83: `Throws`. `.` at the end of the sentence.

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/UiAccessChecker.java
```diff
@@ -78,6 +78,19 @@ public boolean isViewPermitted(Class<?> target) {
...
     }
 
+    /**
+     * Checks for access to the view.
+     * Throws exception if no access.
+     *
+     * @param target class to check
+     * @throws AccessDeniedException if the user does not have access to the view
+     */
+    public void checkViewPermitted(Class<?> target) {
```
gorbunkov, line 88: Why did you introduce a new method here? Why not to use an existing `isViewPermitted()`? As I see this method is used only once inside the `AbstractWindowBuilderProcessor`. Why not to write these lines there?

```java
        if (!accessChecker.isViewPermitted(target)) {
            throw new AccessDeniedException("view", target.getName());
        }
```

NikitaShchienko: A handler is created for `AccessDeniedException`, I find it convenient to have one method that can potentially be used in other parts of the application.

gorbunkov: Now it is only used in one place, but maybe it can be used somewhere else in future. If Gleb is fine with the new method in "UiAccessChecker", then let it be.

glebfox (approved):

Issue #1940, opened 2023-06-27
Remove deprecated code


knstvk:
E.g. `io.jmix.flowui.model.InstanceContainer#getView`

---

**Breaking Changes**

* `EntitiesController.java`
  * Removed optional `view` parameter from `loadEntity` method (used for backward compatibility with CUBA)
  * Removed optional `view` parameter from `loadEntitiesList` method (used for backward compatibility with CUBA)
  * Removed optional `view` parameter from `searchEntitiesListGet` method (used for backward compatibility with CUBA)
* `EntitiesControllerManager.java` - `SearchEntitiesRequestDTO` - removed `view` and `viewName` fields used for backward compatibility with CUBA

PR #2067, opened 2023-07-24
Remove deprecated code (1940)

glebfox:
See #1940


jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/mapping/UuidMappingProcessor.java
```diff
@@ -48,31 +46,28 @@ public class UuidMappingProcessor implements MappingProcessor {
 
     @Override
     public void process(MappingProcessorContext context) {
+        DatabasePlatform platform = context.getSession().getPlatform();
+        Preconditions.checkState(platform instanceof UuidMappingInfo,
```
gorbunkov, line 50: I am not sure that we should forbid using DatabasePlatform that does not implement our `UuidMappingInfo` interface. In theory users may want to use their own database that is not supported by Jmix, but is supported by EclipseLink. In this case, they will get an exception if they don't create a wrapper for the database platform. Maybe we just should not set the converter for UUID fields in case the database platform doesn't implement the `UuidMappingInfo` interface? Let's ask @knstvk , maybe he will remember what was an idea when he introduced the `UuidMappingInfo`.

gorbunkov (changes requested):

jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/mapping/UuidMappingProcessor.java
glebfox: That was my first thought, but since we're relying on `UuidMappingInfo` it's probably better to throw an exception rather than converting to a String without any guarantee that it's a valid UUID.

jmix-rest/rest/src/main/java/io/jmix/rest/impl/controller/EntitiesController.java
```diff
@@ -46,18 +45,16 @@ public class EntitiesController {
...
                              @RequestParam(required = false) Boolean dynamicAttributes,
                              @RequestParam(required = false) String modelVersion) {
-        return entitiesControllerManager.loadEntity(entityName, entityId, StringUtils.defaultString(fetchPlan, view),
+        return entitiesControllerManager.loadEntity(entityName, entityId, fetchPlan,
                 returnNulls, dynamicAttributes, modelVersion);
     }
 
     @GetMapping("/{entityName}")
     public ResponseEntity<String> loadEntitiesList(@PathVariable String entityName,
-                                                   @RequestParam(required = false) String view,
```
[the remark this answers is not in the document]
glebfox: will mention

jmix-rest/rest/src/main/java/io/jmix/rest/impl/service/EntitiesControllerManager.java
```diff
@@ -264,17 +264,9 @@ public EntitiesSearchResult searchEntities(String entityName, String searchReque
             throw new RestAPIException("Cannot parse entities filter", "Entities filter cannot be null", HttpStatus.BAD_REQUEST);
         }
 
-        //for backward compatibility we should support both 'view' and 'viewName' properties. In future the
-        //'viewName' parameter will be removed
-        String view = !Strings.isNullOrEmpty(searchEntitiesRequest.getView()) ?
```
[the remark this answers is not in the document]
glebfox: will mention

jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/mapping/UuidMappingProcessor.java
glebfox: Rolled back `UuidConverter` as it is used as a fallback if the database platform doesn't implement `UuidMappingInfo`. Added more detailed JavaDoc.

knstvk (approved):
👍

PR #2075, opened 2023-07-25
Features/1898 wizard fix

breakingfox:


jmix-reports/reports-flowui/src/main/resources/io/jmix/reportsflowui/view/reportwizard/report-wizard-creator-view.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <view xmlns="http://jmix.io/schema/flowui/view"
-      title="msg://reportWizard.title">
+      title=" ">
```
NikitaShchienko, line 3: If you don't specify a title, will it be equivalent to " "?

breakingfox: it will make close button appear on the top left side of the window, with " " it is on the right side

Issue #1573, opened 2023-04-06
[Filter] Provide the ability to change apply-shortcut


KremnevDmitry:
FlowUI, ClassicUI: added new global application.propertiy
`jmix.ui.component.filter-apply-shortcut`

E.g.
application.properties file
`jmix.ui.component.filter-apply-shortcut = ALT-P`

Can be replaced by `applyShortcut` xml attribute.
The attribute has the highest priority.

PR #2115, opened 2023-08-04
[Filter] Provide the ability to change apply-shortcut

KremnevDmitry:
Related to: #1573
PR only for 2.0 release

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
@@ -378,6 +378,19 @@ public class GenericFilter extends Composite<JmixDetails>
         }
     }
 
+    /**
+     * @return {@link KeyCombination} that is used to apply the filter
+     */
+    @Nullable
+    public KeyCombination getApplyShortcut() {
+        return applyButton.getShortcutCombination();
+    }
+
+    /**
+     * Sets a new {@link KeyCombination} to apply the filter.
+     *
+     * @param applyShortcut string representation of a {@link KeyCombination}
+     */
     public void setApplyShortcut(String applyShortcut) {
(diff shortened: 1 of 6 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -373,6 +378,27 @@ public void setAutoApply(boolean autoApply) {
...
+    public KeyCombination getApplyShortcut() {
+        return applyButton.getShortcutCombination();
+    }
+
+    /**
+     * Sets a new {@link KeyCombination} to apply the filter.
+     *
+     * @param applyShortcut string representation of a {@link KeyCombination}
+     */
+    public void setApplyShortcut(String applyShortcut) {
```
Flaurite, line 394: It seems that method can take Nullable parameter. Check it and if so, add annotation.

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/view/layout.xsd
```diff
@@ -3573,6 +3573,7 @@
 
                 <xs:attribute name="dataLoader" type="xs:string" use="required"/>
                 <xs:attribute name="autoApply" type="xs:boolean"/>
+                <xs:attribute name="applyShortcut" type="xs:string"/>
```
Flaurite, line 3576: Do we have XML loading test for this attribute?

Flaurite (changes requested):

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/view/layout.xsd
KremnevDmitry: Will be added with https://github.com/jmix-framework/jmix/issues/2117

Flaurite (approved):

PR #2118, opened 2023-08-07
[Filter] Provide the ability to change apply-shortcut 1.6.0

KremnevDmitry:


jmix-ui/ui/src/main/java/io/jmix/ui/component/impl/FilterImpl.java
```diff
@@ -208,6 +207,35 @@ public void setAutoApply(boolean autoApply) {
...
+                searchButton.setShortcutCombination(applyShortcut);
+            } else {
+                searchButton.setShortcut(null);
+            }
+        }
+    }
+
+    @Override
+    public void setApplyShortcut(@Nullable String applyShortcut) {
+        if (applyShortcut != null) {
```
Flaurite, line 232: Use ternary operator

jmix-ui/ui/src/main/java/io/jmix/ui/component/impl/FilterImpl.java
```diff
@@ -208,6 +207,35 @@ public void setAutoApply(boolean autoApply) {
...
+        return searchButton.getShortcutCombination();
+    }
+
+    @Override
+    public void setApplyShortcutCombination(@Nullable KeyCombination applyShortcut) {
+        KeyCombination oldValue = this.applyShortcut;
+        if (!Objects.equals(oldValue, applyShortcut)) {
+            this.applyShortcut = applyShortcut;
+
+            if (applyShortcut != null) {
```
Flaurite, line 222: I think, here we can skip null check

Flaurite (changes requested):

jmix-ui/ui/src/main/java/io/jmix/ui/component/impl/FilterImpl.java
KremnevDmitry: `If` operator is useless anyway
KeyCombination.create(applyShortcut) accepts `@Nullable` values

Flaurite (approved):

PR #2157, opened 2023-08-17
[CodeEditor] Implement an application theme listener to change the mode jmix-framework/jmix#2046

KremnevDmitry:


jmix-flowui/flowui-kit/src/main/resources/META-INF/frontend/src/code-editor/jmix-code-editor.js
```diff
@@ -155,7 +158,45 @@ class JmixCodeEditor extends ResizeMixin(InputFieldMixin(ThemableMixin(ElementMi
         });
     }
 
-    /** @protected */
+    initApplicationThemeObserver() {
+        // Apply current application theme as initial value
+        this._applyTheme()
+
+        this._applicationThemeObserver = new MutationObserver(mutations => {
```
KremnevDmitry, line 165: As far as I've found out, mutation is the only way to listen for changes to an element's home.

I have tried:

- `DOMSubtreeModified` - Listens for any change to the element
- `DOMAttrModified event` - Legacy/Deprecated [[doc](http://help.dottoro.com/ljdchxcl.php)]
- `addPropertyChangeListener` - only applies to the server-side component. Implemented via state listening.
- `setInterval ` looks like some kind of hacking + perfomance issue

In this case - mutationObserver looks like a nice convenient way to track changes.

glebfox (approved):

PR #2164, opened 2023-08-23
Romanian translation

sorinfederiga:
Hello,

I have translated to romanian flowui modules
Thank you!


knstvk:
Hi Sorin,
Thank you very much!
Could you fix the messages to include non-ASCII characters directly, as it's done in other translations and in your previous contribution? I believe you should not use the `native2AsciiForPropertiesFiles="true"` option in the IDE.

knstvk:
Thanks!

PR #2166, opened 2023-08-23
Add ScreenSettingsFacet to the Run Reports screen (2024)

breakingfox:


jmix-reports/reports-ui/src/main/resources/io/jmix/reportsui/screen/report/run/report-run.xml
```diff
@@ -36,6 +36,11 @@
 
     <facets>
         <dataLoadCoordinator auto="true"/>
+        <screenSettings id="settingsFacet">
```
NikitaShchienko, line 39: Let's also add `screenSettings` to those screens in which this facet was previously (see version 1.5.3)

PR #2197, opened 2023-08-31
Proposal for fixing validation annotations @Past, @Future, etc 

VitaliiEv:
See issue https://github.com/cuba-platform/cuba/issues/3232


gorbunkov:
ValidationClockProvider class has been completely removed while working on #3172

Issue #1742, opened 2023-05-25
hierarchy-depth for each filter


tychkov:
Hello, guys!

We can use jmix.ui.component.filter-properties-hierarchy-depth for changing deep scanning of filters.
Usually level 3 deep is enough. But some cases demand more deep hierarhy (not for all).

Could you add deep-hierarhy property to Filter for flexible tuning?

yuriy-khrebtov:
Now we can specify `propertyHierarchyDepth` property of a generic filter in a view descriptor or in java code (using `GenericFilter.setPropertyHieratchyDepth(int)`

PR #2215, opened 2023-09-08
hierarchy-depth for each filter #1742

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/1742

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/app/filter/condition/PropertyFilterConditionDetailView.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/app/filter/condition/PropertyFilterConditionDetailView.java
@@ -31,7 +31,6 @@ import io.jmix.core.querycondition.PropertyConditionUtils;
 import io.jmix.flowui.UiComponents;
 import io.jmix.flowui.component.SupportsTypedValue;
 import io.jmix.flowui.component.UiComponentUtils;
-import io.jmix.flowui.component.genericfilter.Configuration;
 import io.jmix.flowui.component.genericfilter.FilterMetadataTools;
 import io.jmix.flowui.component.propertyfilter.PropertyFilter;
 import io.jmix.flowui.component.propertyfilter.PropertyFilterSupport;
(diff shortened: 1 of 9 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -996,6 +998,30 @@ protected void clearValues() {
         }
     }
 
+    /**
+     * @return hierarchy depth of entity properties available for filtering.
+     * @see #setPropertyHierarchyDepth(Integer)
+     */
+    @Nullable
+    public Integer getPropertyHierarchyDepth() {
```
glebfox, line 1006: This is not entity attribute, why `Integer` and `@Nullable`?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterMetadataTools.java
```diff
@@ -53,24 +54,32 @@ public FilterMetadataTools(MetadataTools metadataTools,
         this.metadata = metadata;
     }
 
-    public List<MetaPropertyPath> getPropertyPaths(MetaClass filterMetaClass,
-                                                   String query,
-                                                   @Nullable Predicate<MetaPropertyPath> propertiesFilterPredicate) {
-        return getPropertyPaths(filterMetaClass, query, filterMetaClass, 0, "",
-                propertiesFilterPredicate);
+    public List<MetaPropertyPath> getPropertyPaths(GenericFilter filter) {
```
glebfox, line 57: We don't change public API without a real need. In case of `propertyHierarchyDepth`:
1. `GenericFilter` component can load default value from properties, so it will return a value. 
2. If you need to get actual value from a filter, create a new method that accepts this value. The old method should pass a value from properties as a default value.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/app/filter/condition/PropertyFilterConditionDetailView.java
```diff
@@ -108,9 +106,8 @@ protected void onReady(@SuppressWarnings("unused") ReadyEvent event) {
     }
 
     protected void initPropertyField() {
-        if (filterMetaClass != null) {
-            List<String> properties = filterMetadataTools.getPropertyPaths(filterMetaClass, query,
-                            propertiesFilterPredicate).stream()
+        if (currentConfiguration != null) {
```
glebfox, line 109: all of this solves some bug? I don't get how it relates to propertyHierarchyDepth

glebfox (changes requested):

glebfox (approved):

Issue #1514, opened 2023-03-17
Prevent from creating filter configurations with the same name


knstvk:
Now a user can save generic filter configuration with the same name.
See [forum topic](https://forum.jmix.io/t/unique-save-filter-name-configuration/2676).

yuriy-khrebtov:
We implemented validation for uniqueness of filter configuration names when create or edit a configuration. Now, when a user tries to save a configuration with name that is already assigned to another configuration, save operation will fail with notification if the scope of configuration to save is equal to the scope of existing configuration. 
For example, if a user saves **global** configuration with name "conf1" and there is already **user** configuration with the same name then save operation will be successful. But if he will try to save another **user** configuration with the same name, save operation will fail.
Name validation can be enabled/disabled using `jmix.ui.component.filter-configuration-unique-names` (default is true).
Also, global configs in filter popup button are marked with * symbol to distinct them from user configs.

PR #2220, opened 2023-09-11
Prevent from creating filter configurations with the same name #1514

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/1514

### DIFF
--- a/jmix-ui/ui-data/src/main/java/io/jmix/uidata/action/filter/FilterSaveAsAction.java
+++ b/jmix-ui/ui-data/src/main/java/io/jmix/uidata/action/filter/FilterSaveAsAction.java
@@ -50,6 +50,7 @@ import org.apache.commons.lang3.BooleanUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Consumer;
 
 import static io.jmix.ui.component.filter.FilterUtils.generateConfigurationId;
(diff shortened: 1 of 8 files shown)


jmix-ui/ui/src/main/java/io/jmix/ui/component/filter/FilterSupport.java
```diff
@@ -82,10 +82,10 @@ public void removeCurrentFilterConfiguration(Filter filter) {
 
     public ScreenFragment createFilterConfigurationFragment(FrameOwner owner,
                                                             boolean isNewConfiguration,
-                                                            Filter.Configuration currentConfiguration) {
+                                                            Filter filter) {
         Fragments fragments = UiControllerUtils.getScreenContext(owner).getFragments();
         FilterConfigurationModelFragment fragment = fragments.create(owner, FilterConfigurationModelFragment.class);
-        initFilterConfigurationFragment(fragment, isNewConfiguration, currentConfiguration);
+        initFilterConfigurationFragment(fragment, isNewConfiguration, filter.getCurrentConfiguration());
```
glebfox, line 88: What the point of this change in public API (even though it's an internal bean) if you still get `CurrentConfiguration` from passed filter.

jmix-ui/ui-data/src/main/java/io/jmix/uidata/filter/UiDataFilterSupport.java
```diff
@@ -117,6 +118,7 @@ public ScreenFragment createFilterConfigurationFragment(FrameOwner owner,
                 UiDataFilterConfigurationModelFragment.class);
 
         fragment.setDefaultForMeFieldVisible(defaultForMeFieldVisible);
+        fragment.setFilter(filter);
```
glebfox, line 121: `io.jmix.ui.component.Filter.Configuration#getOwner` should solve your task without changing public API

jmix-ui/ui-data/src/main/java/io/jmix/uidata/action/filter/FilterSaveAsAction.java
```diff
@@ -185,6 +186,14 @@ protected void openInputDialog() {
                                 "logicalFilterConditionEdit.uniqueConfigurationId"));
                     }
 
+                    String name = validationContext.getValue("nameField");
```
glebfox, line 189: What will happen if I mark my personal config as global?

glebfox (changes requested):

jmix-ui/ui-data/src/main/java/io/jmix/uidata/action/filter/FilterSaveAsAction.java
yuriy-khrebtov: Added separate validations for user and global configurations

Issue #1416, opened 2023-02-07
Save file with parameters using FileStorage


gorbunkov:
The `FileStorage` interface has a single method to store the file:

```java
FileRef saveStream(String fileName, InputStream inputStream);
```

We can add another method that additionally accepts a map of parameters:

```java
FileRef saveStream(String fileName, InputStream inputStream, Map<String, Object> parameters);
```

This will give us an ability to configure how filestorage implementation should store the file. Also we will be able to provide meta-information that may be stored in the FileRef URL, e.g. file size.

PR #2224, opened 2023-09-12
Save file with parameters using FileStorage jmix-framework/jmix#1416

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/1416

### DIFF
--- a/jmix-awsfs/awsfs/src/main/java/io/jmix/awsfs/AwsFileStorage.java
+++ b/jmix-awsfs/awsfs/src/main/java/io/jmix/awsfs/AwsFileStorage.java
@@ -17,6 +17,7 @@
 package io.jmix.awsfs;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.Maps;
 import io.jmix.core.FileRef;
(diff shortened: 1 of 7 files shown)


gorbunkov (approved):
@yuriy-khrebtov Please add the "breaking changes" label to the issue and write a comment there explaining that a new method has been added to the interface and if the project contains a custom file storage implementation, it must implement the new method.

yuriy-khrebtov:
### Breaking changes:
New method has been added to FileStorage interface to support additional save parameters:
```
/**
     * Saves an InputStream contents into the file storage using additional parameters.
     *
     * @param fileName    file name
     * @param inputStream input stream, must be closed in the calling code
     * @param parameters additional parameters that can be used in specific storage implementation
     *                   (for example, can be converted to FileRef parameters)
     * @return file reference
     * @throws IllegalArgumentException if arguments are incorrect
     * @throws FileStorageException     if something goes wrong
     */
    FileRef saveStream(String fileName, InputStream inputStream, Map<String, Object> parameters);
```
So if there is some custom implementation of the interface in a project it must implement the new method

PR #2230, opened 2023-09-12
Persian Translation

Abdurrahman494:


knstvk:
Hi @Abdurrahman494 
I'm very sorry, we somehow missed your PR.
Could you update it to make it compatible with the current master branch?
Also, it would be better to split it to separate PRs for the new Persian translation and updating Arabic and Turkish ones.

PR #2242, opened 2023-09-14
add CORS support for JWT authenticated API requests

acgiesbrecht:
add CORS support for JWT authenticated API requests

### DIFF
--- a/jmix-oidc/oidc-starter/src/main/java/io/jmix/autoconfigure/oidc/OidcAutoConfiguration.java
+++ b/jmix-oidc/oidc-starter/src/main/java/io/jmix/autoconfigure/oidc/OidcAutoConfiguration.java
@@ -143,7 +143,8 @@ public class OidcAutoConfiguration {
         public SecurityFilterChain securityFilterChain(HttpSecurity http,
                                                        JmixJwtAuthenticationConverter jmixJwtAuthenticationConverter,
                                                        ApplicationEventPublisher applicationEventPublisher) throws Exception {
-            http.oauth2ResourceServer(resourceServer -> {
+            http.cors().and()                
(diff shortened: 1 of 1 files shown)


jmix-oidc/oidc-starter/src/main/java/io/jmix/autoconfigure/oidc/OidcAutoConfiguration.java
```diff
@@ -143,7 +143,8 @@ public static class OAuth2ResourceServerConfiguration {
         public SecurityFilterChain securityFilterChain(HttpSecurity http,
                                                        JmixJwtAuthenticationConverter jmixJwtAuthenticationConverter,
                                                        ApplicationEventPublisher applicationEventPublisher) throws Exception {
-            http.oauth2ResourceServer(resourceServer -> {
+            http.cors().and()
```
gorbunkov, line 146: The `org.springframework.security.config.annotation.web.builders.HttpSecurity#cors()` method is deprecated. Could you please replace it with the `org.springframework.security.config.annotation.web.builders.HttpSecurity#cors(org.springframework.security.config.Customizer>)`?
I guess, It should look like `.cors(Customizer.withDefaults())`.

gorbunkov (changes requested):

jmix-oidc/oidc-starter/src/main/java/io/jmix/autoconfigure/oidc/OidcAutoConfiguration.java
acgiesbrecht: for spring used jmix 1.5.x it is not yet deprecated, but the requested change works for that version also.

PR #2259, opened 2023-09-19
Prevent from creating filter configurations with the same name #1514

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/1514


glebfox:
Check Flow UI GenericFilter if it also needs similar fix

PR #2291, opened 2023-10-02
WebDav add-on with Flow UI#1720

dtaimanov:


jmix-flowui/flowui/src/main/resources/io/jmix/flowui/view/layout.xsd
```diff
@@ -3525,6 +3525,7 @@
                     <xs:element name="responsiveSteps" type="responsiveStepsType" minOccurs="0"/>
                     <xs:element name="formItem" type="formItemType" minOccurs="0"/>
                     <xs:group ref="standardComponent"/>
+                    <xs:any namespace="##other" minOccurs="0" maxOccurs="unbounded"/><!--todo check for order-->
```
glebfox, line 3528: Either remove todo or fix it)

glebfox (changes requested):

glebfox:
What about translation? I think that at least english can be copied from webdav module.

Issue #2285, opened 2023-09-27
Support all click events


Fedoseew:
See Studio issue: https://youtrack.jmix.io/issue/JST-4268/ClickEvent-for-button-are-duplicated.

**Description:**
We have 3 similar handlers from com.vaadin.flow.component.ClickNotifier:
- addClickListener
- addDoubleClickListener
- addSingleClickListener

**Possible solution:** 
Add parameter "method" (or "subject") in @Subsribe annotation to define handler method.

KremnevDmitry:
The following logic for scanning for a method for subscription has been implemented:
1. When scanning methods, methods with the same signature are identified and written to the separate map
2. Now, when resolving the `@Subscribe` annotation, the presence of the `subject` attribute is taken into account

# E.g.
### Case 1:
No `subject`:
In this case, the method that is in the cache will be applied.
```java
    @Subscribe(value = "button")
    public void onButtonClick(final ClickEvent<JmixButton> event) {
        notifications.show("Clicked");
    }
```

### Case 2: 
A `subject` attribute for a method with an **clashing signature** has been specified.
```java
    @Subscribe(value = "button", subject = "doubleClickListener")
    public void onButtonDoubleClick(final ClickEvent<JmixButton> event) {
        notifications.show("Double clicked");
    }
```

### Case 3:
A `subject` attribute for a method with a **default signature** has been specified.
```java
    @Subscribe(value = "textArea", subject = "valueChangeListener")
    public void onTextAreaComponentValueChange(final AbstractField.ComponentValueChangeEvent<JmixTextArea, Object> event) {
        notifications.show("Text area value changed");
    }
```

PR #2307, opened 2023-10-03
Support all click events  jmix-framework/jmix#2285

KremnevDmitry:
See: #2285

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/Subscribe.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/Subscribe.java
@@ -58,6 +58,11 @@ public @interface Subscribe {
     String id() default "";
 
     /**
+     * @return property name that will be set using annotated method
+     */
+    String subject() default "";
+
+    /**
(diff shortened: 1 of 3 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/Subscribe.java
```diff
@@ -57,6 +57,11 @@
     @AliasFor("value")
     String id() default "";
 
+    /**
+     * @return property name that will be set using annotated method
```
glebfox, line 61: what? isn't it explicit method name?

glebfox (approved):

glebfox (changes requested):
1. Test
2. `subject` should work for non method clash cases

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/Subscribe.java
KremnevDmitry: Fixed

glebfox (approved):

Issue #1972, opened 2023-07-03
Remove CaseConversion annotation


gorbunkov:
The `io.jmix.core.entity.annotation.CaseConversion` is not used in Jmix 2.0+ and should be removed.

KremnevDmitry:
Studio issue has already been solved: https://youtrack.jmix.io/issue/JST-3812

PR #2320, opened 2023-10-04
Remove CaseConversion annotation jmix-framework/jmix#1972

KremnevDmitry:
See: #1972


jmix-core/core/src/main/java/io/jmix/core/entity/annotation/CaseConversion.java
```diff
@@ -1,44 +0,0 @@
...
-import static java.lang.annotation.ElementType.TYPE_USE;
-
-/**
- * Indicates that auto case conversion should be used for text input fields bound with marked entity field.
- */
-
-@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
-@Retention(RetentionPolicy.RUNTIME)
-@MetaAnnotation
-public @interface CaseConversion {
```
glebfox, line 37: Make it Deprecated in 2.0 with indication that it will be removed

glebfox (approved):

jmix-core/core/src/main/java/io/jmix/core/entity/annotation/CaseConversion.java
KremnevDmitry: Got it

KremnevDmitry: Fixed by: c0d6457c30abce28773b8060f8080212dc9d76a3

Issue #1923, opened 2023-06-23
Support Items fetch callback


glebfox:
Consider implementing com.vaadin.flow.data.provider.CallbackDataProvider instead of separate component

## Solution

The `setItemsFetchCallback` method is added for the following components:

* `ComboBox`
* `EntityComboBox`
* `MultiselectComboBox`
* `MultiselectComboBoxPicker`

### XML

All components support an optional `itemsQuery` element that enables defining a query for selecting suggested values.

The `itemsQuery` element for the `EntityComboBox` has the following attributes:
* `class` (required) - a full qualified name of an entity class.
* `fetchPlan` - an optional attribute that specifies the fetch plan to be used for loading the queried entity.
* `escapeValueForLike` - enables searching for the values that contain special symbols: `%`, `\`, etc. The default value is `false`.
* `searchStringFormat` - a string that contains a variable to be substituted with actual value by `io.jmix.flowui.sys.substitutor.StringSubstitutor`. The default implementation delegates its work to `org.apache.commons.text.StringSubstitutor`.

The `itemsQuery` element for the `EntityComboBox` has the following nested elements:
* `query` - an element that contains a JPQL query
* `fetchPlan` - an optional descriptor of inline fetch plan

```
<entityComboBox id="assigneeField" metaClass="User" width="25em">
    <actions>
        <action id="entityLookup" type="entity_lookup"/>
        <action id="entityClear" type="entity_clear"/>
    </actions>
    <itemsQuery class="com.company.demo.entity.User"
                searchStringFormat="(?i)%${inputString}%"
                escapeValueForLike="true">
        <fetchPlan extends="_instance_name"/>
        <query>
            <![CDATA[select e from User e where e.username like :searchString escape '\' order by e.username asc]]>
        </query>
    </itemsQuery>
</entityComboBox>
```

The `itemsQuery` element for the `ComboBox` has the following attributes:
* `escapeValueForLike` - enables searching for the values that contain special symbols: `%`, `\`, etc. The default value is `false`.
* `searchStringFormat` - a string that contains a variable to be substituted with actual value by `io.jmix.flowui.sys.substitutor.StringSubstitutor`. The default implementation delegates its work to `org.apache.commons.text.StringSubstitutor`.

The `itemsQuery` element for the `EntityComboBox` has the following nested element:
* `query` - an element that contains a JPQL query

[...]

PR #2366, opened 2023-10-16
Support Items fetch callback (improvements)

glebfox:
Improvements for https://github.com/jmix-framework/jmix/issues/1923

Changed:
* `searchStringFormat` attribute assumes `inputString` as a value placeholder instead of `searchString`

Added: 
* Nested `query` element, similar to data loaders
* Nested `fetchPlan` element that defines inline fetch plan, similar to data containers

Note: the `fetchPlan` attribute is still present and defines named fetch plan, same as for data containers


knstvk:
👍

PR #2383, opened 2023-10-20
Change nav display style from block to flex (2382)

breakingfox:


jmix-flowui/flowui-themes/src/main/resources/META-INF/resources/themes/jmix-lumo/app/main-view.css
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+nav {
```
glebfox, line 17: You must not change style of html element. In case of MainView, corresponding nav element has custom class name: `jmix-main-view-navigation`. Provide styles for it.

glebfox (changes requested):

glebfox (approved):

Issue #2213, opened 2023-09-08
Support for initial layout in MainView


glebfox:
In case no view is opened, it is required to display some default layout in the Main View, e.g. with logo or dash board.

---

Let's add `initialLayout` element to the main view, so if no view is opened, its content is displayed. The logic of handling switching between a view and initial layout must be located in the base class. 

### Solution

Optional `<initialLayout>` element is introduced inside `<appLayout>`, e.g.:

```
<appLayout>
    ...
    <initialLayout>
        <textField id="textField" label="TextField" clearButtonVisible="true"/>
    </initialLayout>
</appLayout>
```

---

When added in XML, the root `<initialLayout>` element effectively is `VerticalLayout`. 

Alternatively `StandardMainView`'s java API can be used to set/get initial layout:

* `StandardMainView#getInitialLayout` - returns root component of initial layout
* `StandardMainView#setInitialLayout` - sets a component that will be an initial layout

---

Components inside `<initialLayout>` can be injected in the `MainView`, e.g.:

```
@Route("")
@ViewController("MainView")
@ViewDescriptor("main-view.xml")
public class MainView extends StandardMainView {

    @ViewComponent
    private TypedTextField<String> textField;

    @Subscribe
    public void onInit(final InitEvent event) {
        textField.setValue("Programmatic Value");
        textField.setWidth("25em");
    }

    @Subscribe("textField")
    public void onTextFieldTypedValueChange(final TypedValueChangeEvent<TypedTextField<String>, String> event) {
        Notification.show(event.getValue());
    }
}
```

glebfox:
Related Studio bug: https://youtrack.jmix.io/issue/JST-4429/Studio-incorrectly-handles-initialLayout-element-of-MainView

PR #2420, opened 2023-10-31
[MainView] Support for initial layout (2213)

glebfox:
See: https://github.com/jmix-framework/jmix/issues/2213

### DIFF
--- a/jmix-flowui/flowui/src/test/resources/main_view/view/main-view.xml
+++ b/jmix-flowui/flowui/src/test/resources/main_view/view/main-view.xml
@@ -33,6 +33,7 @@
             </section>
         </drawerLayout>
         <initialLayout>
+            <textField id="textField" label="TextField" clearButtonVisible="true"/>
         </initialLayout>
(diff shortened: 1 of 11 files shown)


jmix-flowui/flowui/src/test/resources/main_view/view/main-view.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<mainView xmlns="http://jmix.io/schema/flowui/main-view"
```
Flaurite, line 2: Main view already exists in tests: `TestMainView`. May be we can move testing code in it?

Flaurite (approved):

jmix-flowui/flowui/src/test/resources/main_view/view/main-view.xml
glebfox: I saw this but decided against it so as not to affect other views.

Issue #1725, opened 2023-05-23
Don't reload entity on save and close screen


linkinek:
For now, the `windowCommitAndClose` action saves and exits the screen.
But after saving, the DataManager reloads the entity, and it doesn’t do it correctly, which is why Jmix loads more data than was originally indicated in the fetchPlan.

I think that the `windowCommitAndClose` action should not reload the entity, save with the `discardSaved` flag and close screen.

![CleanShot 2023-05-23 at 10 47 16@2x](https://github.com/jmix-framework/jmix/assets/14567888/a26d5589-fd40-4be7-9a92-32a2fd5b5044)

dtaimanov:
Note: It may cause performance issues on screen close in some specific cases.

knstvk:
If we don't reload the entity, the parent browser won't get the new instance and it needs to be refreshed.

> it doesn’t do it correctly, which is why Jmix loads more data than was originally indicated in the fetchPlan

It makes a new fetch plan corresponding to all loaded attributes in the graph, see `io.jmix.core.EntityStates#getCurrentFetchPlan`.

yuriy-khrebtov:
We changed default behaivour of detail views - now edited entity is not reloaded after save-and-close action if the view was opened by navigation or no parent data container was set to the view. This behaviour can be overriden setting "reloadSaved" property of custom detail view (which extends StandardDetailView) to false:
```
    @Subscribe
    public void onInit(final InitEvent event) {
        setReloadSaved(false);
    }
```

PR #2491, opened 2023-11-20
Don't reload entity on save and close screen #1725

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/1725


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/builder/DetailWindowBuilderProcessor.java
```diff
@@ -105,6 +105,10 @@ protected <E, V extends View<?>> void setupListDataComponent(DetailWindowBuilder
             return;
         }
 
+        if (detailView instanceof StandardDetailView<E> standardDetailView) {
```
glebfox, line 108: Probably, it's better moving `setReloadSaved` to `io.jmix.flowui.view.DetailView` interface, to make it more generic

glebfox (changes requested):

PR #2552, opened 2023-12-04
 ShowEntityInfoAction is applicable for several entities

Flamesson:
Ticket - https://github.com/jmix-framework/jmix/issues/2558


jmix-datatools/datatools-flowui/src/main/java/io/jmix/datatoolsflowui/action/ShowEntityInfoAction.java
```diff
@@ -117,4 +117,11 @@ public void showInfo(Object entity) {
         dialog.getView().setEntity(entity);
         dialog.open();
     }
+
+    @Override
+    protected boolean isApplicable() {
+        return super.isApplicable()
+                && target != null
```
glebfox, line 124: `target != null` is checked in the base class, so no need to duplicate

glebfox (changes requested):

Issue #2538, opened 2023-12-01
TimerFacet cannot be aded to a view with Content that doesn't implement HasComponents


glebfox:
### Environment

Jmix version: 2.1+

### Bug Description

Source: [forum](https://forum.jmix.io/t/how-to-implement-a-timer-facet-with-certain-logic-that-runs-on-all-screen/4032)

`TimerFacet`  uses `JmixTimer` component that is added to a DOM tree of a particular `View`. Currently, `JmixTimer` cannot be added to MainView because `JmixAppLayout` doesn't implement `HasComponents`

### Steps To Reproduce

Add Timer facet to the MainView

### Current Behavior

`TimerActionEvent` will never be fired. 

### Expected Behavior

`TimerActionEvent` is fired according to timer settings.

PR #2566, opened 2023-12-05
TimerFacet for views without HasComponent content #2538

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/2538

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/impl/TimerImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/impl/TimerImpl.java
@@ -17,10 +17,11 @@
 package io.jmix.flowui.facet.impl;
 
 import com.vaadin.flow.component.ComponentEventListener;
-import com.vaadin.flow.component.HasComponents;
+import com.vaadin.flow.component.dialog.Dialog;
 import com.vaadin.flow.shared.Registration;
-import io.jmix.flowui.kit.component.timer.JmixTimer;
+import io.jmix.flowui.component.UiComponentUtils;
 import io.jmix.flowui.facet.Timer;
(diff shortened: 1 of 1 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/impl/TimerImpl.java
```diff
@@ -104,46 +105,52 @@ public void setId(@Nullable String id) {
...
-                unregisterInView((View<? extends HasComponents>) this.owner);
+            if (this.owner != null) {
+                unregisterForView(this.owner);
             }
             super.setOwner(null);
         }
     }
 
-    protected void registerInView(View<? extends HasComponents> owner) {
+    protected void registerForView(View<?> owner) {
```
glebfox, line 117: Unnecessary API renaming, even though it's protected API.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/impl/TimerImpl.java
```diff
@@ -104,46 +105,52 @@ public void setId(@Nullable String id) {
...
-            registerOnAttach(owner);
+            addTimerOnAttach(owner);
         }
         addDetachListener(owner);
     }
 
-    protected void attachTimer(View<? extends HasComponents> owner) {
-        owner.getContent().add(timerImpl);
+    protected void addTimer(View<?> owner) {
+        Dialog dialog = UiComponentUtils.findDialog(owner);
```
glebfox, line 127: You don't need to add a timer to UI or Dialog. Actually, you still can add timer to a view, which is more correct, using the following code:
```
owner.getContent().getElement().appendChild(timerImpl.getElement());
```

In fact, `HasComponents` API do the same.

glebfox (changes requested):

Issue #1584, opened 2023-04-11
Implement MakeDefault action for genericFilter component


KremnevDmitry:
Wait for: https://github.com/jmix-framework/jmix/issues/1578

For QA:

Make default - implemented by `GenericFilterMakeDefaultAction` (`type="genericFilter_makeDefault"` in XML) which makes the filter configuration for this view for current user. The `genericFilter` will be automatically displayed on the filter panel when the view is opened.

Also check the "default for me" setting for the configuration detail view.

PR #2629, opened 2023-12-15
Implement MakeDefault action for genericFilter component jmix-framework/jmix#1584

KremnevDmitry:
See: #1584


jmix-flowui/flowui-data/src/main/java/io/jmix/flowuidata/genericfilter/UiDataGenericFilterSupport.java
```diff
@@ -244,7 +245,29 @@ protected void onUserRemove(UserRemovedEvent event) {
 
     protected boolean isDefaultForMeFieldVisible(Configuration currentConfiguration,
                                                  FilterConfiguration configurationModel) {
-        //TODO: kremnevda, viewSettingsFacet? 04.04.2023
+        View<?> currentView = UiComponentUtils.findView(currentConfiguration.getOwner());
+        if (currentConfiguration.getOwner().getId().isPresent() && currentView != null) {
```
glebfox, line 249: We need more nesting...

KremnevDmitry: We do what we do...

Issue #2574, opened 2023-12-07
Not a unique name in the Category attributes


rusiaikinat:
Jmix version: 2.1.999-SNAPSHOT

![image](https://github.com/jmix-framework/jmix/assets/114649487/9cccc10a-9ef9-4eca-98a9-c000cc75b8c3)

rusiaikinat:
![Image](https://github.com/jmix-framework/jmix/assets/114649487/e8f82e5e-a7cf-4407-89fa-04b1f32388fd)

PR #2632, opened 2023-12-15
fix Not a unique name in the Category attributes #2574

KartnDev:
#Solution
I decide that adding validation also for JPA level can break already existing settings, so validation works only as UI restriction

## Points
1. Added UI validation
2. JPA `unique = true` not applied on entity
3. Changelog also not updated

image


jmix-dynattr/dynattr-flowui/src/main/resources/io/jmix/dynattrflowui/messages.properties
```diff
@@ -41,10 +41,12 @@ io.jmix.dynattrflowui.view.category/localizationTab.title=Localization
 io.jmix.dynattrflowui.view.category/categoryAttrsGrid.defaultValue=Default value
 io.jmix.dynattrflowui.view.category/categoryAttrsGrid.currentDate=Current date
 io.jmix.dynattrflowui.view.category/categoryAttrsGrid.entityNotFound=Entity not found
-io.jmix.dynattrflowui.view.category/sourceGridTitle=Source attributes 
+io.jmix.dynattrflowui.view.category/sourceGridTitle=Source attributes
 io.jmix.dynattrflowui.view.category/targetColumnGridTitle=Column
 
 # CategoryAttrs Details
+io.jmix.dynattrflowui.view.categoryattr/notUniqueAttributeName=Name "%s" is not unique
```
gorbunkov, line 48: Perhaps, instead of **Name "%s" is not unique** it will be better to write **Attribute this name already exists in the category**. WDYT?

In the text **Name "abc" is not unique** I see the following problems:
* It's not clear where is this attribute is not unique: in the whole database, in this category only
* There is no need to duplicate the field value in the validation message - it is already displayed right in the text field.

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -312,6 +312,34 @@ protected void onAfterShow(BeforeShowEvent event) {
         loadTargetViews();
     }
 
+    @Install(to = "nameField", subject = "validator")
+    protected void nameFieldUniqueNameValidator(String value) {
+        validateUniqueStringOnAttribute(value, CategoryAttribute::getName, "notUniqueAttributeName");
```
gorbunkov, line 317: Could you please check using the debugger how much time this validator is fired on each textField value change? I have a suspicion that it is invoked more than once.

gorbunkov (changes requested):

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
KartnDev: 3 times invoking name validator (if code wasnt empty)
4 times invoking name validator + 1 time code validator (if code was empty)

KartnDev: UPD message
image

gorbunkov: > 4 times invoking name validator

Do you have an idea why this happens? Shouldn't it be called just once?

KartnDev: > > 4 times invoking name validator
> 
> Do you have an idea why this happens? Shouldn't it be called just once?

Here code `CategoryAttributesDetailView::refreshAttributesValues` that refrashing and triggering recalculations and because if it we have `isRefreshing` variable.

But i think its ok that validator invoking more than one time

KartnDev:
@gorbunkov can we merge it?

gorbunkov (approved):

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
KartnDev: @gorbunkov All fine, invoking only one time

PR #2637, opened 2023-12-15
Add require option for dynamic attributes

KartnDev:
image
image


jmix-dynattr/dynattr-flowui/src/main/resources/io/jmix/dynattrflowui/view/categoryattr/category-attributes-detail-view.xml
```diff
@@ -122,6 +122,7 @@
                                 <tooltip text="msg://lookupField.contextHelp"/>
                             </checkbox>
                             <checkbox id="isCollectionField" property="isCollection" visible="false"/>
+                            <checkbox id="isRequiredField" property="required" visible="false"/>
```
gorbunkov, line 125: is this attribute visible for all attribute types? If so then why do we set visible="false" and then write instruction like `FIELDS_VISIBLE_FOR_TYPES.put(BOOLEAN, "isRequiredField");` for all possible types? Can't we just make it always visible here?

gorbunkov (changes requested):

jmix-dynattr/dynattr-flowui/src/main/resources/io/jmix/dynattrflowui/view/categoryattr/category-attributes-detail-view.xml
KartnDev: No, this is not visible until we choose any of types. Also, this code following all UX rules presented inside dynamic attributes (but maybe require checkbox should be placed before `lookup` and `isCollection` field)

gorbunkov: @KartnDev I looked at how it was implemented in Classic UI:

Screenshot 2023-12-21 at 14 48 13

The "Required" checkbox was always visible, even before you selected any attribute type.
```xml
<form id="attributeForm" dataContainer="categoryAttributeDc" height="100%" width="100%">
                        <column width="100%">
                            <textField id="nameField" property="name" required="true"
                                       requiredMessage="msg://nameField.required"/>
                            <textField id="codeField" property="code" required="true"
                                       requiredMessage="msg://codeField.required"/>
                            <textArea id="descriptionField" property="description" maxLength="1000" rows="3"/>
                            <checkBox id="requiredField" property="required"/>
                            <comboBox id="dataTypeField" property="dataType" required="true"
                                         requiredMessage="msg://dataTypeField.required"/>
                            <sourceCodeEditor id="validationScriptField" property="configuration.validatorGroovyScript"
                                              mode="Groovy" highlightActiveLine="false" showGutter="false"
                                              height="220px"/>
                        </column>
                    </form>
```

What was the reason for moving the checkbox to another place and hiding it until the type is selected?

gorbunkov (approved):

Issue #2582, opened 2023-12-10
Support double click actions in DataGrid


glebfox:
### Feature Description

When double clicking on an item, the following should happen for List view:
1. Find an action with `Enter` shortcut
2. Find Edit action
3. Find Read action

In case of lookup view, double clicking on an item selects the item

glebfox:
## Solution

In addition to double click, the correct <kbd>Enter</kbd> press handling must be provided. In order to do this, the `SupportsEnterPress` interface is added. In case of `DataGrid`/`TreeDataGrid` the `setEnterPressHandler` method does the following:

Sets code to execute when <kbd>Enter</kbd> key is pressed. If such code is not set, this component responds to <kbd>Enter</kbd> press by attempting to find and execute the following actions:
* Action assigned to Enter key by setting its `KeyCombination`
* `io.jmix.flowui.action.list.EditAction`
* `io.jmix.flowui.action.list.ReadAction`

If one of these actions is found and enabled, it is executed.

**Note:** if no explicit double click listeners are added, then the above rule is used to handle double clicks on this component.

---

`StandardListView` sets `EnterPressHandler` if possible, As a result, both enter press and double click select the item(s) if any.

PR #2646, opened 2023-12-18
Support double click actions in DataGrid (2582)

glebfox:
See #2582


glebfox:
Will be implemented a little bit differently

PR #2676, opened 2023-12-22
late loading of entity combo boxes in filter

wwnjj13:
When opening the Filter Add Condition screen, every FilterCondition that used an EntityComboBox would load its data as part of the screen creation process. This was almost always an unnecessary data load. I therefore delayed the loading of data to the point at which the selected FilterConditions are actually added to the filter.

This has enormous performance impacts when projects seek to use exclusively EntityComboBoxes for Entity selection. Instead of loading data for potentially dozens of MetaPropertyPaths, it only loads data for the selected FilterCondition(s).


jmix-ui/ui-data/src/main/java/io/jmix/uidata/filter/FilterConfigurationConverter.java
```diff
@@ -1,76 +1,83 @@
...
+        Filter.Configuration configuration = new RunTimeConfiguration(id, logicalFilterComponent, filter);
+        configuration.setName(configurationModel.getName());
+        configuration.setAvailableForAllUsers(configurationModel.getUsername() == null);
+
+        for (FilterComponent filterComponent : logicalFilterComponent.getFilterComponents()) {
+            if (filterComponent instanceof SingleFilterComponent) {
+                configuration.setFilterComponentDefaultValue(((SingleFilterComponent<?>) filterComponent).getParameterName(),
+                        ((SingleFilterComponent<?>) filterComponent).getValue());
+
+                if (((SingleFilterComponent) filterComponent).getValueComponent() instanceof EntityComboBox) {
```
wwnjj13, line 74: Loading for a configuration

jmix-ui/ui/src/main/java/io/jmix/ui/action/filter/FilterAddConditionAction.java
```diff
@@ -1,220 +1,228 @@
...
+                    }
+
+                    FilterComponent filterComponent = converter.convertToComponent(selectedCondition);
+                    currentConfiguration.getRootLogicalFilterComponent().add(filterComponent);
+                    currentConfiguration.setFilterComponentModified(filterComponent, true);
+
+                    if(filterComponent instanceof SingleFilterComponent &&
+                            ((SingleFilterComponent)filterComponent).getValueComponent() instanceof EntityComboBox){
+                        EntityComboBox entityComboBox = (EntityComboBox) (((SingleFilterComponent)filterComponent).getValueComponent());
+                        entityComboBox.setOptions(new ContainerOptions(entityFieldCreationSupport.createCollectionContainer(entityComboBox.getMetaClass())));
```
wwnjj13, line 169: Loading for selected FilterConditions

jmix-ui/ui/src/main/java/io/jmix/ui/component/impl/EntityFieldCreationSupport.java
```diff
@@ -138,9 +138,7 @@ protected EntityPicker createFieldComponent(MetaClass metaClass, @Nullable Optio
 
         if (options != null || EntityComboBox.NAME.equals(componentName)) {
             EntityComboBox entityComboBox = uiComponents.create(EntityComboBox.class);
-            entityComboBox.setOptions(options != null ?
-                    options :
-                    new ContainerOptions(createCollectionContainer(metaClass)));
+            entityComboBox.setMetaClass(metaClass);
```
wwnjj13, line 141: Don't load, but set the metaClass for use later.

jmix-ui/ui/src/main/java/io/jmix/ui/component/impl/AbstractSingleFilterComponent.java
```diff
@@ -1,533 +1,544 @@
...
+        checkNotNullArgument(valueComponent);
+
+        if (this.valueComponent != null) {
+            root.remove(this.valueComponent);
+        }
+
+        this.valueComponent = valueComponent;
+        if (this.valueComponent instanceof EntityComboBox) {
+            EntityComboBox entityComboBox = (EntityComboBox) (this.valueComponent);
+            entityComboBox.setOptions(new ContainerOptions(entityFieldCreationSupport.createCollectionContainer(entityComboBox.getMetaClass())));
```
wwnjj13, line 315: Loads the options every time the new valueComponent is an EntityComboBox

glebfox:
`EntityFieldCreationSupport` is used in many places, not only by a property filter (e.g. Bulk edit, Input dialog, etc.), so you fix unfortunately makes new bugs. We agree that current implementation may lead to performance issues, so we'd highly appreciate if you create an issue and our developers will look into it.

PR #2680, opened 2023-12-22
fix Unique constraint violation is not handled correctly when using Iatest MS SQLServer

KartnDev:
image

### DIFF
--- a/jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/dbms/SqlServerDbmsFeatures.java
+++ b/jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/dbms/SqlServerDbmsFeatures.java
@@ -52,7 +52,7 @@ public class SqlServerDbmsFeatures implements DbmsFeatures {
 
     @Override
     public String getUniqueConstraintViolationPattern() {
-        return "with unique index \'(.+)\'";
+        return "(Violation of UNIQUE KEY constraint '(.+?)'|with unique index '(.+)')";
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/dbms/SqlServerDbmsFeatures.java
```diff
@@ -52,7 +52,7 @@ public String getTransactionTimeoutStatement() {
 
     @Override
     public String getUniqueConstraintViolationPattern() {
-        return "with unique index \'(.+)\'";
+        return "(Violation of UNIQUE KEY constraint '(.+?)'|with unique index '(.+)')";
```
gorbunkov, line 55: This pattern doesn't work as expected. 

Please read the issue description again and compare your result with the expected one:

> ER:
Notification "A department with the same name already exists"

See docs: https://docs.jmix.io/jmix/flow-ui/exception-handlers.html#unique-constraint-violation-handler

gorbunkov (changes requested):

KartnDev:
## Before (2 groups bug)
### Regex was:
`(Violation of UNIQUE KEY constraint '(.+?)'|with unique index '(.+)')`

* **If matching** `somewith sad sa as d with unique index 'ASDSAD_ASDSADSAD'`:
    * **1 group**: with unique index 'ASDSAD_ASDSADSAD'
    * **2 group**: ASDSAD_ASDSADSAD
* **If matching** `Violation of UNIQUE KEY constraint ASD_ASDSADS_ADSADASDSAD'`:
    * **1 group**: Violation of UNIQUE KEY constraint 'ASD_ASDSADS_ADSADASDSAD'
    * **2 group**: ASD_ASDSADS_ADSADASDSAD
    
image
## After (bracket group escape)
### Regex was:
`Violation of UNIQUE KEY constraint '(.+?)'|with unique index '(.+)'` => **REMOVED () root regex group**

* **If matching** `somewith sad sa as d with unique index 'ASDSAD_ASDSADSAD'`:
    * **1 group**: ASDSAD_ASDSADSAD
* **If matching** `Violation of UNIQUE KEY constraint ASD_ASDSADS_ADSADASDSAD'`:
    * **1 group**: ASD_ASDSADS_ADSADASDSAD

image
## Solved problem:
I created 2 groups, so internal component sees that the first group givin not-null string and interpret it as contstraint name

gorbunkov (approved):

Issue #2667, opened 2023-12-21
Improve ListMenu


yuriy-khrebtov:
Improve ListMenu:
- move expand button to the end of menu item
- support badges (suffix components)
- support setting `Component` as icon for menu items

glebfox:
> support setting Component as icon for menu items

Let's make it more generic, instead of setting icon, we will set prefix component. Same as in Vaadin SideNav

MaxKatyshev:
_Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT5973-233
IntelliJ version: IntelliJ IDEA 2023.3.2 (Community Edition)_
verified

PR #2683, opened 2023-12-25
Improve ListMenu #2667

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/2667


jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/main/ListMenu.java
```diff
@@ -617,6 +692,73 @@ public MenuItem withShortcutCombination(@Nullable KeyCombination shortcutCombina
...
+            this.suffixComponent = suffixComponent;
+            return this;
+        }
+
+        /**
+         * Sets suffix component of the item.
+         *
+         * @param suffixComponent component to set
+         */
+        public void setSuffixComponent(@Nullable Component suffixComponent) {
```
glebfox, line 719: Let's add `setter` methods for the other properties to make it consistent and more convenient, since it's hard to figure it out that there are `withXXX` methods.

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/main/ListMenu.java
```diff
@@ -347,27 +355,30 @@ protected Details createMenuBarComponent(MenuBarItem menuBarItem) {
         menuItemComponent.addClassName(JMIX_MENUBAR_ITEM_CLASS_NAME);
         menuItemComponent.addClassNames(menuBarItem.getClassNames().toArray(new String[0]));
         menuItemComponent.setOpened(menuBarItem.isOpened());
+        menuItemComponent.addThemeVariants(DetailsVariant.REVERSE);
```
glebfox, line 358: Let's make it component setting of the ListMenu component. Be default, keep at it now, so we don't unexpectedly change users' UI.

Create `ListMenuVariant` with a single constant (at least for now), implement `HasThemeVariant<ListMenuVariant>` and other xml, studio stuff. To make it simple, you don't need to set `reverse` theme name to nested details explicitly, just style them according to the parent element.

glebfox (changes requested):

jmix-flowui/flowui-themes/src/main/resources/META-INF/resources/themes/jmix-lumo/component/jmix-list-menu.css
```diff
@@ -44,19 +45,41 @@
...
 }
 
+.jmix-list-menu .jmix-menu-item-link .suffix-component,
+.jmix-list-menu .jmix-menubar-item .jmix-menubar-summary-icon-container .suffix-component {
+  margin-inline-start: auto;
+  flex-shrink: 0;
+}
+
+.jmix-list-menu .jmix-menubar-item .jmix-menubar-summary-icon-container > :last-child {
+  margin-right: var(--lumo-space-s);
```
glebfox, line 73: you can't use left/right variants, because they are not adapted to **rtl** mode. Use `inline-start/end` instead.

Please, fix all styles in this file, even existed

glebfox (changes requested):

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/main/ListMenu.java
yuriy-khrebtov: We decided not to implement additional setter methods now because common property setting logic (subscribtion to some property-change events) should be refactored first.

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/main/ListMenuVariant.java
```diff
@@ -0,0 +1,38 @@
...
+package io.jmix.flowui.kit.component.main;
+
+import com.vaadin.flow.component.shared.ThemeVariant;
+
+/**
+ * Set of theme variants applicable for the {@link ListMenu} component.
+ */
+public enum ListMenuVariant implements ThemeVariant {
+
+    TOGGLE_BUTTON_LAST("toggle-button-last");
```
glebfox, line 26: Let's name it similar to Vaddin Details component: `TOGGLE_REVERSE`. 
This is not a button because we can click on the item itself, and `reverse` because in case of `rtl` I'm not sure that it's last or first

JavaDoc is appreciated

glebfox (changes requested):

glebfox (approved):

PR #2751, opened 2024-01-16
Fix that the name in the generated Code field does not start with a capital letter

KartnDev:
I choose to concate category name and attribute name by **_** instead of trying to predict which case user chose for name:
samples:
1. How to concate when category name is FULLUPPERCASE or fULLCASEWITHUNDERCUT
2. Same as first but for attribute name

Besides, we can do not change names and just concate both with "_" and we will prevent any misunderstood cases

Advantages:
1. We will not take any responsibility of name changing
2. More understandable name-code (if names contains more then one word / contains same words => mass of camelcase)
3. Real word split and user always sees what category he uses (+SomeCategory_SomeName)

image
image

### DIFF
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
@@ -80,6 +80,7 @@ import io.jmix.flowui.sys.ViewSupport;
 import io.jmix.flowui.view.*;
 import io.jmix.flowui.view.builder.LookupWindowBuilder;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.CaseUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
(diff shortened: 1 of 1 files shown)


gorbunkov:
@KartnDev 

> How to concate when category name is FULLUPPERCASE or fULLCASEWITHUNDERCUT

First of all, let's stick to real-life cases. 99.9% of users won't write a category or attribute name this way. They will do smth like:
* Some nice category
* some nice category
* someNiceCategory
* SomeNiceCategory
* some_nice_category (also possible)

The same for attribute name.

I agree with your arguments about readability, but on the other hand I would prefer that the code of the dynamic attribute looks like any other entity attribute name (we see them in metamodel, in REST API, etc.). We usually start entity attribute names with the lowercase and use camelCase. If dynamic attributes looked the same they, that would be the best choice, in my opinion:

Car
* vin
* productionDate
* +someCarCategorySomeAttributeName

So, what I would do while building an attribute code:

* convert both a category and a name to the camel case (e.g. by using `org.apache.commons.text.CaseUtils#toCamelCase`)
* start category name with the lowercase letter 
* start attribute name with the uppercase letter
* concatenate them

"Some car category"."some attribute" -> "someCarCategorySomeAttribute"

gorbunkov (changes requested):
See my comment https://github.com/jmix-framework/jmix/pull/2751#issuecomment-1898019429
WDYT?

gorbunkov (changes requested):
Could you please re-read my comment again and try the example that is written there:

```
"Some car category"."some attribute" -> "someCarCategorySomeAttribute"
```

Your solution works differently

KartnDev:
> some attribute

image
image

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -890,7 +891,17 @@ protected void refreshCodeFieldValue() {
...
                 categoryName = StringUtils.defaultString(attribute.getCategory().getName());
             }
-            codeField.setValue(StringUtils.deleteWhitespace(categoryName + attribute.getName()));
+            char[] delimiters = {' ', '.', '_', '-', '\t'};
+
+            String categoryNameInCamelCaseUncapitalized = CaseUtils.toCamelCase(categoryName, false, delimiters);
+            String attributeNameInCamelCaseUncapitalized = CaseUtils.toCamelCase(attribute.getName(), false, delimiters);
+
+            // `org.springframework.util.StringUtils::hasText` simplifies expression `isNotEmpty && isNotBlank` into one `hasText`
+            String resultCodeName = org.springframework.util.StringUtils.hasText(categoryNameInCamelCaseUncapitalized) ?
```
gorbunkov, line 900: In framework code we mostly use Guava `com.google.common.base.Strings#isNullOrEmpty` for that

gorbunkov (approved):

Issue #2221, opened 2023-09-11
Grid selection with versioned Entity


ig0rix:
Hi, everyone!

Jmix 2.0.2

Grid items operations raises OptimisticLockingException  if modified entity is versioned.
It happens because grid selection contains old version of entity.
Steps for reproduce ( for Full-Stack Application(Java) project template ):
**Step 0:** 
Modify user-list-view.xml - configure dialog mode for list_edit action like below:
`
                <action id="edit" type="list_edit">
                    <properties>
                        <property name="openMode" value="DIALOG"/>
                    </properties>
                </action>
`
Run application and go to Users page

**Step 1:** 
Select user and click Edit Button -> Change user time zone in opened window and Save changes
**Step 2:**
After closing dialog window click Remove Button -> Error message has been raised (Unable to save the 'User' object changes)

Error screenshot:
![GridSelectionBug](https://github.com/jmix-framework/jmix/assets/2056403/1be2dd29-506a-443c-ac3b-967945fe6f1b)

Project sample:
[GridSelectionBug.zip](https://github.com/jmix-framework/jmix/files/12575685/GridSelectionBug.zip)

Issue #2642, opened 2023-12-18
Synchronize the Vaadin's DataGrid selection model and visual representation of data


KremnevDmitry:
See forum topic: https://forum.jmix.io/t/v2-1-datagrid-multiselect-issue/4144

Changing the visibility of table elements doesn't affect the selection model. This behaviour is misleading.

yuriy-khrebtov:
Fixed in https://github.com/jmix-framework/jmix/issues/2221

PR #2753, opened 2024-01-17
Grid selection with versioned Entity #2221

yuriy-khrebtov:
See https://github.com/jmix-framework/jmix/issues/2221

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractGridDelegate.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractGridDelegate.java
@@ -179,6 +179,8 @@ public abstract class AbstractGridDelegate<C extends Grid<E> & ListDataComponent
         closeEditorIfOpened();
         component.getDataCommunicator().reset();
         updateAggregationRow();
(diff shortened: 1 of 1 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractGridDelegate.java
```diff
@@ -197,6 +199,28 @@ && getComponent().getEditor().isOpen()) {
...
+            if (prevSelectedItemsToRefresh.remove(item)) {
+                itemsToSelect.add(item);
+            }
+
+            //skip further checks if no more items were selected
+            if (prevSelectedItemsToRefresh.isEmpty()) {
+                break;
+            }
+        }
+        deselectAll();
```
glebfox, line 220: let's add a detailed description why we do this, because it will cause in several selection events that may cause problems if users use selection event to load related data

glebfox (changes requested):

glebfox:
Also check of your fix has effect on https://github.com/jmix-framework/jmix/issues/2642

glebfox (approved):

PR #2773, opened 2024-01-18
Add an understandable description of errors in the Rest API (2605)

NikitaShchienko:


NikitaShchienko:
It may break custom clients since we are changing the status code from 500 to 400.

glebfox (approved):

gorbunkov (approved):

PR #2783, opened 2024-01-21
Add DataManager.saveAll(Collection) method

knstvk:


jmix-core/core/src/main/java/io/jmix/core/UnconstrainedDataManager.java
```diff
@@ -76,6 +78,14 @@ public interface UnconstrainedDataManager {
      */
     EntitySet save(Object... entities);
 
+    /**
+     * Saves collection of entities to their data stores.
```
gorbunkov, line 82: Saves **a** collection of entities - the article is missing

gorbunkov (approved):

Issue #2788, opened 2024-01-23
Uniform handling of JPA and DTO entities in standard list/detail views


knstvk:
Currently due to #1330, #1357 and #1401 the create/edit actions and standard detail view behave differently for DTO entities: 
- The entity ID is not passed in URL route, which makes impossible to reload the entity in detail view as it is done for JPA. 
- The standard template for DTO detail view has ugly override of `findEntityId()` method which normally extracts ID from URL.

To simplify work with DTO entities loaded from external data sources the UI framework should make no difference between JPA and DTO entities. The only thing the developer should do for DTO is to define a load delegate. 

The mechanism of working with in-memory-only DTO entities like security models needs to be updated accordingly.

knstvk:
## Added methods

- `EntityStates.setNew(Object entity, boolean isNew)` - manages the "New" state of the entity instance.
- `RemoveAction.setDelegate()`, `RemoveOperation.RemoveBuilder.withRemoveDelegate()` - delegate to be invoked instead of DataManager to remove the entities from a storage
- `StandardDetailView.setReloadEdited(boolean)` - whether the edited entity should be reloaded before setting to the data container. True by default.

## Removed methods

- Protected `StandardDetailView.entityCanBeLoaded()`
- Protected `DetailViewNavigationProcessor.isNeedToSetEntityToEdit()`
- Protected `DetailViewNavigationProcessor.getEntityToEdit()`

## Behavior changes

The framework now doesn't make any difference between JPA and DTO entities when navigating to a detail view: it passes the entity ID in the route parameter. The detail view for DTO entity is supposed to get this ID and load the entity instance from some data storage using the load delegate. If the `new` constant is passed instead of ID, the view creates a new instance.

If the whole entity instance is passed instead of ID (e.g. when opening in dialog window), `EntityStates.isNew()` is used to distinguish between Edit and Create mode. Consequently, it's important to **set the entity in the not-new state after loading** it from a storage. For a DTO entity it can be done using the new `EntityStates.setNew()` method, for JPA entity it's done by the standard JPA data store implementation.

If the edited entity should not be reloaded from the data storage before setting to the data container, call `setReloadEdited(false)` in the detail view constructor or `InitEvent` handler. This is the case for DTO entities existing purely in memory and not mapped directly to external data, like Security or JMX Console model entities.

## Recommendations

### Opening DTO entity detail view

[...]

PR #2819, opened 2024-01-31
Uniform handling of JPA and DTO entities in standard list/detail views

knstvk:
See #2788 for the problem description.

## Added methods

- `EntityStates.setNew(Object entity, boolean isNew)` - manages the "New" state of the entity instance.
- `RemoveAction.setDelegate()`, `RemoveOperation.RemoveBuilder.withRemoveDelegate()` - delegate to be invoked instead of DataManager to remove the entities from a storage
- `StandardDetailView.setReloadEdited(boolean)` - whether the edited entity should be reloaded before setting to the data container. True by default.

## Removed methods

- Protected `StandardDetailView.entityCanBeLoaded()`
- Protected `DetailViewNavigationProcessor.isNeedToSetEntityToEdit()`
- Protected `DetailViewNavigationProcessor.getEntityToEdit()`

## Behavior changes

The framework now doesn't make any difference between JPA and DTO entities when navigating to a detail view: it passes the entity ID in the route parameter. The detail view for DTO entity is supposed to get this ID and load the entity instance from some data storage using the load delegate. If the `new` constant is passed instead of ID, the view creates a new instance.

If the whole entity instance is passed instead of ID (e.g. when opening in dialog window), `EntityStates.isNew()` is used to distinguish between Edit and Create mode. Consequently, it's important to **set the entity in the not-new state after loading** it from a storage. For a DTO entity it can be done using the new `EntityStates.setNew()` method, for JPA entity it's done by the standard JPA data store implementation.

If the edited entity should not be reloaded from the data storage before setting to the data container, call `setReloadEdited(false)` in the detail view constructor or `InitEvent` handler. This is the case for DTO entities existing purely in memory and not mapped directly to external data, like Security or JMX Console model entities.

### Opening DTO entity detail view

Previously, the framework automatically adjusted the process of opening detail views for DTO entities. In particular, when navigating, it didn't provide the entity ID in route parameters and instead passed the entity instance in `AfterViewNavigationEvent` handler.

If you have used this standard approach for navigation and want to keep it, you should replicate the old behavior in create/edit actions of the list view as in the following example for the `Phone` DTO entity:

```
@Autowired  
private ViewNavigators viewNavigators;
@ViewComponent  
private DataGrid<Phone> phonesDataGrid;

@Subscribe("phonesDataGrid.create")  
public void onPhonesDataGridCreate(final ActionPerformedEvent event) {  
    viewNavigators.detailView(phonesDataGrid)  
            .withViewClass(PhoneDetailView.class)  
            .withRouteParameters(RouteParameters.empty())  
            .withAfterNavigationHandler(afterViewNavigationEvent -> {  
                Phone phone = metadata.create(Phone.class);  
                afterViewNavigationEvent.getView().setEntityToEdit(phone);  
            })  
            .navigate();  
}  
  
@Subscribe("phonesDataGrid.edit")  
public void onPhonesDataGridEdit(final ActionPerformedEvent event) {  
    viewNavigators.detailView(phonesDataGrid)  
            .withViewClass(PhoneDetailView.class)  
            .withRouteParameters(RouteParameters.empty())  
            .withAfterNavigationHandler(afterViewNavigationEvent ->  
                    afterViewNavigationEvent.getView()  
                            .setEntityToEdit(Objects.requireNonNull(phonesDataGrid.getSingleSelectedItem())))  
            .navigate();  
}
```

As you can see, the action handler methods set the empty route parameters and pass the entity in `AfterViewNavigationEvent` handler.

Opening detail view in dialog window should work as before without changes in list view actions.

Both for the navigation and dialog mode, call `setReloadEdited(false)` in the detail view:
```
public PhoneDetailView() {  
    setReloadEdited(false);  
}
```

### Setting non-new state for loaded DTO entities

To make sure the framework correctly distinguish new instances from already existing in a storage, call `EntityStates.setNew(entity, false)` for entities loaded from external storage.

### Assigning ID to new DTO entities

If the new entity ID is assigned by the storage, set the ID to the original instance too to let the framework match the saved instance with the original one. For example:

```
public Task saveTask(Task task) {
    String url = task.getId() != null ? TASKS_BASE_URL + "/"  + task.getId() : TASKS_BASE_URL;
    ResponseEntity<Task> response = restTemplate.postForEntity(url, task, Task.class);
    Task savedTask = Objects.requireNonNull(response.getBody());
    if (task.getId() == null) {
        task.setId(savedTask.getId());
    }
    entityStates.setNew(savedTask, false);
    return savedTask;
}
```

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
@@ -97,16 +97,6 @@ public class DetailViewNavigationProcessor extends AbstractNavigationProcessor<D
         }
     }
 
-    protected Object getEntityToEdit(DetailViewNavigator<?> navigator) {
-        return switch (navigator.getMode()) {
-            case CREATE -> metadata.create(navigator.getEntityClass());
-            case EDIT -> navigator.getEditedEntity().orElseThrow(() ->
-                    new IllegalStateException(String.format(
-                            "Detail View of %s cannot be open with mode EDIT, entity is not set",
-                            navigator.getEntityClass())));
-        };
-    }
-
     protected RouteParameters generateNewEntityRouteParameters(DetailViewNavigator<?> navigator) {
         return routeSupport.createRouteParameters("id", StandardDetailView.NEW_ENTITY_ID);
     }
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
@@ -23,7 +23,6 @@ import io.jmix.core.MetadataTools;
 import io.jmix.core.annotation.Internal;
 import io.jmix.core.entity.EntityValues;
 import io.jmix.flowui.sys.ViewSupport;
-import io.jmix.flowui.view.DetailView;
 import io.jmix.flowui.view.DetailViewMode;
 import io.jmix.flowui.view.StandardDetailView;
 import io.jmix.flowui.view.View;
@@ -83,14 +82,10 @@ public class DetailViewNavigationProcessor extends AbstractNavigationProcessor<D
     @Override
     protected RouteParameters getRouteParameters(DetailViewNavigator<?> navigator) {
         return navigator.getRouteParameters().orElseGet(() -> {
-            if (metadataTools.isJpaEntity(navigator.getEntityClass())) {
-                return switch (navigator.getMode()) {
-                    case CREATE -> generateNewEntityRouteParameters(navigator);
-                    case EDIT -> generateEditEntityRouteParameters(navigator);
-                };
-            } else {
-                return RouteParameters.empty();
-            }
+            return switch (navigator.getMode()) {
+                case CREATE -> generateNewEntityRouteParameters(navigator);
+                case EDIT -> generateEditEntityRouteParameters(navigator);
+            };
         });
     }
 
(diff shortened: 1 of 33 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
```diff
@@ -99,18 +94,9 @@ protected void fireAfterViewNavigation(DetailViewNavigator<?> navigator, View<?>
...
         }
     }
 
-    protected boolean isNeedToSetEntityToEdit(DetailViewNavigator<?> navigator, View<?> view) {
-        return !metadataTools.isJpaEntity(navigator.getEntityClass())
-                && view instanceof DetailView
-                && navigator.getRouteParameters().isEmpty();
-    }
-
     protected Object getEntityToEdit(DetailViewNavigator<?> navigator) {
```
glebfox, line 100: Unused. Seems that can be removed

glebfox (approved):

jmix-templates/content/flowui/list-dto/controller.java
```diff
@@ -21,7 +21,14 @@ public class ${viewControllerName} extends StandardListView<${entity.className}>
 
     @Install(to = "${tableDl}", target = Target.DATA_LOADER)
     protected List<${entity.className}> ${tableDl}LoadDelegate(LoadContext<${entity.className}> loadContext) {
-        // Here you can load entities from an external store
-        return Collections.emptyList();
+        // Here you can load entities from an external storage
```
gorbunkov, line 24: It is worth mentioning here that loaded DTO entities must be "not new", i.e. EntityStates.setNew(entity, false) must be invoked for them.

gorbunkov (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/navigation/DetailViewNavigationProcessor.java
knstvk: done

jmix-templates/content/flowui/list-dto/controller.java
knstvk: done

jmix-templates/content/flowui/list-dto/controller.java
```diff
@@ -21,7 +21,14 @@ public class ${viewControllerName} extends StandardListView<${entity.className}>
...
+        return List.of();
+    }
+
+    @Install(to = "${tableId}.remove", subject = "delegate")
+    private void ${tableId}RemoveDelegate(final Collection<${entity.className}> collection) {
```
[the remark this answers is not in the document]
knstvk: done

PR #2844, opened 2024-02-05
added context support for groovy formatter 2587

syncro:


jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/GroovyScriptParametersProvider.java
```diff
@@ -41,4 +41,6 @@ public interface GroovyScriptParametersProvider {
      * @return map of objects
      */
     Map<String, Object> getParametersForValidationParameters();
+
```
NikitaShchienko, line 44: Javadoc

NikitaShchienko (approved):

jmix-reports/reports/src/main/java/io/jmix/reports/libintegration/GroovyScriptParametersProvider.java
syncro: added

PR #2848, opened 2024-02-06
Skip enhancing of entities if entities weren't changed

gorbunkov:
### Implementation Details

During entities enhancement, the Jmix Gradle plugin evaluates the checksum of enhanced entities classes and saves it to a file in the temporary directory. When next compilation is executed, the plugin checks that entity classes inside the project build directory are the same as the saved ones (have the same checksum). There will be difference if any entity class is changed or a new entity is created. In this case the build directory will contain a modified (non-enhanced) entity class. If such differences are found then entity enhancing step will be run.

To disable this behavior and always enhance all entities on each compilation the following config should be added to the build.gradle:

```
jmix {
    entitiesEnhancing {
        skipUnmodifiedEntitiesEnhancing = false
    }
}
```

By default, the entities enhancing skipping functionality is enabled.

### DIFF
--- a/jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/EnhancingAction.groovy
+++ b/jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/EnhancingAction.groovy
@@ -26,6 +26,10 @@ import org.gradle.api.Task
 import org.gradle.api.file.FileCollection
 import org.gradle.api.tasks.SourceSet
 
+import java.nio.file.Files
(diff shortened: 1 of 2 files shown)


knstvk:
LGTM, but one question: why not just store a single hash for all previous files combined instead of copying and storing the files themselves?

gorbunkov:
> LGTM, but one question: why not just store a single hash for all previous files combined instead of copying and storing the files themselves?

@knstvk Good catch! Fixed that.

PR #2855, opened 2024-02-07
Field validator is invoked 5 times on each value change (2821)

Flaurite:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/upload/FileUploadField.java
```diff
@@ -113,7 +113,6 @@ public void setValueSource(@Nullable ValueSource<byte[]> valueSource) {
 
     @Override
     public boolean isInvalid() {
-        validate();
```
Flaurite, line 116: According to other fields `isInvalid()` must not trigger validation. But probably it can be breaking changes.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractFieldDelegate.java
```diff
@@ -234,8 +234,6 @@ public void setInvalid(boolean invalid) {
 
     public void setConversionInvalid(boolean conversionInvalid) {
         this.conversionInvalid = conversionInvalid;
-
-        updateInvalidState();
```
glebfox, line 238: check that https://github.com/jmix-framework/jmix/issues/1456 still works

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/upload/FileUploadField.java
glebfox: mention this in the issue description

glebfox (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractFieldDelegate.java
Flaurite: Missed one case from #1456 the issue. Now it is fixed

Issue #2706, opened 2023-12-29
Excel Export throws exception for the Float type.


glebfox:
### Environment

Jmix version: 1.5.5

### Bug Description

Excel Export throws exception for the `Float` type.

### Steps To Reproduce

* Add `float` attribute to an entity
* Add `<action id="excelExport" type="excelExport"/>` to the browser screen (Grid Export Actions add-on)

### Current Behavior

```
java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.math.BigDecimal (java.lang.Double and java.math.BigDecimal are in module java.base of loader 'bootstrap')
	at io.jmix.core.metamodel.datatype.impl.FloatDatatype.parse(FloatDatatype.java:99)
	at io.jmix.core.metamodel.datatype.impl.FloatDatatype.parse(FloatDatatype.java:77)
	at io.jmix.core.metamodel.datatype.impl.FloatDatatype.parse(FloatDatatype.java:35)
	at io.jmix.gridexportui.exporter.excel.ExcelExporter.formatValueCell(ExcelExporter.java:713)
	at io.jmix.gridexportui.exporter.excel.ExcelExporter.createRowForEntityInstance(ExcelExporter.java:656)
	at io.jmix.gridexportui.exporter.excel.ExcelExporter.lambda$exportTable$1(ExcelExporter.java:248)
	at io.jmix.gridexportui.exporter.excel.AllRecordsExporter.lambda$exportAll$0(AllRecordsExporter.java:108)
```

### Expected Behavior

No exception

---

Source: [forum](https://forum.jmix.ru/t/oshibka-v-excelexport-s-atributami-tipa-float/1542)

ilushaaa:
Works fine in 2.1

ilushaaa:
https://github.com/jmix-framework/jmix/commit/f9d1d86a42cf53d404f1b724829da350b7605d74

SergeiAksenov2:
Tested on:
Jmix version: 1.5.999-SNAPSHOT
Jmix Studio plugin version: 2.1.3-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Ultimate Edition)

PR #2864, opened 2024-02-08
Excel export throws exception for the float data type 2706

ilushaaa:
#2706


jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/exporter/excel/ExcelExporter.java
```diff
@@ -430,7 +437,7 @@ protected void formatValueCell(Cell cell, @Nullable Object cellValue, @Nullable
 
             cell.setCellValue(date);
 
-            if (Objects.equals(java.sql.Time.class, javaClass)) {
+            if (Objects.equals(Time.class, javaClass)) {
```
glebfox, line 440: We don't change code that is not related to the current issue.

glebfox (changes requested):

glebfox (approved):

Issue #2713, opened 2024-01-09
@Value annotation in screen controllers supports only String


glebfox:
### Environment

Jmix version: 1.5.5

### Bug Description

A description of what the bug is.

### Steps To Reproduce

1. Define a custom int app property in `application.properties`, e.g.: `app.test-int=10`
2. Inject it in a screen using `@Value`, e.g.: 
```
@Value("${app.test-int}")
private Integer appTestInt;
```

### Current Behavior

```
Caused by: java.lang.IllegalArgumentException: Can not set java.lang.Integer field com.company.demo.screen.sandbox.SandboxScreen.appTestInt to java.lang.String

	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171) ~[na:na]
	at java.base/jdk.internal.reflect.UnsafeObjectFieldAccessorImpl.set(UnsafeObjectFieldAccessorImpl.java:81) ~[na:na]
	at java.base/java.lang.reflect.Field.set(Field.java:799) ~[na:na]
	at io.jmix.ui.sys.UiControllerDependencyInjector.assignValue(UiControllerDependencyInjector.java:770) ~[jmix-ui-1.5.5.jar:na]
	at io.jmix.ui.sys.UiControllerDependencyInjector.doInjection(UiControllerDependencyInjector.java:474) ~[jmix-ui-1.5.5.jar:na]
	at io.jmix.ui.sys.UiControllerDependencyInjector.injectValues(UiControllerDependencyInjector.java:262) ~[jmix-ui-1.5.5.jar:na]
	at io.jmix.ui.sys.UiControllerDependencyInjector.inject(UiControllerDependencyInjector.java:99) ~[jmix-ui-1.5.5.jar:na]
	at io.jmix.ui.sys.UiControllerDependencyManager.inject(UiControllerDependencyManager.java:61) ~[jmix-ui-1.5.5.jar:na]
	at io.jmix.ui.sys.ScreensImpl.createScreen(ScreensImpl.java:212) ~[jmix-ui-1.5.5.jar:na]
```

Hint: `UiControllerDependencyInjector#getInjectedInstance` uses `AutowireCapableBeanFactory#resolveEmbeddedValue` for `@Value` annotation and it is always returns `String`.

### Expected Behavior

All value types supported by `@Value` can be used. 

---

[...]

PR #2867, opened 2024-02-09
Value annotation in screen controllers supports only string 2713

ilushaaa:
#2713

Now supports any types for fields with the value annotation (Integer, Double etc)

### DIFF
--- a/jmix-ui/ui/src/main/java/io/jmix/ui/sys/UiControllerDependencyInjector.java
+++ b/jmix-ui/ui/src/main/java/io/jmix/ui/sys/UiControllerDependencyInjector.java
@@ -570,8 +570,9 @@ public class UiControllerDependencyInjector {
 
         } else if (annotationClass == Value.class) {
             AutowireCapableBeanFactory autowireCapableBeanFactory = applicationContext.getAutowireCapableBeanFactory();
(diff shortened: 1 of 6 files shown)


jmix-ui/ui/src/main/java/io/jmix/ui/sys/UiControllerDependencyInjector.java
```diff
@@ -570,8 +572,14 @@ protected Object getInjectedInstance(Class<?> type, String name, InjectElement i
 
         } else if (annotationClass == Value.class) {
             AutowireCapableBeanFactory autowireCapableBeanFactory = applicationContext.getAutowireCapableBeanFactory();
-            return ((AbstractBeanFactory) autowireCapableBeanFactory)
+            String stringValue = ((AbstractBeanFactory) autowireCapableBeanFactory)
                     .resolveEmbeddedValue(element.getAnnotation(Value.class).value());
+            ConversionService conversionService = ((AbstractBeanFactory) autowireCapableBeanFactory)
```
glebfox, line 577: Let's add test like a screen with String and Integer `@Value` fields

glebfox (changes requested):

PR #2869, opened 2024-02-09
add dynAttr api on attribute panel and add api ability for category field visibility

KartnDev:


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/panel/DynamicAttributesPanel.java
```diff
@@ -16,275 +16,38 @@
...
-import io.jmix.flowui.view.ViewValidation;
 
 import javax.annotation.Nullable;
-import java.util.*;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.List;
 
-public class DynamicAttributesPanel extends Composite<VerticalLayout> implements HasSize {
+public interface DynamicAttributesPanel extends HasSize, HasComponents {
```
gorbunkov, line 31: Most of Jmix components don't have an interface. I don't see reasons to have it for dyn attrs panel

gorbunkov (changes requested):

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/panel/DynamicAttributesPanel.java
```diff
@@ -202,20 +212,7 @@ protected Component generateFieldComponent(AttributeDefinition attribute) {
...
         componentContext.setValueSource(valueSource);
 
-        Component resultComponent = uiComponentsGenerator.generate(componentContext);
-
-        setWidth(resultComponent, attribute);
-
-        return resultComponent;
-    }
-
-    protected void setWidth(Component component, AttributeDefinition attribute) {
```
KartnDev, line 212: i prefer to not configure UI component firstly from "dynattr hot settings". Instead of it it should has behave like normal UI component and configure only from java code or from xml
`else {
            ((HasSize) component).setWidth(fieldWidth);
        }`

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/panel/DynamicAttributesPanel.java
```diff
@@ -256,35 +243,119 @@ protected void onInstanceContainerItemChangeEvent(InstanceContainer.ItemChangeEv
...
+        return dynamicAttributesForm;
+    }
+
+    /**
+     * Retrieves the list of fields in the dynamic attribute panel's form.
      *
-     * @param fieldWidth width of the fields
+     * @return The list of dynamic attributes form fields
      */
-    public void setFieldWidth(String fieldWidth) {
```
KartnDev, line 287: Field width cannot be set

gorbunkov (changes requested):
For this PR please do only one thing the issue asks: add a method for setting a visibility for category field. For all other public API that you've added, please, create a separate issue with a description of what API you want and why this API will be useful. We will consider this for future releases.

PR #2873, opened 2024-02-09
Condition improvements: support null params, support by Jmix Data Repositories

dtaimanov:
jmix-framework/jmix 2490


jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
```diff
@@ -99,6 +99,13 @@ public class CoreProperties {
      */
     boolean roundDecimalValueByFormat;
 
+    /**
+     *
+     * Whether JpqlCondition and Property condition should be skipped by default if property value is null or empty
```
gorbunkov, line 104: Write condition names using a single style: either class names (**JpqlCondition** and **PropertyCondition**) or types using separate words (**JPQL condition** and **property condition**)

jmix-core/core/src/main/java/io/jmix/core/impl/repository/query/JmixParameters.java
```diff
@@ -17,27 +17,31 @@
...
 import java.lang.reflect.Method;
 import java.util.List;
+
 /**
  * {@link Parameters} extension required to support {@link io.jmix.core.FetchPlan} special parameter.
  */
 public class JmixParameters extends Parameters<JmixParameters, JmixParameter> {
 
     private final int fetchPlanIndex;
+    private final int jmixArgsIndex;
```
gorbunkov, line 33: The field name and related methods should probably be renamed (jmixDataRepositoryContextIndex, not jmixArgIndex)

jmix-core/core/src/main/java/io/jmix/core/impl/repository/query/JmixCustomLoadQuery.java
```diff
@@ -85,15 +89,19 @@ public JmixCustomLoadQuery(DataManager dataManager, Metadata jmixMetadata, Metho
     @Override
     public Object execute(Object[] parameters) {
         UnconstrainedDataManager dataManager = getDataManager();
-
+        //todo taimanov: rework through LoadContext
```
gorbunkov, line 92: Why this todo is here? Should the code below be reworked?

jmix-core/core/src/main/java/io/jmix/core/impl/repository/query/JmixCustomLoadQuery.java
```diff
@@ -102,10 +110,24 @@ public Object execute(Object[] parameters) {
...
+
+        if (jmixDataRepositoryContext != null) {
+            query.hints(jmixDataRepositoryContext.hints());
+
+            if (jmixDataRepositoryContext.condition() != null) {
+                 conditionQuery = query.condition(jmixDataRepositoryContext.condition());
+            }
+        }
+
+        if(conditionQuery==null){
```
gorbunkov, line 123: Pls, format the if statement properly (add required spaces)

jmix-core/core/src/main/java/io/jmix/core/impl/repository/query/JmixListQuery.java
```diff
@@ -44,17 +46,30 @@ public JmixListQuery(DataManager dataManager, Metadata jmixMetadata, Method meth
 
     @Override
     public Object execute(Object[] parameters) {
+        JmixDataRepositoryContext jmixDataRepositoryContext = jmixContextIndex == -1 ? null :
+                (JmixDataRepositoryContext) parameters[jmixContextIndex];
+
+        if (jmixDataRepositoryContext != null && jmixDataRepositoryContext.condition() != null) {
+            conditions = LogicalCondition.and(conditions, jmixDataRepositoryContext.condition());
+        }
+        //todo taimanov: use LoadContext
```
gorbunkov, line 55: As in another class, the code should be reworked, as todo says

jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/action/AssignToUsersAction.java
```diff
@@ -156,13 +156,13 @@ protected void openDialog() {
 
     protected void configureViewLoader(DataLoader loader) {
         List<String> excludedUsernames = dataManager.load(RoleAssignmentEntity.class)
-                .condition(PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode()))
+                .condition(UIConditions.equal(ROLE_CODE_PROPERTY, selectedItem.getCode()))
```
gorbunkov, line 159: Why do we need to create a separate class UIConditions? Why not to simply add the `.skipNullOrEmpty()` to existing PropertyCondition builder invocation? E.g. `PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode())` -> `PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode()).skipNullOrEmpty()`. What are the benefits of using a new class that fully repeats condition builders?

jmix-core/core/src/main/java/io/jmix/core/repository/JmixDataRepositoryContext.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.lang.Nullable;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Map;
+
+
+
+/**
+ * @param fetchPlan {@link FetchPlan} to load entities with. Default value: {@link FetchPlan#BASE}
```
gorbunkov, line 32: Pls, add the general class description to the javadoc

gorbunkov (changes requested):

jmix-core/core/src/main/java/io/jmix/core/impl/repository/query/JmixCustomLoadQuery.java
dtaimanov: It is just refactoring do not affecting something. I will remove todo and make it during further work.

jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/action/AssignToUsersAction.java
dtaimanov: First of all it improves code readability, especially when condition created in the middle of complex code, e.g.
this part of code:
```java
 List<String> excludedUsernames = dataManager.load(RoleAssignmentEntity.class)
                .condition(UIConditions.equal(ROLE_CODE_PROPERTY, selectedItem.getCode()))
                .list()
                .stream()
                .map(RoleAssignmentEntity::getUsername)
                .collect(Collectors.toUnmodifiableList());
```
looks better than this:
```java
 List<String> excludedUsernames = dataManager.load(RoleAssignmentEntity.class)
                .condition(PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode())
                        .skipNullOrEmpty())
                .list()
                .stream()
                .map(RoleAssignmentEntity::getUsername)
                .collect(Collectors.toUnmodifiableList());
```
or:
```java
  return UIConditions.jpqlCondition("e.parent = :parent", null, ParamsMap.of("parent", parentDocument));
```
vs.
```java
  return JpqlCondition.jpqlCondition("e.parent = :parent", null, ParamsMap.of("parent", parentDocument))
                    .skipNullOrEmpty();
```
also it allows to preserve space and write some statements in one line instead of splitting it.

Second reason: because of code simplification and atomicity it is more error-prone during code refactoring and development. Otherwise I afraid that `.skipNullOrEmpty()` part may be lost or forgotten.

gorbunkov: For me, "code readability" is when you look at the code and understand what it does, not when it takes less space on the screen. 

When I look at:

```
PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode())
```

vs

```
PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode()).skipNullOrEmpty()
```
then I immediately understand the difference between two methods.

When I look at

```
PropertyCondition.equal(ROLE_CODE_PROPERTY, selectedItem.getCode())
```

vs 

```
UIConditions.equal(ROLE_CODE_PROPERTY, selectedItem.getCode())
```

then for me it's harder to get the difference.

Also, for me it's always better to have one entry point to condition creation that having many of them without the clear reason. Remembering the rule: "If I want to create a condition then I should look at the static factory method in its class" is much simpler than "If I want these conditions then I need condition class, if I want that condition, then I need a separate UICondition" class.

Third, I'm not sure that many users except us will create skippable property conditions programmatically.

Let's try to ask a third opinion about this.

dtaimanov: Final Decision: remove `UiConditions` and use `.skipNullOrEmpty()` in each place

gorbunkov (approved):

Issue #2798, opened 2024-01-29
Dynamic attributes panel do not disallowing save entity with empty require attribute inside it


KartnDev:
### Environment

Jmix version: 2.1

### Bug Description
Dynamic attributes panel do not disallowing save entity with empty require attribute inside it
But message should notice that we cannot save enityt without required attribute

### Steps To Reproduce
1. Create dynamic attribute category
2. Create dynamic attribute inside category
3. Set type `"string"`
4. Add require option
5. Add visibility only for list view, or none views has visibility
6. Implement your entity **"Categorized"** interface
7. Add DynamicAttributePanel component onto you detail view
8. Go to the application
9. Choose current category inside `DynamicAttributePanel` 
10. Do not fill required attribute inside panel
11. Save

### Current Behavior

Save allowed if required attribute is only inside `dynAttr Panel`

### Expected Behavior

Save has to be banned. Also notificaton should notice that we cannot save entity without required attribute

rusiaikinat:
Tested on:
Kotlin, Java
Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6068-233
IntelliJ version: IntelliJ IDEA 2023.3.4 (Ultimate Edition)

PR #2882, opened 2024-02-12
fix Dynamic attributes panel do not disallowing save entity with empty require attribute inside it #2798

KartnDev:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/ViewControllerUtils.java
```diff
@@ -136,4 +136,8 @@ public static <T> Registration addInitEntityEvent(StandardDetailView<T> view, Co
     public static Registration addBeforeShowEvent(View<?> view, ComponentEventListener<View.BeforeShowEvent> listener) {
```
KartnDev, line 136: but idk why add-listener named "addBeforeShow**Event**", but i think `addValidationEventListener` is better then `addValidationEvent`

gorbunkov: It seems that that you should ask yourself about this naming:)

image

Name with the Listener suffix seems more correct, so analyze if it's possible to rename all methods that you added previously.

gorbunkov (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/ViewControllerUtils.java
KartnDev: Fixed in: https://github.com/jmix-framework/jmix/issues/2885

PR #2899, opened 2024-02-16
Add removal inside ViewControllerUtils methods that was renamed and mark it deprecated

KartnDev:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/ViewControllerUtils.java
```diff
@@ -141,4 +141,15 @@ public static Registration addBeforeShowEventListener(View<?> view, ComponentEve
     public static Registration addValidationEventListener(StandardDetailView<?> view, ComponentEventListener<StandardDetailView.ValidationEvent> listener) {
         return view.addValidationEventListener(listener);
     }
+
+    @Deprecated(since = "2.2", forRemoval = true)
```
gorbunkov, line 145: Write a `@deprecated` javadoc to explain which method should be used instead

gorbunkov (changes requested):

gorbunkov (approved):

Issue #2853, opened 2024-02-07
AbstractValueComponentDelegate makes unchecked class cast to EntityValueSource


Flaurite:
### Environment
Jmix version: 2.1.0+

### Bug Description
See Russian support forum: [topic](https://forum.jmix.ru/t/classcastexception-pri-ispolzovanii-svoej-realizaczii-valuesource/1614).

Method `AbstractValueComponentDelegate#setValueSource()` takes `ValueSource` type parameter .
But then without checking type casts to `EntityValueSource` type:
```
setupProperties((EntityValueSource<?, T>) valueSource);
```

### Expected Behavior
I think we can change parameter type in `setupProperties()` to `ValueSource` and inheritors will check type itself.

MaxKatyshev:
Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6128-233
IntelliJ version: IntelliJ IDEA 2023.3.4 (Community Edition)
verified

PR #2902, opened 2024-02-16
Change AbstractValueComponentDelegate.setupProperties param type to ValueSource

ilushaaa:
#2853


glebfox (changes requested):
Let's add test that checks that it is possible to set `ValueSource` that isn't implement `EntityValueSource`

glebfox (approved):

Issue #2700, opened 2023-12-28
Add cause exception if screen controller cannot be created 


glebfox:
### Feature Description

In the `io.jmix.ui.sys.ScreensImpl#createController` method if any problem occurs, an exception is thrown with a fixed message, making it difficult to identify the problem without debugging.

SergeiAksenov2:
Tested, but 1.5.999 and 1.6.999 snapshots were not compiled.

SergeiAksenov2:
Tested on:
Jmix version: 1.6.999-SNAPSHOT
Jmix Studio plugin version: 2.1.3-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Ultimate Edition)

PR #2909, opened 2024-02-19
Add cause of exception if screen controller cannot be created 2700

ilushaaa:
#2700


jmix-ui/ui/src/main/java/io/jmix/ui/sys/ScreensImpl.java
```diff
@@ -846,7 +846,8 @@ protected <T extends Screen> T createController(@SuppressWarnings("unused") Wind
             controller = invokeConstructor(screenClass);
         } catch (NoSuchMethodException | InstantiationException
                 | IllegalAccessException | InvocationTargetException e) {
-            throw new RuntimeException("Unable to create instance of screen class " + screenClass);
+            throw new RuntimeException("Unable to create instance of screen class " + screenClass +
```
glebfox, line 849: Cause must be passes as a second parameter

glebfox (changes requested):

glebfox (approved):

Issue #2465, opened 2023-11-14
UserIndicator doesn't reload user when refreshing


glebfox:
If a currently logged in user has been changed, then invoking `UserIndicator#refreshUser` has no effect because `CurrentUserSubstitution#getAuthenticatedUser` returns old instance.

gorbunkov:
Check in 2.x

SergeiAksenov2:
Tested on:
Jmix version: 1.6.999-SNAPSHOT
Jmix Studio plugin version: 2.1.3-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Ultimate Edition)

PR #2911, opened 2024-02-19
Reload user in UserIndicator when refresh method call 2456

ilushaaa:
#2465


jmix-ui/ui/src/main/java/io/jmix/ui/component/mainwindow/impl/UserIndicatorImpl.java
```diff
@@ -57,6 +58,9 @@ public class UserIndicatorImpl extends CompositeComponent<CssLayout> implements
 
     protected Formatter<? super UserDetails> userNameFormatter;
 
+    @Autowired
+    protected UserRepository userRepository;
+
```
glebfox, line 63: field injection is bad practice, take a loot at how other beans are injected

glebfox (changes requested):

Issue #2925, opened 2024-02-21
German date format for excel export incorrect


alferink:
### Environment

Jmix version: 2.2.0-SNAPSHOT

### Bug Description

I have the an application with a german locale. I am exporting an entity with date attributes to excel. The export of LocalDate and LocalDateTime is incorrect. The translation uses Java Format which does not work here.

I have a sample project. Login with german locale and export the customer
[excel-exportdemo.zip](https://github.com/jmix-framework/jmix/files/14358086/excel-exportdemo.zip)

### Current Behavior

Exporting of LocalDate "21/02/2024" is shown in Excel as 45343.
Exporting of LocalDateTime "21/02/2024 01:00" is shown in Excel as 45343,04167.

### Expected Behavior

Exporting of LocalDate "21/02/2024" should be shown in Excel as 21.02.2024.
Exporting of LocalDateTime "21/02/2024 01:00" is shown in Excel as 21.02.2024 01:00.

Related to [#2826](https://github.com/jmix-framework/jmix/issues/2826)

KremnevDmitry:
Port-issue for 1.x = #3150

SergeiAksenov2:
Tested on:
1) Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6306-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Community Edition)
2)  Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6305-241
IntelliJ version: IntelliJ IDEA 2024.1 (Community Edition)

PR #2926, opened 2024-02-21
#2925 date excel export

alferink:
Hi,

I have fixed the translation for dateFormat and dateTimeFormat for excel export.

See https://github.com/jmix-framework/jmix/issues/2925 for details


KremnevDmitry:
Hi Joerg!

Thanks for your contribution!

I fixed this problem as part of **release 2.2.2** (see #2925), and the localization issue arose due to incorrect parsing of localization values with exporting to Excel. 
To the best of my knowledge, the current localization values are correct.

Could you please check the localization of these values in the current **release 2.2.3**?
Can you try exporting an entity with attributes like `localDate` and `localDateTime`. Will the exported values be correct for German localization?

I added a test `localeDateTime` attribute to the `User` entity and exported the values with German localization, are these the correct values?
image
As far as I can see, the localization satisfies the one you specified in the linked task: #2925

> Exporting of LocalDate "21/02/2024" should be shown in Excel as 21.02.2024.
> Exporting of LocalDateTime "21/02/2024 01:00" is shown in Excel as 21.02.2024 01:00.

KremnevDmitry:
If the problem arises again, a new pull request will be opened. 
Current will be closed.

Issue #2180, opened 2023-08-29
HsqlSequenceSupport returns wrong srcipt for current value


dtaimanov:
See `io.jmix.data.impl.dbms.HsqlSequenceSupport#getCurrentValueSql`.

Script returns starting value instead of current.

Instead it should be like:
```sql
CALL CURRENT VALUE FOR {SEQUENCENAME};
```
 where `{SEQUENCENAME}` - name of the sequence without quotes or apostrophes.
 
or another option suggested in [forum topic](https://forum.jmix.io/t/bug-in-io-jmix-data-impl-dbms-hsqlsequencesupport/3486)

PR #2940, opened 2024-02-27
Change hsql script for sequence current value 2180

ilushaaa:
#2180

### DIFF
--- a/jmix-data/data/src/main/java/io/jmix/data/impl/dbms/HsqlSequenceSupport.java
+++ b/jmix-data/data/src/main/java/io/jmix/data/impl/dbms/HsqlSequenceSupport.java
@@ -51,7 +51,7 @@ public class HsqlSequenceSupport implements SequenceSupport {
 
     @Override
     public String getCurrentValueSql(String sequenceName) {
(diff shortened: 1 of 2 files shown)


jmix-data/data/src/main/java/io/jmix/data/impl/dbms/HsqlSequenceSupport.java
```diff
@@ -51,7 +51,7 @@ public String getNextValueSql(String sequenceName) {
 
     @Override
     public String getCurrentValueSql(String sequenceName) {
-        return "select START_WITH from INFORMATION_SCHEMA.SYSTEM_SEQUENCES where SEQUENCE_NAME = '"
+        return "select (cast(next_value as bigint) - 1) from INFORMATION_SCHEMA.SYSTEM_SEQUENCES where SEQUENCE_NAME = '"
```
gorbunkov, line 54: @ilushaaa Won't this query increase the current sequence value? I mean if I call the getCurrentValue() multiple times, then we will get a different value for each invocation, right?

Were there problems with the `CURRENT VALUE FOR {SEQUENCENAME}` approach?

gorbunkov (approved):

Issue #2833, opened 2024-02-02
The filter for the 'Entity type' field does not work


rusiaikinat:
### Environment

Jmix version: 2.2.999-SNAPSHOT

### Bug Description
![image](https://github.com/jmix-framework/jmix/assets/114649487/20fc22fa-7e49-4552-94de-0a30aabaf4c3)

### Steps To Reproduce

Enter the class name in the `Entity type` field

### Expected Behavior

The filter for the entered line does not work

PR #2942, opened 2024-02-28
Remove callback data provider 2833

ilushaaa:
#2833


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/category/CategoryDetailView.java
```diff
@@ -206,9 +205,9 @@ protected void initEntityTypeField() {
         }
         entityTypeField.setItemLabelGenerator(options::get);
         //noinspection unchecked
-        entityTypeField.setItems(DataProviderUtils.createCallbackDataProvider(options.keySet().stream()
+        entityTypeField.setItems(options.keySet().stream()
```
KartnDev, line 208: I can explain why did i use dataprovider. When we used previous version of vaadin, comboboxes didn't work with list items normally or ListDataProvider. 

If right now this working well, we can do it. 

I don't think, that we need in any of DataProvider types(limit/offset paginations inside its api). So, if this tested and working fine, merge

gorbunkov (approved):

PR #2957, opened 2024-02-29
fix NullPointerException: Master-detail view + DynamicAttribute

KartnDev:
Duplicated from issue comment(i created an issues):

There is not errors for master detail screen.
Bug was produced by service that finds elements with "null" id. 
So, we need to fix that method, ignore searching elements without IDs(formLayout and dataGrid).

Issue causes following issues:
* No root form layout id inside root form: https://github.com/jmix-framework/jmix/issues/2958
* No validation inside MasterDetailsView's form layout https://github.com/jmix-framework/jmix/issues/2959

We would not support detail's view behaviour for form inside it, but we can fix it for scaffolded views

### DIFF
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/impl/DynAttrFacetInfo.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/impl/DynAttrFacetInfo.java
@@ -25,6 +25,7 @@ import io.jmix.core.common.xmlparsing.Dom4jTools;
 import io.jmix.core.metamodel.model.MetaClass;
 import io.jmix.dynattrflowui.facet.DynAttrFacet;
 import io.jmix.flowui.component.formlayout.JmixFormLayout;
+import io.jmix.flowui.component.grid.DataGrid;
 import io.jmix.flowui.view.ViewInfo;
 import io.jmix.flowui.view.ViewRegistry;
 import org.apache.commons.lang3.StringUtils;
@@ -223,8 +224,11 @@ public class DynAttrFacetInfo {
     private void findTargetElementNames(List<String> targetElementList, Element searchElement) {
         for (var child : searchElement.elements()) {
             if (child.getQName().getName().equals(JmixFormLayout.QUALIFIED_XML_NAME) ||
-                    child.getQName().getName().equals("dataGrid")) {
-                targetElementList.add(child.attributeValue("id"));
+                    child.getQName().getName().equals(DataGrid.QUALIFIED_XML_NAME)) {
+                String idAttribute = child.attributeValue("id");
+                if(StringUtils.isNotBlank(idAttribute)) {
+                    targetElementList.add(idAttribute);
+                }
             }
             if (!child.elements().isEmpty()) {
                 findTargetElementNames(targetElementList, child);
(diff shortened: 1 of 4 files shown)


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/impl/DynAttrFacetInfo.java
```diff
@@ -223,8 +224,11 @@ public Collection<String> getDynAttrViewTargetComponentIds(MetaClass metaClass,
     private void findTargetElementNames(List<String> targetElementList, Element searchElement) {
         for (var child : searchElement.elements()) {
             if (child.getQName().getName().equals(JmixFormLayout.QUALIFIED_XML_NAME) ||
-                    child.getQName().getName().equals("dataGrid")) {
-                targetElementList.add(child.attributeValue("id"));
+                    child.getQName().getName().equals(DataGrid.QUALIFIED_XML_NAME)) {
+                String idAttribute = child.attributeValue("id");
+                if(StringUtils.isNotBlank(idAttribute)) {
```
gorbunkov, line 229: Add a space between **if** and **(**

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/DataGrid.java
```diff
@@ -59,6 +59,7 @@
 
 public class DataGrid<E> extends JmixGrid<E> implements ListDataComponent<E>, MultiSelectLookupComponent<E>,
         EnhancedDataGrid<E>, SupportsEnterPress<DataGrid<E>>, ApplicationContextAware, InitializingBean {
+    public static final String QUALIFIED_XML_NAME = "dataGrid";
```
gorbunkov, line 62: What is the purpose of the constant? Is it used somewhere else? I'm not sure that we should introduce a constant here.

1. It seems that constant name is not correct. Is this a **qualified** name? In my understanding a qualified name is the name that consist of a schema name and a local name, e.g. `xsl:template`. I don't see any schema here.
2. If we start adding a constant with element name to classes then it looks like a pattern. It is expected that each element must have its name in a constant. But we don't need this yet and also it's not always possible (because some components come from Vaadin and we don't have classes for them inside Jmix.

gorbunkov (changes requested):

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/impl/DynAttrFacetInfo.java
```diff
@@ -222,9 +222,12 @@ public Collection<String> getDynAttrViewTargetComponentIds(MetaClass metaClass,
 
     private void findTargetElementNames(List<String> targetElementList, Element searchElement) {
         for (var child : searchElement.elements()) {
-            if (child.getQName().getName().equals(JmixFormLayout.QUALIFIED_XML_NAME) ||
-                    child.getQName().getName().equals("dataGrid")) {
-                targetElementList.add(child.attributeValue("id"));
+            if (child.getQName().getName().equals(FORM_LAYOUT_QUALIFIED_XML_NAME) ||
```
gorbunkov, line 225: Again, what about my previous comment?

> It seems that constant name is not correct. Is this a qualified name? In my understanding a qualified name is the name that consist of a schema name and a local name, e.g. xsl:template. I don't see any schema here.

Even in the code you first get the **qualified** name by `child.getQName()` and then you get the **local** name from the QName object: `child.getQName().getName()`. So let's remove the word QUALIFIED from constant names. For example, DATAGRID_QUALIFIED_XML_NAME may be just DATAGRID_XML_TAG

gorbunkov (changes requested):

Issue #2960, opened 2024-03-01
gridColumnVisibility component does not show labels for filterable columns


knstvk:
### Environment

Jmix version: 2.2.0

### Bug Description

gridColumnVisibility component does not show label for filterable column.

### Steps To Reproduce

- Create a project
- Add gridColumnVisibility component and `filterable` for some columns in `user-list-view.xml`:

```
<hbox id="buttonsPanel" classNames="buttons-panel">
    <gridColumnVisibility icon="COG" themeNames="icon" dataGrid="usersDataGrid"/>
</hbox>
<dataGrid id="usersDataGrid"
          width="100%"
          minHeight="20em"
          dataContainer="usersDc">
    <columns resizable="true">
        <column property="username" filterable="true" resizable="false" autoWidth="true"/>
        <column property="firstName" filterable="true" resizable="false" autoWidth="true"/>
        <column property="lastName" filterable="true" resizable="false" autoWidth="true"/>
        <column property="email"/>
        <column property="timeZoneId"/>
        <column property="active"/>
```

### Current Behavior

<img width="965" alt="image" src="https://github.com/jmix-framework/jmix/assets/5513878/bd801ea7-f729-46ae-adff-636b0686e4df">

### Expected Behavior

Property names are used as for other non-filterable columns.

SergeiAksenov2:
Tested on:
1) Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6254-233
IntelliJ version: IntelliJ IDEA 2023.3.6 (Ultimate Edition)
2) Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6256-241
IntelliJ version: IntelliJ IDEA 2024.1 (Ultimate Edition)

![image](https://github.com/jmix-framework/jmix/assets/158277660/ff3e5c1e-6373-4380-8ccc-9d78718b5f8a)

PR #3036, opened 2024-03-22
Get column header text from filter header internal component 2960

ilushaaa:
#2960


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/gridcolumnvisibility/JmixGridColumnVisibility.java
```diff
@@ -201,6 +202,11 @@ protected String getColumnHeaderText(DataGridColumn<?> column) {
             if (headerComponent instanceof HasText hasText) {
                 headerText = hasText.getText();
             }
+            if (Strings.isNullOrEmpty(headerText)
```
glebfox, line 205: What if we still can't get the `title` from the `header`? For example, a user set a custom component to a header? 

In the issue, the is a comment that 
> property names are used for other columns

Take a look at how the default `title` is generated for a column and let's apply the same for the `menuItem` in case we cannot obtain `title` from column.

glebfox (changes requested):

glebfox (approved):

Issue #3020, opened 2024-03-19
NPE is raised when we apply the settings for the dataGrid with editorActionsColumn


ilushaaa:
### Environment

Jmix version: 2.x

### Bug Description

NPE is raised when we apply the settings for the dataGrid with editorActionsColumn

### Steps To Reproduce

1. Create dataGrid which has an editorActionsColumn.
2. Make columns resizable.
3. Add settings facet.
4. Open screen, change columns size, close screen, open again.

### Current Behavior

editorActionsColumn doesn't init key property by default and it equals to null. We save column with `key == null`. 
And when we apply settings and try to find column with `key == null` the NPE is raised.

`java.lang.NullPointerException: null
	at java.base/java.util.Objects.requireNonNull(Objects.java:209) ~[na:na]
	at io.jmix.flowui.facet.settings.component.binder.AbstractGridSettingsBinder.applySettings(AbstractGridSettingsBinder.java:59) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowui.facet.settings.component.binder.AbstractGridSettingsBinder.applySettings(AbstractGridSettingsBinder.java:37) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowuidata.settings.ViewSettingsComponentManagerImpl.applySettings(ViewSettingsComponentManagerImpl.java:68) ~[jmix-flowui-data-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowui.facet.impl.SettingsFacetImpl.applyViewSettings(SettingsFacetImpl.java:305) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowui.facet.impl.SettingsFacetImpl.applySettings(SettingsFacetImpl.java:105) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowui.facet.impl.SettingsFacetImpl.onViewReady(SettingsFacetImpl.java:283) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
	at io.jmix.flowui.facet.impl.SettingsFacetImpl$ViewEventListener.lambda$subscribe$9b1b5227$1(SettingsFacetImpl.java:401) ~[jmix-flowui-2.3.999-SNAPSHOT.jar:na]
[...]

PR #3052, opened 2024-03-27
Set default value for key if initial value is empty

ilushaaa:
#3020

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractGridLoader.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractGridLoader.java
@@ -17,6 +17,7 @@
 package io.jmix.flowui.xml.layout.loader.component;
 
 import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.grid.ColumnRendering;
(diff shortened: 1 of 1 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractGridLoader.java
```diff
@@ -256,6 +258,11 @@ protected void loadEditorActionsColumn(T resultComponent, Element columnElement)
         editColumn.setEditorComponent(actions);
 
         loadString(columnElement, "key", editColumn::setKey);
+        if (Strings.isNullOrEmpty(editColumn.getKey())) {
```
glebfox, line 261: This doesn't fix existing settings, so anyone upgrading to the new version will still run into this issue

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractGridLoader.java
```diff
@@ -255,7 +244,10 @@ protected void loadEditorActionsColumn(T resultComponent, Element columnElement)
 
         editColumn.setEditorComponent(actions);
 
-        loadString(columnElement, "key", editColumn::setKey);
+        //If the key is null then NPE will rise when the settings are applied
+        loadString(columnElement, "key").ifPresentOrElse(
+                editColumn::setKey,
+                () -> editColumn.setKey(EDITOR_ACTION_COLUMN_DEFAULT_KEY));
```
glebfox, line 250: Let's use element name as Id. Having element name in plural form and id in singular looks weird

glebfox (changes requested):

glebfox (approved):

glebfox:
Add solution description to the original issue + `breaking change` label

Issue #1442, opened 2023-02-20
Implement IN and DATE_INTERVAL operations in PropertyFilter


glebfox:

Flaurite:
Consider to add permission to `DateIntervalDialog`, since it is not available even with Filter role. See: https://github.com/jmix-framework/jmix/issues/3121

KremnevDmitry:
> Consider to add permission to `DateIntervalDialog`, since it is not available even with Filter role. See: #3121

Fixed with https://github.com/jmix-framework/jmix/commit/a673fab24ec11f9b2a8ee3bf2a288967f54259c6

SergeiAksenov2:
Tested on:
1) Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6329-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Community Edition)
2) Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6332-241
IntelliJ version: IntelliJ IDEA 2024.1 (Community Edition)

Operations are available in propertyFilter and in Filter.

KremnevDmitry:
Doc issue linked: https://github.com/jmix-framework/jmix-docs/issues/785

PR #3060, opened 2024-03-28
Implement IN and DATE_INTERVAL operations in PropertyFilter jmix-framework/jmix#1442

KremnevDmitry:
See: #1442


jmix-flowui/flowui/src/main/java/io/jmix/flowui/app/propertyfilter/dateinterval/DateIntervalDialog.java
```diff
@@ -0,0 +1,272 @@
...
+import org.springframework.lang.Nullable;
+
+import java.sql.Time;
+import java.time.LocalTime;
+import java.time.OffsetTime;
+import java.util.*;
+
+@ViewController("flowui_DateIntervalDialog")
+@ViewDescriptor("date-interval-dialog.xml")
+@DialogMode(height = "AUTO", width = "50em", resizable = true)
```
glebfox, line 52: 1. isn't height AUTO by default?
2. As I remember, we decided to make it resizable=false

glebfox: 3. 
Screenshot 2024-04-01 at 11 41 21

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/app/propertyfilter/dateinterval/DateIntervalDialog.java
KremnevDmitry: 1. Yep, u are right(
2. Yes, cause when the dialog is displayed on a mobile device, it takes on the maximum available width by default. Fixed.
3. I do vanga, u haven't installed `jmix-data` module, which contains the `RelativeDateTimeMoment` enumeration and its localization

glebfox: but I can select *relative date and time* without this module. Why do I need to install additional module just to get a localized messages?

KremnevDmitry: No, you cannot select the _relative date and time_ without this module. The entity that will be processed on the backend is determined in this module
`RelativeDateTimeMoment` location - `io.jmix.data.impl.queryconstant`
Therefore the localization is in the `jmix-data` module

glebfox (approved):

Issue #3032, opened 2024-03-22
Add access denied page


NikitaShchienko:
## Environment

Jmix version: 2.1.3
Jmix Studio plugin version: 2.2.0-233
IntelliJ version: IntelliJ IDEA 2023.3.2 (Ultimate Edition)

## Steps To Reproduce
1 - Create a project with the OIDC Add-on
2 - Login to SSO as a user who does not have Jmix roles
3 - Go to Jmix application

## Current Behavior

A page appears with the message "Could not navigate to ''"

**Example:**
<img width="1792" alt="image" src="https://github.com/jmix-framework/jmix/assets/47520144/47901081-0b3d-46ef-8580-7ad3da52d350">

gorbunkov:
See item 1 in #2928

PR #3093, opened 2024-04-03
#3032: Redirect to AccessDeniedException instead of NotFoundException

swapdewalkar:
Address [3032](https://github.com/jmix-framework/jmix/issues/3032)

Redirect to com.vaadin.flow.router.AccessDeniedException in case of view is not permitted by security permissions.

ran unit tests


glebfox:
Hi, thank you for your contribution, but this is no longer relevant since the security has been reworked, see https://github.com/jmix-framework/jmix/pull/3182.

Issue #2918, opened 2024-02-20
UserIndicator doesn't reload user when refreshing v2.x


ilushaaa:
### Environment

Jmix version: 2.x

### Bug Description

If a currently logged in user has been changed, then invoking `UserIndicator#refreshUser` has no effect because `CurrentUserSubstitution#getAuthenticatedUser` returns old instance.

### Steps To Reproduce

1. Create two users.
2. Set User1 as substitution for User2
3. Login as User1.
4. Change User1 first name.
5. Change user in the user indicator panel.

### Current Behavior

User doesn't change and shows old user first name.

### Expected Behavior

User indicator should show the actual user info.

Related to #2465

ilushaaa:
The [PR](https://github.com/jmix-framework/jmix/pull/3098) should fix the problem.

SergeiAksenov2:
Tested on Java and Kotlin on:
1) Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6289-241
IntelliJ version: IntelliJ IDEA 2024.1 (Community Edition)
2) Jmix version: 2.2.999-SNAPSHOT
Jmix Studio plugin version: 2.2.SNAPSHOT6292-233
IntelliJ version: IntelliJ IDEA 2023.3.3 (Community Edition)

Users' names are change - Ok.

PR #3098, opened 2024-04-04
Refresh authenticated user in JmixUserIndicator 2918

ilushaaa:
#2918

UsernameNotFound exception handling to prevent tests falling if they use UserIndicator component

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/main/JmixUserIndicator.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/main/JmixUserIndicator.java
@@ -46,7 +46,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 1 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/main/JmixUserIndicator.java
```diff
@@ -110,6 +120,11 @@ public void refreshUser() {
         getContent().removeAll();
 
         UserDetails user = currentUserSubstitution.getAuthenticatedUser();
```
glebfox, line 122: Technically, we want to set a user from `currentUserSubstitution` only if we can't obtain a fresh one from `userRepository`, so this code can be moved to the `catch`

glebfox (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/main/JmixUserIndicator.java
ilushaaa: We load user from `userRepository` by `userName` which we get from `currentUserSubstitution.getAuthenticatedUser()`

Issue #3054, opened 2024-03-27
Ability to display HTML content in background task dialog


knstvk:
`Dialogs.BackgroundTaskDialogBuilder` should implement `HasContent` interface as other dialog builders.

PR #3112, opened 2024-04-09
Ability to display HTML content in background task dialog 3054

ilushaaa:
#3054

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
@@ -881,16 +881,9 @@ public class DialogsImpl implements Dialogs {
 
         @Override
         public BackgroundTaskDialogBuilder<T, V> withText(String text) {
-            if (!(content instanceof Span)) {
-                if (content != null) {
-                    layout.remove(content);
-                }
-                content = uiComponents.create(Span.class);
-                layout.addComponentAtIndex(0, content);
-            }
-            ((Span) content).setText(text);
-
-            return this;
+            Span span = uiComponents.create(Span.class);
+            span.setText(text);
+            return withContent(span);
         }
 
         @Nullable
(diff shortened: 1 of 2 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
```diff
@@ -769,6 +769,8 @@ public DialogWindow<InputDialog> build() {
     public class BackgroundTaskDialogBuilderImpl<T extends Number, V> implements BackgroundTaskDialogBuilder<T, V> {
 
         protected Dialog dialog;
+        protected VerticalLayout layout;
+        protected Component content;
```
glebfox, line 773: It makes no sense to store both `content` and `messageSpan` as a separate fields since we can't display them both as the same time. Take a look at the other dialogs and how `text`/`content` are handled.

glebfox (changes requested):

glebfox:
Since we make changes into dialog, let's fix `DialogsImpl.MessageDialogBuilderImpl#withContent` and `DialogsImpl.OptionDialogBuilderImpl#withContent` methods - they are not storing new `content` in a field.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
```diff
@@ -879,14 +881,22 @@ public String getHeader() {
 
         @Override
         public BackgroundTaskDialogBuilder<T, V> withText(String text) {
-            this.messageText = text;
+            if (!(content instanceof Span)) {
```
glebfox, line 884: if content is a `Span` it doesn't mean that it's a Span that is created for `text`. Let's simplify this code to 

```
Span span = uiComponents.create(Span.class);
span.setText(text);
return withContent(span);
```

So, we don't duplicate code of `withContent` method.

glebfox (changes requested):

glebfox (approved):

Issue #889, opened 2022-07-26
[Flow UI] Disable Edit button in multi select Grid?


pbaris:
Having a Grid with selection mode MULTI, when i select more than one rows and click Edit button, editor opens with the first selected row's entity.

I think it would be better if Edit button was disabled when multiple rows selected.

![jmix_flowui_multi_edit](https://user-images.githubusercontent.com/242612/181098207-c61bc17a-43c5-45df-9e6a-f796a32f52c3.gif)

SergeiAksenov2:
Tested on Java:
Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6351-241
IntelliJ version: IntelliJ IDEA 2024.1 (Community Edition)

Tested on Kotlin:
Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6350-233
IntelliJ version: IntelliJ IDEA 2023.3.6 (Community Edition)

The 'Edit' button only is avaliable in a Grid table when one entity is selected.

PR #3115, opened 2024-04-09
Disable Edit button in multi select Grid 889

ilushaaa:
#889


jmix-flowui/flowui/src/main/java/io/jmix/flowui/action/list/EditAction.java
```diff
@@ -246,6 +246,7 @@ public void setText(@Nullable String text) {
     protected boolean isPermitted() {
         if (target == null
                 || target.getSingleSelectedItem() == null
+                || target.getSelectedItems().size() != 1
```
glebfox, line 249: is `EditAction` the only action that requires such adjustment?

glebfox (changes requested):

glebfox (approved):

Issue #2958, opened 2024-02-29
Add id for master-detail form inside master-detail-view studio template


KartnDev:
I found issue when fixing https://github.com/jmix-framework/jmix/pull/2957: Inside master-detail-view that scaffold by studio has high-level `formLayout` without id. 
```xml
<layout padding="false">
        <formLayout classNames="master-detail-layout" height="100%"> 
        ...
</layout>
```

That will be great if id would be scaffolded too.

PR #3119, opened 2024-04-09
Add id for master-detail form inside master-detail-view studio template 2958

ilushaaa:
#2958

### DIFF
--- a/jmix-templates/content/flowui/master-detail/descriptor.xml
+++ b/jmix-templates/content/flowui/master-detail/descriptor.xml
@@ -90,7 +90,7 @@ def optionsDatasourceXml = api.processSnippet('optionsDsSource.xml',
         <action id="discardAction" type="lookup_discard"/>
     </actions>
     <layout padding="false">
-        <formLayout id="masterDetailForm" classNames="master-detail-layout" height="100%">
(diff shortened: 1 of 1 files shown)


jmix-templates/content/flowui/master-detail/descriptor.xml
```diff
@@ -90,7 +90,7 @@ def optionsDatasourceXml = api.processSnippet('optionsDsSource.xml',
         <action id="discardAction" type="lookup_discard"/>
     </actions>
     <layout padding="false">
-        <formLayout classNames="master-detail-layout" height="100%">
+        <formLayout id="masterDetailForm" classNames="master-detail-layout" height="100%">
```
glebfox, line 93: `masterDetail` is the name of view template. the `formLayout` is just a form in the detail part of the view. Give it the same id as in the regular detail view

glebfox (changes requested):

jmix-templates/content/flowui/master-detail/descriptor.xml
ilushaaa: The descriptor already contains formLayout with id is equals "form". I suggest one of the following options:
1. masterForm and form
2. form and detailsForm
3. masterForm and detailsForm

PR #3130, opened 2024-04-12
jmix.core.exclude-beans defined in project overrides values from Jmix modules (3129)

### LINKED ISSUE #3129: jmix.core.exclude-beans defined in project overrides values from Jmix modules
Currently, `jmix.core.exclude-beans` is defined in the **flowui** module. If `jmix.core.exclude-beans` is defined in a project, it overrides value set in **flowui** module which leads to errors.

**Possible solution** is to collect all `jmix.core.exclude-beans` values from all modules, same as for `jmix.ui.menu-config`.

glebfox:
Fixes #3129


knstvk (approved):
👍

Issue #3092, opened 2024-04-03
Generic Filter reloads data loaders redundandly


dgodlevskiy:
### Environment

Jmix version: 2.2.1

### Bug Description

When using generic filter with saved configuration it forces linked data loaders to reload multiple times:
- When configuration is applied
- Each time a default value is set to the filter condition.

If using xml filter configuration it happens before the pagination is applied.

The above significantly decreases performance of views on loading.

### Steps To Reproduce

On a standard list view:

1. Add a condition to generic filter with some value
2. Save the configuration with values and make it default
3. Open the screen and check how many times collection data load reloads (e.g. by debugger)

### Current Behavior

Data loader loads multiple times.

### Expected Behavior

Data loader loads just once with filter conditions and pagination applied

PR #3165, opened 2024-04-17
Generic Filter reloads data loaders redundandly 3092

ilushaaa:
#3092

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SingleFilterComponentBase.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SingleFilterComponentBase.java
@@ -60,7 +60,6 @@ public abstract class SingleFilterComponentBase<V> extends CustomField<V>
 
     protected DataLoader dataLoader;
     protected boolean autoApply;
-    protected AutoApplyResolver autoApplyResolver;
     protected Condition queryCondition;
 
     @Internal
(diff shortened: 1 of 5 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/GenericFilterLoader.java
```diff
@@ -239,4 +251,52 @@ protected ActionLoaderSupport getActionLoaderSupport() {
 
         return actionLoaderSupport;
     }
+
+    protected void applyFilterIfNeeded() {
+        getComponentContext().addInitTask((context, view) -> {
```
glebfox, line 256: Filter shouldn't load data if a user don't want to load data, just add conditions

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SingleFilterComponentBase.java
```diff
@@ -203,12 +204,16 @@ public Condition getQueryCondition() {
     public void apply() {
         if (dataLoader != null) {
             setupLoaderFirstResult();
-            if (autoApply) {
+            if (shouldAutoApply() && autoApply) {
```
glebfox, line 207: is `SingleFilterComponent` is the problem? For example, if I have several GenericFilter configurations saved, when I open a view, the number of data refreshes is comparable to the number of configurations, not the number of conditions.

glebfox (changes requested):

glebfox (approved):

Issue #3174, opened 2024-04-22
Exception when settings facet is aplied to grid with columns without key specified


syncro:
### Environment

Jmix version: 2.x

### Bug Description

When grid column key not specified and settings facet presents on view it produces runtime Exception

### Steps To Reproduce

1, Create project
2, Add view with grid and settings facet
3. Add column with custom renderer programmatically from view context and no key specified
4. Run the project, navigate to view and see exception popup

### Current Behavior

The exception is thrown when column key is not specified

### Expected Behavior

The exception is not thorwn when column key is not specified

[testgridcols.zip](https://github.com/jmix-framework/jmix/files/15057876/testgridcols.zip)

PR #3175, opened 2024-04-22
Exception when settings facet is aplied to grid with columns without …

syncro:
[…key specified 3174](https://github.com/jmix-framework/jmix/issues/3174)

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
@@ -84,7 +84,7 @@ public class DataGridSettings implements Settings {
         protected int flexGrow;
 
         public String getKey() {
-            return key;
+            return key != null ? key : this.toString();
         }
 
         public void setKey(String key) {
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
@@ -60,11 +60,12 @@ public class DataGridSettings implements Settings {
         protected String key;
         protected String sortDirection;
 
+        @Nullable
         public String getKey() {
             return key;
         }
 
-        public void setKey(String key) {
+        public void setKey(@Nullable String key) {
             this.key = key;
         }
 
@@ -88,7 +89,7 @@ public class DataGridSettings implements Settings {
             return key;
         }
 
-        public void setKey(String key) {
+        public void setKey(@Nullable String key) {
             this.key = key;
         }
 
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
@@ -83,8 +83,9 @@ public class DataGridSettings implements Settings {
         protected Boolean visible;
         protected int flexGrow;
 
+        @Nullable
         public String getKey() {
-            return key != null ? key : this.toString();
+            return key;
         }
 
         public void setKey(String key) {
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
@@ -26,6 +26,8 @@ import io.jmix.flowui.model.CollectionContainer;
 import io.jmix.flowui.model.CollectionLoader;
 import io.jmix.flowui.model.HasLoader;
 import org.apache.commons.collections4.CollectionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.lang.Nullable;
 
 import java.util.ArrayList;
@@ -37,6 +39,8 @@ import static org.apache.commons.collections4.CollectionUtils.isEmpty;
 
 public abstract class AbstractGridSettingsBinder<V extends Grid<?>, S extends DataGridSettings>
         implements DataLoadingSettingsBinder<V, S> {
+
+    private static final Logger log = LoggerFactory.getLogger(AbstractGridSettingsBinder.class);
 
     @SuppressWarnings({"unchecked", "rawtypes"})
     @Override
@@ -107,6 +111,16 @@ public abstract class AbstractGridSettingsBinder<V extends Grid<?>, S extends Da
         }
     }
 
+    protected List<? extends Grid.Column<?>> getApplicableColumns(V component) {
+        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component);
+        List<? extends Grid.Column<?>> applicableColumns = componentColumns.stream()
+                .filter(c -> Objects.nonNull(c.getKey())).toList();
+        if (componentColumns.size() != applicableColumns.size()) {
+            log.debug("Grid has column without key specified, settings for it would not be stored");
+        }
+        return applicableColumns;
+    }
+
     @Override
     public boolean saveSettings(V component, S settings) {
         boolean changed = false;
@@ -116,10 +130,9 @@ public abstract class AbstractGridSettingsBinder<V extends Grid<?>, S extends Da
             setSortOrderToSettings(sortOrder, settings);
             changed = true;
         }
-        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component).stream()
-                .filter(c -> !isNull(c.getKey())).toList();
-        if (isColumnSettingsChanged(componentColumns, settings.getColumns())) {
-            setColumnsToSettings(componentColumns, settings);
+        List<? extends Grid.Column<?>> applicableColumns = getApplicableColumns(component);
+        if (isColumnSettingsChanged(applicableColumns, settings.getColumns())) {
+            setColumnsToSettings(applicableColumns, settings);
             changed = true;
         }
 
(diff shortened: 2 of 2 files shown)


syncro:
@glebfox I made some more testing of this and using hashCode dependant value seems unreliable to me. The developer can enable sorting for column without key specified and get unexpected behaviour

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
```diff
@@ -84,7 +84,7 @@ public static class Column {
         protected int flexGrow;
 
         public String getKey() {
-            return key;
+            return key != null ? key : this.toString();
```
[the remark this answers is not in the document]
syncro: changed the behaviour of grid settings binder to be null-key tolerant

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -113,7 +116,8 @@ public boolean saveSettings(V component, S settings) {
             setSortOrderToSettings(sortOrder, settings);
             changed = true;
         }
-        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component);
+        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component).stream()
+                .filter(c -> !isNull(c.getKey())).toList();
```
glebfox, line 120: 1. `java.util.Objects#nonNull`
2. This method exists to be used as a `java.util.function.Predicate`, `filter(Objects::nonNull)`. In your case, you have to call `c.getKey()` so using `Objects` makes no sense.
3. As discussed, we need to log **warning** that column is ignored.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -128,7 +132,8 @@ public S getSettings(V component) {
         settings.setId(component.getId().orElse(null));
 
         setSortOrderToSettings(component.getSortOrder(), settings);
-        setColumnsToSettings(getOrderedColumns(component), settings);
+        setColumnsToSettings(getOrderedColumns(component).stream()
```
glebfox, line 135: Duplicate code can be moved to a method

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -166,7 +171,7 @@ protected boolean isColumnSortOrderChanged(@Nullable List<? extends GridSortOrde
 
             DataGridSettings.SortOrder sSortOrder = settingsSortOrder.get(i);
 
-            if (!key.equals(sSortOrder.getKey())) {
+            if (key != null && !key.equals(sSortOrder.getKey())) {
```
glebfox, line 174: if `sSortOrder.getKey()` is nut null, then this code can be simplified to `!sSortOrder.getKey().equals(key)` or `java.util.Objects#equals`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -86,6 +88,7 @@ public void applyDataLoadingSettings(V component, S settings) {
         }
 
         List sortOrder = settings.getSortOrder().stream()
+                .filter(o -> !Objects.isNull(o.getKey()))
```
glebfox, line 91: if `SortOrder#getKey` can be `null` then it makes sense to add `@Nullable`

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/DataGridSettings.java
```diff
@@ -83,6 +83,7 @@ public static class Column {
         protected Boolean visible;
         protected int flexGrow;
 
+        @Nullable
```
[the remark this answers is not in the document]
syncro: added

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: changed and added debug logging

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: extracted to method

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: sort order key can be null, added @Nullable

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: solved by nullable annotation

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -104,6 +111,16 @@ public void applyDataLoadingSettings(V component, S settings) {
         }
     }
 
+    protected List<? extends Grid.Column<?>> getApplicableColumns(V component) {
+        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component);
+        List<? extends Grid.Column<?>> applicableColumns = componentColumns.stream()
+                .filter(c -> Objects.nonNull(c.getKey())).toList();
+        if (componentColumns.size() != applicableColumns.size()) {
+            log.debug("Grid has column without key specified, settings for it would not be stored");
```
glebfox, line 119: ... one or more columns ..., ... for them won't be stored

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
```diff
@@ -104,6 +111,16 @@ public void applyDataLoadingSettings(V component, S settings) {
         }
     }
 
+    protected List<? extends Grid.Column<?>> getApplicableColumns(V component) {
+        List<? extends Grid.Column<?>> componentColumns = getOrderedColumns(component);
+        List<? extends Grid.Column<?>> applicableColumns = componentColumns.stream()
+                .filter(c -> Objects.nonNull(c.getKey())).toList();
```
glebfox, line 117: This method exists to be used as a `java.util.function.Predicate`, `filter(Objects::nonNull)`. In your case, you have to call `c.getKey()` so using `Objects` makes no sense.

P.S. this is applicable to all `!isNull`

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: changed to null key lookup

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: changed Objects to operator comparison

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
glebfox: I didn't mean code changes. You iterate over column anyway and even with this implementation you still can have several columns with no `id`, so why not use a more general message? Also, log message must be **warning** as I mentioned before, so devs notice it.

One more thing, we prefer not hard code component name, since it can be changed + there are `Grid` and `TreeGrid` components. `component.getClass().getSimpleName()` would be better.

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/settings/component/binder/AbstractGridSettingsBinder.java
syncro: warn will polute the logs in case the developer is happy with columns without key

syncro: added component.getClass().getSimpleName()

glebfox: > warn will polute the logs in case the developer is happy with columns without key

that is the point
1. setting won't be applied if actual columns number isn't equal to number of columns in saved settings. So, with your current implementation, adding settings facet makes no sense, hence we need to **warn** devs, so it's easier for them to find the problem
2. if a dev adds settings facet, they assume that settings for all columns will be saved.

> added component.getClass().getSimpleName()
You still log message about a single column. I've already mentioned that there may be several and we need to use more general message.

syncro: changed to warn, and log message, also they are currently not compared by size but there lookup for null-keyed column is performed

glebfox (approved):

Issue #3199, opened 2024-04-26
Wrong Russian translation for EntityType for process listener in 


KartnDev:
### Environment

Jmix version: 2.2

### Bug Description
2х `Задача`
![image](https://github.com/jmix-framework/jmix/assets/43281522/c093bb28-96db-47f7-89a0-e74b05156530)
```

io.jmix.bpmflowui.view.modeler.properties.eventlistener/job=Задача
io.jmix.bpmflowui.view.modeler.properties.eventlistener/attachment=Вложения
io.jmix.bpmflowui.view.modeler.properties.eventlistener/comment=Комментарий
io.jmix.bpmflowui.view.modeler.properties.eventlistener/identity-link=Идентификационная ссылка
io.jmix.bpmflowui.view.modeler.properties.eventlistener/task=Задача
```

### Should be 

`io.jmix.bpmflowui.view.modeler.properties.eventlistener/job=Работа`

SergeiAksenov2:
The "Задача" item is displayed once:

![image](https://github.com/jmix-framework/jmix/assets/158277660/eb76a010-6323-494b-96a9-bc193d830099)

PR #3200, opened 2024-04-26
fix Wrong Russian translation for EntityType for process listener in bpm #3199

KartnDev:


KartnDev:
Close due to unactual file and its diff

Issue #327, opened 2022-03-01
About the total size of inputstream


GuoRuiLv:
Hi Team,

I have a question about following code block:

```
try (BufferedInputStream bos = new BufferedInputStream(inputStream, chunkSizeBytes)){

            OSS client = clientReference.get();
            int totalSize = bos.available();
            List<PartETag> partETags = new ArrayList<>();
```

There using  `available() ` method to get total size of InputStream. In my experience ,the method will return 0 in some cases. For example:
```
RecordPictureResponse recordPicture(@RequestHeader String sn, @RequestHeader String id, @RequestHeader String name,
                                        @RequestHeader String regTime,
                                        @RequestPart("file") MultipartFile file) {
        RecordPictureResponse response=new RecordPictureResponse();
        try {
            String imageUrl=personCmdSupport.processRecordPicture(sn,id,name,regTime,file.getInputStream(),file.getSize());
```
The MultipartFile.getInputStream().available() will return 0.  In this case, it  cannot get correct size, then we cannot save file processed by MultipartFile. 
Is this an expected behavior?

gorbunkov:
Hi @GuoRuiLv,

Could you please explain, which Jmix version and which class and line are you talking about? Maybe I'm missing something but I don't find the code block you mentioned.

GuoRuiLv:
Hi @gorbunkov , The issue is moved from jmix-awsfs ,   the code link  is : https://github.com/jmix-framework/jmix/blob/e22ac4cca2b3ebbea23254066d82d9e442275b93/jmix-awsfs/awsfs/src/main/java/io/jmix/awsfs/AwsFileStorage.java#L196

gorbunkov:
Fixed in #3212

PR #3212, opened 2024-04-28
Improvment awsfs save stream

rassafel:
- Reduce the number of requests to upload small files
- Replace BufferedInputStream.available() (#327)
- Reduce memory allocation for multipart upload
- Make upload bucket name thread safe when refresh s3 client


gorbunkov:
@rassafel Thank you for the contribution!

PR #3238, opened 2024-05-08
Using DTO entities and non-persistent attributes in filter components

KremnevDmitry:
This implementation removes the restriction on JPA properties for the filter. Thus, entities and non-persistent properties are added to a set of meta property paths available for filtering.

E.g.: DTO Entity:
```java
@JmixEntity
public class TestEntity {
    
    @JmixGeneratedValue
    @JmixId
    private UUID id;

    @InstanceName
    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public UUID getId() {
        return id;
    }

    public void setId(UUID id) {
        this.id = id;
    }
}
````

The properties become available in the `Add Condition` dialog:
image

When you select any property, the **query** is modified in the data loader. This information is passed to the `LoadContext`, which can be processed in the `LoadDelegate`:
image

The you can use this information to process a set of non-persistent entities.
For example:
```java
    @Install(to = "testEntitiesDl", target = Target.DATA_LOADER)
    protected List<TestEntity> testEntitiesDlLoadDelegate(LoadContext<TestEntity> loadContext) {
        LoadContext.Query query = loadContext.getQuery();
        if (query == null) {
            return Collections.emptyList();
        }

        return entities.stream()
                .skip(query.getFirstResult())
                .limit(query.getMaxResults())
                .filter(testEntity -> isAcceptable(testEntity, query.getCondition()))
                .collect(Collectors.toList());
    }

    protected boolean isAcceptable(TestEntity entity, Condition condition) {
        // implement custom filter logic
        if (condition instanceof LogicalCondition logicalCondition
                && logicalCondition.getConditions().size() == 1
                && logicalCondition.getConditions().get(0) instanceof PropertyCondition propertyCondition
                && propertyCondition.getOperation().equals(PropertyCondition.Operation.CONTAINS)
                && "name".equals(propertyCondition.getProperty())
                && propertyCondition.getParameterValue() != null) {
            return entity.getName().contains(String.valueOf(propertyCondition.getParameterValue()));
        }

        return true;
    }
```

Ideally, we need to write a converter of `io.jmix.core.querycondition.Condition` into some kind of predicate.
This example uses a hardcoded check for the `name contains` operation.

> **Note:** sorting and pagination for DTO entities works out of the box

The test project attached: [dto-filter.zip](https://github.com/jmix-framework/jmix/files/15249866/dto-filter.zip)

### DIFF
--- a/jmix-data/data/src/main/java/io/jmix/data/impl/JpqlQueryBuilder.java
+++ b/jmix-data/data/src/main/java/io/jmix/data/impl/JpqlQueryBuilder.java
@@ -24,7 +24,10 @@ import io.jmix.core.common.util.StringHelper;
 import io.jmix.core.impl.QueryParamValuesManager;
 import io.jmix.core.metamodel.model.MetaClass;
 import io.jmix.core.metamodel.model.MetaProperty;
+import io.jmix.core.metamodel.model.MetaPropertyPath;
 import io.jmix.core.querycondition.Condition;
+import io.jmix.core.querycondition.LogicalCondition;
+import io.jmix.core.querycondition.PropertyCondition;
 import io.jmix.data.JmixQuery;
 import io.jmix.data.QueryTransformer;
 import io.jmix.data.QueryTransformerFactory;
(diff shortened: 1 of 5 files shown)


knstvk:
@KremnevDmitry - Add an application property to switch off showing non-JPA attributes in filters: `jmix.ui.component.filter-show-non-jpa-properties`. It should be true by default.
We will set it to false when migrating projects, otherwise some filter properties can unintentionally appear in existing views.

jmix-data/data/src/main/java/io/jmix/data/impl/JpqlQueryBuilder.java
```diff
@@ -265,8 +268,45 @@ protected void applyFiltering() {
                         .processParameters(resultParameters, queryParameters, actualized, entityName);
             }
 
+            Condition jpaOnlyCondition = removeNonJpaPropertyConditions(actualized);
```
knstvk, line 271: This is required when a filter for JPA entity contains conditions by non-JPA attributes. They should not be included in JPQL.

KremnevDmitry:
> @KremnevDmitry - Add an application property to switch off showing non-JPA attributes in filters: `jmix.ui.component.filter-show-non-jpa-properties`. It should be true by default. We will set it to false when migrating projects, otherwise some filter properties can unintentionally appear in existing views.

Got it
Migration issue for the Studio will be created

KremnevDmitry:
@knstvk, application property has been added

You can find studio issue here: [JST-4959](https://youtrack.jmix.io/issue/JST-4959/Migrator-to-Jmix-23-should-add-filter-show-non-jpa-properties-application-property)

Issue #2707, opened 2023-12-29
XML descriptor for composite component


knstvk:
### Feature Description

Composite components are important building blocks in large applications. We should provide a way to describe their layout in XML descriptors similar to views.

As an additional benefit, a composite component with a XML descriptor will simplify creating internal layout of [virtualList](https://demo.jmix.io/ui-samples/sample/virtual-list-simple) component.

See also [forum discussion](https://forum.jmix.io/t/does-flow-ui-support-fragments/2607).

glebfox:
Implemented in #841

PR #3268, opened 2024-05-21
XML descriptor for composite component (2707)

glebfox:
Fixes: #2707

## Jmix CompositeComponent

New base class for composite components: `CompositeComponent`.

A composite encapsulates a `Component` tree to allow creation of new components by composing existing components. By encapsulating the component, its API can be hidden or presented in a different way for the user of the composite.

The encapsulated component tree is available through `getContent()`. Composite will by default look at the generic type declaration of its subclass to find the content type and create an instance using `UiComponents`. You can also override `initContent()` to manually create the component tree or define components tree using XML markup and bind it to a composite component using `CompositeDescriptor` annotation.

Composite is a way to hide API on the server side. It does not contribute any element to the `Element` tree.

`CompositeComponent` extends Vaadin `com.vaadin.flow.component.Composite` to have specific fixes, e.g. see `com.vaadin.flow.component.Composite#setContent` comment.

### @CompositeDescriptor annotation

Specifies a string value that is a file path to an xml descriptor that can be used for a `CompositeComponent` initialization.  If the **value** contains a file name only (i. e. don't start with `/`), it is assumed that the file is located in the **package** of the composite component class.

**Note:** if it's assumed that composite component can be extended, the file path should be defined as full path.

### Supported elements

There are two root elements:

1. `content` - required element that contains composite layout (similar to view `layout` element). Since the root component of Composite can be any component, `content` has no attributes and doesn't represent any component.
2. `actions` - optional composite actions (similar to view `actions` element). Actions defined for composite component have shortcut bound to root composite component, i.e. shortcut can be triggered only if focus within composite layout.

```
<composite xmlns="http://jmix.io/schema/flowui/composite">
    <actions>
        <action id="testAction" text="Test" icon="PLUS"/>
    </actions>
    <content>
        <vbox id="contentBox" classNames="overflow-auto">
            <hbox id="buttonsPanel" classNames="buttons-panel">
                <button id="createBtn" action="dataGrid.create"/>
                <button id="editBtn" action="dataGrid.edit"/>
                <button id="removeBtn" action="dataGrid.remove"/>
                <button id="testBtn" action="testAction"/>
                <gridColumnVisibility dataGrid="dataGrid" icon="COG_O"/>
            </hbox>
            <dataGrid id="dataGrid"
                      width="100%"
                      minHeight="20em"
                      metaClass="test_Product">
                <actions>
                    <action id="create" type="list_create"/>
                    <action id="edit" type="list_edit"/>
                    <action id="remove" type="list_remove"/>
                </actions>
                <columns resizable="true">
                    <column property="name"/>
                    <column property="price"/>
                </columns>
            </dataGrid>
        </vbox>
    </content>
</composite>
```

**Note:** Nested components also support their actions, e.g. `DataGrid`.

### Unsupported elements

* `facets` - Closely connected with Views
* `data` - UI component shouldn't load data itself + Composite doesn't have enough lifecycle events + DataLoadCoordinator facet isn't available

### Specifics of defining a layout in XML

To avoid **id** duplicates in components tree when several composite components of the same type is used simultaneously or composite inner components have the same **id** as other components in the components tree, the value of `id` attribute of each component defined in composite descriptor (so called **composite id** or Shadow XML id) is stored in the component `Attributes` field instead of setting it as **actual id** using `Component. setId(String)`. 

## CompositeComponent API

* `findInnerComponent` / `getInnerComponent` - return the inner component with given **composite id**. These methods search among components added via an XML descriptor only.
* `PostInitEvent` - The event that is fired after the composite component and all its declaratively defined inner components are created and fully initialized. In this event listener, you can make final configuration of the composite component and its inner components, e.g. obtain inner components using `getInnerComponent(String)` and add their specific event listeners. 

**Note:** `PostInitEvent` is not fired if components tree is created without using XML descriptor, i.e. by overriding `initContent()` to manually create the component tree.


glebfox:
Obsolete, will be implemented in https://github.com/jmix-framework/jmix/pull/3336

Issue #841, opened 2022-07-18
Fragments


glebfox:
## Fragment

New base class for fragments: `Fragment`.

A fragment encapsulates a `Component` tree to allow creation of new components by composing existing components. The main purpose of fragments is to be used as a part of views and other fragments. By encapsulating the component, its API can be hidden or presented in a different way for the user of the fragment.

The encapsulated component tree is available through `getContent()`. Fragment will by default look at the generic type declaration of its subclass to find the content type and create an instance using `UiComponents`. You can also override `initContent()` to manually create the component tree or define components tree using XML markup and bind it to a fragment using `FragmentDescriptor`.

Fragment is a way to hide API on the server side. It does not contribute any element to the `Element` tree.

`Fragment` extends Vaadin `com.vaadin.flow.component.Composite` to have specific fixes, e.g. see `com.vaadin.flow.component.Composite#setContent` comment.

### @FragmentDescriptor annotation

Specifies a string value that is a file path to an xml descriptor that can be used for a `Fragment` initialization.  If the **value** contains a file name only (i. e. don't start with `/`), it is assumed that the file is located in the **package** of the fragment class.

**Note:** if it's assumed that fragment can be extended, the file path should be defined as full path.

### Supported elements

There are three root elements:

1. `content` - required element that contains fragment layout (similar to view `layout` element). Since the root component of Fragment can be any component, `content` has no attributes and **doesn't represent any component**.
2. `actions` - optional fragment actions element (similar to view `actions` element). Actions defined for fragment have shortcut bound to root fragment, i.e. shortcut can be triggered only if focus within fragment layout.
3. `data` - optional fragment data element (similar to view `data` element). A fragment can have its own data containers and loaders, defined in the data XML element. At the same time, the framework creates a single instance of `DataContext` for the view and all its fragments. Therefore all loaded entities are merged to the same context and their changes are saved when the host view is committed. Data components defined with `provided="true"` attribute which means that the container with the same id must exist in a host screen or enclosing fragment, i.e it must be provided from outside.

```
<fragment xmlns="http://jmix.io/schema/flowui/fragment">
    <data>
        <collection id="productsDc" provided="true"
                    class="test_support.entity.sales.Product"/>
    </data>
    <actions>
        <action id="testAction" text="msg://testAction.text" icon="PLUS"/>
    </actions>
    <content>
        <vbox id="contentBox" classNames="overflow-auto">
            <hbox id="buttonsPanel" classNames="buttons-panel">
[...]

PR #3271, opened 2024-05-21
Generic Component XML Element (841)

glebfox:
Fixes: #841

## Generic Component XML  element 

Added `component` xml element that is used to create any UI component by FQN. In order to define component properties the nested `property` elements can be used (similar to actions). 

For example:

```
<component id="button" component="com.vaadin.flow.component.button.Button">
    <properties>
        <property name="text" value="Button"/>
        <property name="icon" value="PLUS" type="ICON"/>
    </properties>
</component>

<component id="propertyFilter" component="io.jmix.flowui.component.propertyfilter.PropertyFilter">
    <properties>
        <property name="property" value="name"/>
        <property name="operation" value="CONTAINS"/>
        <property name="dataLoader" value="productsDl" type="LOADER_REF"/>
        <property name="parameterName" value="productName"/>
        <property name="label" value="Property Filter"/>
        <property name="tabIndex" value="2"/>
    </properties>
</component>

<component id="dataGridPanel" component="component.composite.component.TestDataGridPanel">
    <properties>
        <property name="dataContainer" value="productsDc" type="CONTAINER_REF"/>
    </properties>
</component>
```

### Properties limitations and specifics

* The `name` attribute value is converted to setter method name, e.g. `text` -> `setText`. 
* The first method with this name and a single input parameter is used to set the value.
   * (limitation) if there are several methods with the same name, no guarantee that the one you need is used to set the value
* The found method input parameter type is used to convert string value of the `value` attribute into actual value.
* The `type` attribute is used to indicate that string value representation must be converted by one of pluggable `ComponentPropertyParser` beans.

### ComponentPropertyParser

As been mentioned above, if string value cannot be parsed using input parameter type, e.g. icon which is a `Component` or data container which is defined by its id, pluggable `ComponentPropertyParser` beans is used. 

Currently there are three implementations:
* `IconComponentPropertyParser` - converts string into a new `Icon` instance with the icon determined by the passed string. If a passed string contains `:` delimiter then a new `Icon` is created using **icon collection** and **icon name** values, otherwise the passed string is considered as `VaadinIcon` constant name.
* `DataContainerComponentPropertyParser` - returns **data container** with the given `id` from corresponding view's `ViewData` object.
* `DataLoaderComponentPropertyParser` - returns **data loader** with the given `id` from corresponding view's `ViewData` object.

### DIFF
--- a/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/StudioPropertyType.java
+++ b/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/StudioPropertyType.java
@@ -91,6 +91,11 @@ public enum StudioPropertyType {
     ENUM_CLASS,
 
     /**
+     * FQN of UI Component class.
+     */
+    COMPONENT_CLASS,
+
+    /**
      * Value from a strict list of property options.
      */
     ENUMERATION,
--- a/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
+++ b/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
@@ -16,6 +16,7 @@
 
 package io.jmix.flowui.kit.meta.component;
 
+import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.Html;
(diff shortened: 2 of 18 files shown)


glebfox:
Obsolete, implemented in https://github.com/jmix-framework/jmix/pull/3336

PR #3296, opened 2024-05-29
feat: Add annotation configuration rest service

kernelmix:
Add annotation configuration rest service and Improve openapi documentation


knstvk (changes requested):
Hi @kernelmix 
Thank you for the contribution!

Let me suggest a few changes.
1. What about renaming annotations to more descriptive and concrete `@RestService`, `@RestMethod`?
2. The annotated methods can now be invoked only through HTTP POST. Let's either enable any method, as in XML-based methods, or add an explicit parameter to the annotation, like `String[] httpMethods() default {"GET", "POST"};`

knstvk (approved):

PR #3318, opened 2024-05-30
Lists of enumerations in UI should include enumerations that are not used in entity attributes

knstvk:


jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -708,20 +718,22 @@ public Collection<MetaClass> getAllJpaEmbeddableMetaClasses() {
...
+                Set<String> classNames = classpathScanner.getClassNames(EnumDetector.class);
+                for (String className : classNames) {
+                    try {
+                        Class<?> enumClass = ReflectionHelper.loadClass(className);
+                        tmpEnums.add(enumClass);
+                    } catch (ClassNotFoundException e) {
+                        log.warn("Cannot load enum class {}", className, e);
                     }
                 }
+                enums = tmpEnums;
```
gorbunkov, line 736: Maybe we should return unmodifiable set?

```suggestion
                enums = Collections.unmodifiableSet(tmpEnums);
```

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -708,20 +718,22 @@ public Collection<MetaClass> getAllJpaEmbeddableMetaClasses() {
...
         if (enums == null) {
             synchronized (this) {
-                enums = new HashSet<>();
-                for (MetaClass metaClass : metadata.getSession().getClasses()) {
-                    for (MetaProperty metaProperty : metaClass.getProperties()) {
-                        if (metaProperty.getRange().isEnum()) {
-                            Class<?> c = metaProperty.getRange().asEnumeration().getJavaClass();
-                            enums.add(c);
-                        }
+                Set<Class<?>> tmpEnums = new HashSet<>();
```
gorbunkov, line 726: It is not directly related to the current fix, but since we're inside this method, it looks that the second null check statement is missing inside the synchronized block.

```java
        if (enums == null) {
            synchronized (this) {
                if (enums == null) {
```

gorbunkov (approved):

jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
knstvk: Makes sense. Done.

Issue #3286, opened 2024-05-24
Unclear default visibility for loginForm elements


iudaev:
### Environment

Jmix version: <2.2.3>

### Bug Description

Default values for `forgotPasswordButtonVisibility`, `localesVisible`, and `remeberMeVisible` doesn't align with the values in the inspector.

### Current Behavior

`<loginForm/>` produces a login form with "Remember me" checkbox, locales combobox, and "Forgot password" button despite their visibility attributes are unchecked. It would be more appropriate if unchecked attributes were defaulted to false.

![image](https://github.com/jmix-framework/jmix/assets/128021710/2f0123cf-30e6-4e8a-bce7-a52cec2de742)

### Expected Behavior

`<loginForm/>` produces a login form **without** "Remember me" checkbox, locales combobox, and "Forgot password" button when their visibility attributes are unchecked.

![image](https://github.com/jmix-framework/jmix/assets/128021710/761adf9b-74d8-49c8-87e1-76d7d3094e46)

### Changes
`localesVisible` default value was changed (`<loginForm/>` element property).
Was: `true`
Now: `false`

For QA

- Since we have changed the default value, we need to make sure that users of versions prior to 2.3.0 still have locales visible when they upgrade to 2.3.1. If they don't have a value set in their descriptor, the locales should be visible; if a value is set, the visibility should match it. For a new project (2.3.1) the behaviour is the same.
- Check all other `<loginForm/>` attributes. They should not change after the migration.
- Check the value of the `<loginFrom/>` attributes in the descriptor and in the inspector. They should match.
- Check the localized `<form/>` and `<errorMessage/>` strings: all messages should be initialized correctly.

PR #3328, opened 2024-06-03
Unclear default visibility for loginForm elements 3286

ilushaaa:
#3286


jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractLoginFormLoader.java
```diff
@@ -28,7 +28,9 @@ public abstract class AbstractLoginFormLoader<C extends AbstractLogin> extends A
     public void loadComponent() {
         componentLoader().loadEnabled(resultComponent, element);
         componentLoader().loadClassNames(resultComponent, element);
-        loadBoolean(element, "forgotPasswordButtonVisible", resultComponent::setForgotPasswordButtonVisible);
+        loadBoolean(element, "forgotPasswordButtonVisible").ifPresentOrElse(
```
glebfox, line 31: re-do

glebfox (changes requested):

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/EnhancedLoginForm.java
```diff
@@ -212,6 +213,12 @@ public void setLocaleItemLabelGenerator(@Nullable Function<Locale, String> local
         this.localeItemLabelGenerator = localeItemLabelGenerator;
     }
 
+    @Override
+    public void setI18n(LoginI18n i18n) {
```
Flaurite, line 217: Do we need to override this method? It passes the same value.

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
```diff
@@ -30,14 +30,16 @@ public class JmixLoginI18n extends LoginI18n {
...
+
     /*
      * CAUTION! Copied from com.vaadin.flow.component.login.LoginI18n
      */
     static {
         try {
             final JsonFactory JSON_FACTORY = new JreJsonFactory();
             JMIX_DEFAULT_I18N = JSON_FACTORY.parse(
-                    IOUtils.toString(LoginI18n.class.getResource("i18n.json"),
+                    IOUtils.toString(JmixLoginI18n.class.getResource("i18n.json"),
```
Flaurite, line 42: Since we have fully overrided this object, why we should use JSON file for storing defaults? Would not be simpler to define them in `createDefault()` while building an object?

Flaurite (changes requested):

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
```diff
@@ -47,10 +49,22 @@ public class JmixLoginI18n extends LoginI18n {
...
+
+    public JmixForm getJmixForm() {
+        return jmixForm;
+    }
+
+    public void setJmixForm(JmixForm jmixForm) {
+        this.jmixForm = jmixForm;
     }
 
-    public static class JmixForm extends Form {
```
glebfox, line 53: why? You've made a breaking change that produces compilation error

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/EnhancedLoginForm.java
```diff
@@ -21,9 +21,12 @@
 import com.vaadin.flow.component.login.LoginForm;
 import com.vaadin.flow.internal.JsonSerializer;
 import com.vaadin.flow.shared.Registration;
-
 import jakarta.annotation.Nullable;
-import java.util.*;
+
+import java.util.ArrayList;
```
glebfox, line 26: Not related changes

jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/AbstractLoginFormLoader.java
```diff
@@ -58,25 +58,24 @@ protected void loadHeader(C resultComponent, Element element, LoginI18n i18n) {
         }
     }
 
-    protected void loadForm(C resultComponent, Element element, LoginI18n i18n) {
+    protected void loadForm(C resultComponent, Element element, JmixLoginI18n i18n) {
         Element formElement = element.element("form");
         if (formElement != null) {
-            JmixLoginI18n.JmixForm form = new JmixLoginI18n.JmixForm();
+            JmixLoginI18n.JmixForm jmixForm = new JmixLoginI18n.JmixForm();
```
glebfox, line 64: unnecessary renaming

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
```diff
@@ -2388,12 +2388,15 @@ public interface StudioComponents {
...
+                            defaultValue = "true"),
                     @StudioProperty(xmlAttribute = "id", type = StudioPropertyType.COMPONENT_ID),
-                    @StudioProperty(xmlAttribute = "rememberMeVisible", type = StudioPropertyType.BOOLEAN),
+                    @StudioProperty(xmlAttribute = "rememberMeVisible", type = StudioPropertyType.BOOLEAN,
+                            defaultValue = "true"),
                     @StudioProperty(xmlAttribute = "enabled", type = StudioPropertyType.BOOLEAN,
                             defaultValue = "true"),
-                    @StudioProperty(xmlAttribute = "localesVisible", type = StudioPropertyType.BOOLEAN),
+                    @StudioProperty(xmlAttribute = "localesVisible", type = StudioPropertyType.BOOLEAN,
+                            defaultValue = "false"),
```
glebfox, line 2399: it's still visible, even if `true` is not set explicitly. 

When you fix it, the other problem will appear that you need to set it to `true` explicitly, otherwise all existing views will be broken. Let's set it to true if `io.jmix.flowui.kit.component.loginform.EnhancedLoginForm#setLocaleItems` with not empty value to avoid breaking changes

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
```diff
@@ -50,6 +52,14 @@ public static JmixLoginI18n createDefault() {
         return JsonSerializer.toObject(JmixLoginI18n.class, JMIX_DEFAULT_I18N);
     }
 
+    public void setForm(JmixForm form) {
```
glebfox, line 55: if we call `com.vaadin.flow.component.login.LoginI18n#setForm`, then `getForm()` will return `null`. Override super `setForm` and check that passed parameter is instance of `JmixForm`, otherwise throw illegal argument exception

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
```diff
@@ -50,6 +52,14 @@ public static JmixLoginI18n createDefault() {
...
+        this.form = form;
+    }
+
+    public JmixForm getForm() {
```
glebfox, line 59: @Override

glebfox (changes requested):

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
ilushaaa: We can create new JmixForm instance and init it by LoginI18n.getForm. In this case rememberMe field will be empty. What is correct.

If we override setForm, then we'll get that IllegalState exception. Because the argument of the method has type LoginI18n.Form, and exactly this type will be created when converting Json to Object.

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
ilushaaa: If we set localesVisible to true depending on the number of available locales - do we need this property at all?

loginForm.localesSelect.hidden = locales.length > 1 ? false : true;

The locales select with single item doesn't have any matter.

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/component/loginform/JmixLoginI18n.java
```diff
@@ -17,37 +17,46 @@
...
+
+        return jmixLoginI18n;
+    }
+
+    @Override
+    public void setForm(Form form) {
+        if (form instanceof JmixForm jmixForm) {
+            this.form = jmixForm;
+        } else {
+            throw new IllegalStateException("Setter doesn't support value of %s" + form.getClass());
```
glebfox, line 53: Exception should indicate what type of value is expected. 

```
@Override
public void setForm(Form form) {
    Preconditions.checkArgument(form instanceof JmixForm, "Passed value must be an instance of " + JmixForm.class.getSimpleName());
    this.form = ((JmixForm) form);
}
```

glebfox (changes requested):

glebfox (approved):

PR #3342, opened 2024-06-04
Change CREATED_BY to varchar(255) to reflect username size being the same

Jojo134:
In our project we served an error in the following lib regarding the createdBy size if the name exceeds 50 chars. For example in emails. This should fix the table for saving the flowui settings for a user.


glebfox (approved):

glebfox:
Thank you for your contribution. Also, I've created an issue to fix other entities: #3351

Issue #2482, opened 2023-11-16
Error starting application if additional data store has no entities


Gasloff:
Jmix 2.1.0

Create new Java project
Add additional datastore (tested for PostgreSQL & MSSQL)
Run application
AR: bootRun fails
```
Error starting ApplicationContext. To display the condition evaluation report re-run your application with 'debug' enabled.
2023-11-16T11:50:47.729+04:00 ERROR 31172 --- [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'pglEntityManagerFactory' defined in class path resource [com/company/untitled31/PglStoreConfiguration.class]: Index 0 out of bounds for length 0
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1770) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:598) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:520) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:325) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:323) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199) ~[spring-beans-6.0.12.jar:6.0.12]
	at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1158) ~[spring-context-6.0.12.jar:6.0.12]
[...]

PR #3355, opened 2024-06-06
Error starting application if additional data store has no entities jmix-framework/jmix#2482

dtaimanov:


jmix-data/data/src/main/java/io/jmix/data/impl/JmixBaseEntityManagerFactoryBean.java
```diff
@@ -45,9 +45,11 @@ protected void setupPersistenceUnit() {
                 return;
             }
         }
-        log.warn("Cannot find persistence.xml for '{}' store. Falling back to classpath scan for entity classes.", storeName);
-        setPersistenceUnitName(storeName);
-        setPackagesToScan("");
+
+        throw new RuntimeException(String.format("Cannot find persistence.xml for '%1$s' store.\n" +
+                "Please make sure that property 'jmix.core.additional-stores' defined in application.properties " +
+                        "and contains this store name or at least one entity with '@Store(name = \"%1$s\")' annotation present.",
```
gorbunkov, line 51:
```suggestion
                "Please make sure that the property 'jmix.core.additional-stores' is defined in application.properties " +
                        "and contains this store name, or that at least one entity with the '@Store(name = \"%1$s\")' annotation is present.",
```

jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/EnhancingAction.groovy
```diff
@@ -71,6 +76,42 @@ class EnhancingAction implements Action<Task> {
         }
     }
 
+    protected void addAbsentEmptyStores(ClassesInfo classesInfo, SourceSet sourceSet, Project project) {
```
gorbunkov, line 79: Pls, add javadoc for this method. It's not fully clear from this method name:
* add where?
* what store is "empty"?
* absent "where"?

gorbunkov (changes requested):

jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/EnhancingAction.groovy
```diff
@@ -71,6 +76,47 @@ class EnhancingAction implements Action<Task> {
         }
     }
 
+    /**
+     * Completes {@code classesInfo} with "empty" data stores in order to `persistence.xml` and `orm.xml` descriptors will be generated for them.
+     * "Empty" store is a store without entities. Such stores will not be added to {@code classesInfo} by
+     * the {@link io.jmix.gradle.EnhancingAction#collectClasses} discovery method.
```
gorbunkov, line 82:
```suggestion
     * Completes {@code classesInfo} with "empty" data stores so that `persistence.xml` and `orm.xml` descriptors will be generated for them.
     * An "empty" store is a store without entities. Such stores will not be added to {@code classesInfo} by
     * the {@link io.jmix.gradle.EnhancingAction#collectClasses} discovery method.
```

gorbunkov (approved):

Issue #3533, opened 2024-07-31
IllegalArgumentException while editing a policy in the resource role editor


gorbunkov:
### Environment

Jmix version: 2.3.1

### Bug Description

Exception appears while editing the resource level policy:

```
java.lang.IllegalArgumentException: Null reference passed as parameter
	at io.jmix.core.common.util.Preconditions.checkNotNullArgument(Preconditions.java:87) ~[jmix-core-2.3.1.jar:na]
	at io.jmix.core.EntityStates.isLoadedWithFetchPlan(EntityStates.java:286) ~[jmix-core-2.3.1.jar:na]
	at io.jmix.flowui.view.StandardDetailView.doNotReloadEditedEntity(StandardDetailView.java:635) ~[jmix-flowui-2.3.1.jar:na]
	at io.jmix.flowui.view.StandardDetailView.setupEntityToEdit(StandardDetailView.java:610) ~[jmix-flowui-2.3.1.jar:na]
	at io.jmix.flowui.view.StandardDetailView.setEntityToEdit(StandardDetailView.java:543) ~[jmix-flowui-2.3.1.jar:na]
	at io.jmix.flowui.view.builder.DetailWindowBuilderProcessor.build(DetailWindowBuilderProcessor.java:89) ~[jmix-flowui-2.3.1.jar:na]
	at io.jmix.flowui.view.builder.AbstractWindowBuilder.build(AbstractWindowBuilder.java:113) ~[jmix-flowui-2.3.1.jar:na]
	at io.jmix.securityflowui.view.resourcerole.ResourceRoleModelDetailView.buildResourcePolicyDetailView(ResourceRoleModelDetailView.java:356) ~[jmix-security-flowui-2.3.1.jar:na]
	at io.jmix.securityflowui.view.resourcerole.ResourceRoleModelDetailView.onResourcePoliciesTableEdit(ResourceRoleModelDetailView.java:327) ~[jmix-security-flowui-2.3.1.jar:na]
```

### Steps To Reproduce

1. Create new resource role
2. Add new entity attribute policy
3. Save the role
4. Open the role
5. Open the policy editor, change some value, e.g. Action
6. Click OK
7. Open the policy editor again

### Current Behavior

Exception described in the issue description appears.

### Expected Behavior

Policy editor is opened with no errors.

PR #3575, opened 2024-08-14
IllegalArgumentException while editing a policy in the resource role editor

Gavrilov-Ivan:
See #3533

Policy is a DTO and doesn't have fetch plan - there should be no check for `.isLoadedWithFetchPlan` in this case

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/StandardDetailView.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/StandardDetailView.java
@@ -632,7 +632,12 @@ public class StandardDetailView<T> extends StandardView implements DetailView<T>
     private boolean doNotReloadEditedEntity() {
         if (isEntityModifiedInParentContext()) {
             InstanceContainer<T> container = getEditedEntityContainer();
-            return getEntityStates().isLoadedWithFetchPlan(entityToEdit, container.getFetchPlan());
+            FetchPlan fetchPlan = container.getFetchPlan();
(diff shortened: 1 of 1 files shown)


glebfox (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/StandardDetailView.java
```diff
@@ -632,7 +632,12 @@ protected void setupEntityToEdit(T entityToEdit) {
     private boolean doNotReloadEditedEntity() {
         if (isEntityModifiedInParentContext()) {
             InstanceContainer<T> container = getEditedEntityContainer();
-            return getEntityStates().isLoadedWithFetchPlan(entityToEdit, container.getFetchPlan());
+            FetchPlan fetchPlan = container.getFetchPlan();
+            if (fetchPlan == null) {
```
glebfox, line 636: Probably we need to call `io.jmix.core.MetadataTools#isJpaEntity(io.jmix.core.metamodel.model.MetaClass)` to check if it's really DTO entity?

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/StandardDetailView.java
Gavrilov-Ivan: Done

glebfox (approved):

Issue #3487, opened 2024-07-17
Add Settings facet to Reports Run view


knstvk:
Originates from [forum topic](https://forum.jmix.io/t/order-reports-run/5203).

gorbunkov:
Group may be empty. We can add the **settings** facet to the screen.

SergeiAksenov2:
Tested on:
1) Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6700-233
IntelliJ version: IntelliJ IDEA 2023.3.7 (Community Edition)
2) Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6702-241
IntelliJ version: IntelliJ IDEA 2024.1.4 (Community Edition)

Settings facet is working on the Reports Run view.

PR #3581, opened 2024-08-14
Add Settings facet to Reports Run view

ilushaaa:
#3487

### DIFF
--- a/jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/view/run/ReportRunView.java
+++ b/jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/view/run/ReportRunView.java
@@ -171,7 +171,8 @@ public class ReportRunView extends StandardListView<Report> {
         reportsDc.setItems(reports);
     }
 
-    protected @Nullable Sort getReportGridSort() {
+    @Nullable
+    protected Sort getReportGridSort() {
         if (reportDataGrid.getSortOrder().isEmpty()) {
             return null;
         }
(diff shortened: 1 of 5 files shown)


jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/view/run/ReportRunView.java
```diff
@@ -174,6 +171,18 @@ protected void filterReports() {
         reportsDc.setItems(reports);
     }
 
+    protected @Nullable Sort getReportGridSort() {
```
glebfox, line 174: We usually place `@Nullable` above method:
```
@Nullable
protected Sort getReportGridSort() {
```

jmix-reports/reports/src/main/java/io/jmix/reports/ReportSecurityManager.java
```diff
@@ -113,10 +113,11 @@ public void applyPoliciesByEntityParameters(LoadContext lc, @Nullable MetaClass
      * @param screenId            - id of the screen
      * @param user                - caller user
      * @param inputValueMetaClass - meta class of report input parameter
-     *
+     * @param sort                - sorting type for the reports list
      * @return list of available reports
      */
-    public List<Report> getAvailableReports(@Nullable String screenId, @Nullable UserDetails user, @Nullable MetaClass inputValueMetaClass) {
+    public List<Report> getAvailableReports(@Nullable String screenId, @Nullable UserDetails user,
```
glebfox, line 119: You cannot change public API like this, it's breaking changes. Add a new method with `Sort` parameter and call it from all one. Also it will let you avoid passing `null` explicitly

glebfox (changes requested):

glebfox (approved):

Issue #3586, opened 2024-08-14
Generic filter fails with dynamic attributes


fractal3000:
### Environment

Jmix version: 2.3.1

### Bug Description

Generic filter fails with dynamic attributes

### Steps To Reproduce

1. Create any dynamic attribute
2. Add it to list view
3. Try to create filter condition of generic filter with this create dinamic attribute
4. Copy URL of the screen and open it in other browser tab

### Current Behavior

Runtime error with closing list view with filter.
### Expected Behavior

Filter works correctly.

konyashkina:
Tested on
Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6714-241
IntelliJ version: IntelliJ IDEA 2024.1.1

PR #3601, opened 2024-08-15
Generic filter fails with dynamic attributes #3586

fractal3000:


jmix-core/core/src/main/java/io/jmix/core/metamodel/model/MetaClass.java
```diff
@@ -69,6 +69,10 @@ public interface MetaClass extends MetadataObject {
 
     /**
      * Returns MetaPropertyPath object, representing path to the property from the current class
+     * Note that this method don't take into account extensions(e.g. Dynamic Attributes add-on) and it returns null
+     * in case of not static attributes(attributes from extension).
+     * For getting all attributes (including attributes from extensions) use {@link io.jmix.core.MetadataTools#resolveMetaPropertyPath(MetaClass, String)}
```
glebfox, line 74: > 120 symbols

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
```diff
@@ -243,7 +247,7 @@ protected List<FilterComponent> deserializeConditions(List<String> conditionPara
     protected boolean isPermitted(DataLoader dataLoader, FilterComponent filterComponent) {
         if (filterComponent instanceof PropertyFilter<?> propertyFilter && propertyFilter.getProperty() != null) {
             MetaClass entityMetaClass = dataLoader.getContainer().getEntityMetaClass();
-            MetaPropertyPath propertyPath = entityMetaClass.getPropertyPath(propertyFilter.getProperty());
+            MetaPropertyPath propertyPath = metadataTools.resolveMetaPropertyPathOrNull(entityMetaClass, propertyFilter.getProperty());
```
glebfox, line 250: Remove explicit passing of `MetadataTools` in all constructors. It's a breaking change, because you change public API. You have `ApplicationContext` use it to obtain required bean. If a bean is intended to be used several times, make it lazy initialized, see `SingleFilterSupport` as an example.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
```diff
@@ -76,6 +77,7 @@ public class GenericFilterUrlQueryParametersBinder extends AbstractUrlQueryParam
     protected String conditionParam;
 
     protected ApplicationContext applicationContext;
+    private final MetadataTools metadataTools;
```
glebfox, line 80: 1. we don't use private
2. follow code style of the class

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
fractal3000: Fixed. The field was removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
```diff
@@ -262,6 +263,10 @@ protected boolean isPermitted(DataLoader dataLoader, FilterComponent filterCompo
         return true;
     }
 
+    private MetadataTools recieveMetadataTools() {
```
glebfox, line 266: 1. We **do not** use private modifier
2. Use the same naming as for other similar methods, either `getFoo()` or `foo()`
3. There is no point in this method right now, because it calls `applicationContext.getBean(MetadataTools.class)` every time. Take a look at `singleFilterSupport` in this class and do the same - lazy initialization.
4. Place your method near similar `getSingleFilterSupport()` since they are logically the same.

glebfox (changes requested):

jmix-core/core/src/main/java/io/jmix/core/metamodel/model/MetaClass.java
glebfox: This line still > 120 symbols

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
fractal3000: fixed, but it is strange practice to use only protected modifiers. It obviously conflicts with Open-Close principle of SOLID principes.

glebfox (approved):

Issue #375, opened 2021-08-05
Problem with setting default value to dynamic attribute which is of collection type


maistrenkoIulia:
Jmix version: 1.0.0-SNAPSHOT
Jmix Studio plugin version: 1.1.SNAPSHOT3371-211
IntelliJ version: IntelliJ IDEA 2021.2 (Ultimate Edition)

TC:
create dynamic attribute with integer type with is collection = true
add to screens> apply changes
create new entity

AR:
```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.util.List (java.lang.Integer and java.util.List are in module java.base of loader 'bootstrap')
	at io.jmix.dynattrui.impl.AttributeDefaultValues.setDefaultAttributeValue(AttributeDefaultValues.java:70)
	at io.jmix.dynattrui.impl.AttributeDefaultValues.initDefaultAttributeValues(AttributeDefaultValues.java:52)
	at io.jmix.dynattrui.facet.DynAttrFacetImpl.initEntityInStandardEditor(DynAttrFacetImpl.java:62)
	at io.jmix.core.common.event.EventHub.publish(EventHub.java:170)
	at io.jmix.ui.screen.Screen.fireEvent(Screen.java:124)
	at io.jmix.ui.screen.StandardEditor.setupEntityToEdit(StandardEditor.java:218)
	at io.jmix.ui.screen.StandardEditor.beforeShow(StandardEditor.java:155)
	at io.jmix.core.common.event.EventHub.publish(EventHub.java:170)
	at io.jmix.ui.screen.Screen.fireEvent(Screen.java:124)
	at io.jmix.ui.screen.UiControllerUtils.fireEvent(UiControllerUtils.java:58)
	at io.jmix.ui.sys.ScreensImpl.fireScreenBeforeShowEvent(ScreensImpl.java:1358)
	at io.jmix.ui.sys.ScreensImpl.show(ScreensImpl.java:354)
	at io.jmix.ui.screen.Screen.show(Screen.java:306)
	at io.jmix.ui.action.list.CreateAction.execute(CreateAction.java:358)
	at io.jmix.ui.action.list.CreateAction.actionPerform(CreateAction.java:300)
	at io.jmix.ui.component.impl.ButtonImpl.buttonClicked(ButtonImpl.java:75)
	at io.jmix.ui.widget.JmixButton.fireClick(JmixButton.java:77)
	at com.vaadin.ui.Button$1.click(Button.java:57)
[...]

PR #3627, opened 2024-08-20
Problem with setting default value to dinamic attribute which is of collection type #375

fractal3000:
…ollection type

### DIFF
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
@@ -57,6 +57,7 @@ import io.jmix.dynattrflowui.view.localization.AttributeLocalizationComponent;
 import io.jmix.flowui.*;
 import io.jmix.flowui.action.multivaluepicker.MultiValueSelectAction;
 import io.jmix.flowui.action.valuepicker.ValueClearAction;
+import io.jmix.flowui.component.UiComponentUtils;
 import io.jmix.flowui.component.checkbox.JmixCheckbox;
 import io.jmix.flowui.component.codeeditor.CodeEditor;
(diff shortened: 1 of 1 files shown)


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -1137,4 +1163,32 @@ protected void preCommitConfiguration() {
...
+    protected boolean isScalar() {
+        return !isCollectionField.getValue();
+    }
+
+    protected void setEnabledForDefaultValueControlsWithClearing(boolean isEnabled, boolean clearValues) {
+        DEFAULT_VALUE_COMPONENT_IDS
+                .forEach((key, value) -> setEnabledAndClear(key, isEnabled && value.get(), clearValues));
+    }
+
+    protected void setEnabledAndClear(String componentId, boolean isEnabled, boolean clearValues) {
```
glebfox, line 1186: Rename method and input parameters

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -1137,4 +1163,32 @@ protected void preCommitConfiguration() {
...
+        Component component = UiComponentUtils.getComponent(this, componentId);
+        ((HasEnabled)component).setEnabled(isEnabled);
```
glebfox, line 1188: 1. Better to check if a component implements an interface before cast (for both interfaces).
2. Format code

glebfox (changes requested):

glebfox:
@fractal3000 this solution prevents from creating **new** incorrect configurations in simple way, but it doesn't fixes existing configurations. 

In fact, we can define the default **collection** value using the same UI component as for editing such values.

Gavrilov-Ivan (changes requested):

PR #3630, opened 2024-08-21
Problem with setting default value to dinamic attribute which is of collection type #375

fractal3000:


jmix-dynattr/dynattr/src/main/java/io/jmix/dynattr/impl/CommonAttributeDefinition.java
```diff
@@ -229,32 +229,39 @@ public String getEnumeration() {
...
-                case STRING:
-                case ENUMERATION:
-                    return attribute.getDefaultString();
-                case ENTITY:
-                    if(attribute.getDefaultEntity() == null) {
-                        return null;
-                    }
-                    return attribute.getDefaultEntity().getObjectEntityId();
+            Object singleDefaultValue = getSingleDefaultValue(dataType);
+            if(!attribute.getIsCollection()){
```
Gavrilov-Ivan, line 233: Format code

Gavrilov-Ivan (changes requested):

jmix-dynattr/dynattr/src/main/java/io/jmix/dynattr/impl/CommonAttributeDefinition.java
fractal3000: Fixed

Gavrilov-Ivan (approved):

glebfox (approved):

PR #3640, opened 2024-08-22
REST DataStore

knstvk:
The purpose of the REST DataStore is to provide an easy way of integrating different Jmix applications. The result of the integration is the ability to access external entities from a remote Jmix application through the standard `DataManager` interface in the same way as local JPA entities. The external entities can be displayed in UI, updated and saved back to the remote application using the standard CRUD functionality provided by Jmix, without writing any specific code.

In the text below, we will use the following terms:
- Service Application - a Jmix application providing data through Generic REST API
- Client Application - a Jmix application consuming data from the Service Application using the REST DataStore

The service and client applications can use different versions of Jmix.

See usage examples:
- https://github.com/jmix-framework/jmix-integrated-apps-sample
- https://github.com/jmix-framework/jmix-separate-tiers-sample
- https://github.com/jmix-projects/sample-rest-ds-usage

See [jmix-restds/restds/README.md](https://github.com/jmix-framework/jmix/blob/feature/rest-ds/jmix-restds/restds/README.md) for how to run tests.

## Setup

In the service application project:
- Add Generic REST and Authorization Server add-ons
- Set up Client Credentials grant for a client

In the client application project:
- Add `implementation 'io.jmix.restds:jmix-restds-starter'` dependency

- Add an additional data store with `restds_RestDataStoreDescriptor` descriptor, for example:
```
jmix.core.additional-stores = customers  
jmix.core.store-descriptor-customers = restds_RestDataStoreDescriptor
```

- Specify service connection properties for the data store by its name:
```
customers.baseUrl = http://localhost:8081  
customers.clientId = orders  
customers.clientSecret = orders123
```

If you want to authenticate real users in the service app as demonstrated in https://github.com/jmix-framework/jmix-separate-tiers-sample, set up the Password Grant in the service and add the following properties to the client:

```
backend.authenticator = restds_RestPasswordAuthenticator
jmix.restds.authentication-provider-store = backend
```

Here `backend` is the name of the REST data store representing the service app.

## Data Model

In the client project, create DTO entities equivalent to service entities. Entity classes can have different names, but attributes should match exactly (however DTOs can have fewer attributes). Annotate the DTOs with `@Store` to specify the additional data store. For example:
```java
@Store(name = "customers")  
@JmixEntity  
public class CustomerRegion {  
    @JmixGeneratedValue  
    @JmixId    
    private UUID id;  
  
    private Integer version;  
  
    @InstanceName  
    @NotNull    
    private String name;

	// getters and setters
}
```

If the client entity name differs from the service one, use the `@RestDataStoreEntity` annotation, for example:
```java
@Store(name = "customers")  
@RestDataStoreEntity(remoteName = "CustomerRegion")  
@JmixEntity  
public class CustomerRegionDto {
```

For embedded attributes on the client side use the `@JmixEmbedded` annotation instead of JPA's `@Embedded`, for example:
```java
@Store(name = "customers")
@JmixEntity  
public class Customer {
	// ...
	
	@JmixEmbedded  
	private CustomerAddress address;
```

For one-to-many composition attributes on the client side define the inverse attribute:
```java
@Store(name = "customers")  
@JmixEntity  
public class Customer {
	// ...

	@Composition(inverse = "customer")  
	private Set<CustomerContact> contacts;
```

## Fetch Plans

When you load an external entity in the client application, you can specify a fetch plan to load references. The Generic REST currently supports only named fetch plans defined in fetch plans repository.  So the REST DataStore will request data from the service providing a fetch plan name. 

Therefore, both service and client applications must define all fetch plans in the their fetch plan repositories, with corresponding names. Inline fetch plans in XML and programmatically built fetch plans are not supported.

## Entity Events

REST DataStore sends `EntitySavingEvent` and `EntityLoadingEvent` the same as JpaDataStore. But it doesn't send `EntityChangedEvent` because it cannot provide information about attributes changed since load. Instead of `EntityChangedEvent`, REST DataStore sends two specific events:
- `RestEntitySavedEvent` - sent after the entity is successfully saved to the service. It contains the saved entity instance with the state right before sending to the service.
- `RestEntityRemovedEvent` - sent after the entity is removed from the service. It contains removed entity with the state right before sending to the service.

## Loaded State

If a fetch plan doesn't include an attribute, this attribute is not loaded. Unlike for JPA entities, not loaded attributes of a REST entities just have `null` value and don't throw any exception on accessing them.

When updating an entity, the REST DataStore saves only loaded attributes. If an attribute was not loaded from the service, but changed from null to some value afterwards, it is considered loaded and hence the new value is saved.

The `EntityStates.isLoaded(entity, property)` method correctly returns the information whether a particular attribute of a REST entity is loaded. For this, the `EntityEntry` interface and its `BaseEntityEntry` implementation were changed:
- Added `getLoadedProperties()` method returning the names of loaded properties
- Added internal collection of `EntityPropertyChangeEvent` listeners not wrapped in weak references. Such listeners do not require links to them from outside the instance to not be removed by GC. 
  The REST DataStore adds a listener to this collection for each loaded instance. The listener adds names of attributes changed after the entity was loaded to the `loadedProperties` set.
  *Unrelated idea: custom non-weak listeners can be used to calculate attribute values based on other attributes.*

## Security

REST data store applies **entity operations policy** defined by resource roles and **predicate policy** defined by row-level roles.

Authentication in REST data store can be done using Client Credentials Grant or Password Grant provided by the Authorization Server add-on. The latter requires setting the following properties:

```
# Sets specific authenticator for RestInvoker
<ds-name>.authenticator = restds_RestPasswordAuthenticator

# The specified data store will be used for authentication when RestAuthenticationToken is provided to AuthenticationManager
jmix.restds.authentication-provider-store = <ds-name>
``` 

## Filtering Loaded Data

The following filtering options are supported when loading external entities through `DataManager`:
- By condition:
```java
dataManager.load(Customer.class)  
        .condition(PropertyCondition.equal("firstName", "John")) 
        .list();
```
- By query, which is a JSON expression supported by Generic REST in `/search` endpoint:
```java
String query = """
		{
		  "property": "firstName",
		  "operator": "=",
		  "value": "John"
		}
		""";
List<Customer> customers = dataManager.load(Customer.class)
		.query(query)
		.list();
```
- By identifiers:
```java
dataManager.load(Customer.class).ids(id1, id2).list();
```

The JSON query can be specified in XML descriptors for data containers and itemsQuery elements:
```xml
<entityComboBox id="customerField" property="customer">  
    <itemsQuery class="com.company.orders.entity.customer.Customer" 
                searchStringFormat="${inputString}">  
        <fetchPlan extends="_base"/>  
        <query>  
            <![CDATA[  
            {  
              "property": "lastName",  
              "operator": "contains",  
              "parameterName": "searchString"  
            }  
            ]]>  
        </query>  
    </itemsQuery>  
</entityComboBox>
```
To specify a parameter instead of a literal value in JSON conditions, use `parameterName` key instead of `value` as shown above. The REST DataStore will substitute this property with `"value": <parameter-value>`  in the resulting request.

## Limitations

- No lazy loading
- No EntityChangeEvent with AttributeChanges
- `DataManager.loadValues()` and `loadValue()` methods throw UOE

## Breaking changes

- `CrudEntityConstraint` and `InMemoryCrudEntityConstraint` moved from `io.jmix.securitydata.constraint` to `io.jmix.security.constraint`


jmix-core/core/src/main/java/io/jmix/core/datastore/security/BaseEntityAttributesEraser.java
```diff
@@ -112,7 +110,7 @@ protected void restoreCollectionAttribute(Object entity, MetaProperty metaProper
         }
         for (Object id : ids) {
             MetaClass metaClass = metaProperty.getRange().asClass();
-            Object reference = dataManager.getReference(metaClass.getJavaClass(), id);
+            Object reference = getEntityReference(metaClass, id); //dataManager.getReference(metaClass.getJavaClass(), id);
```
dtaimanov, line 113: Looks like an obsolete comment left

dtaimanov (approved):

jmix-core/core/src/main/java/io/jmix/core/datastore/security/BaseEntityAttributesEraser.java
knstvk: Fixed

Issue #3437, opened 2024-07-02
Support trim attribute for TextField


glebfox:
Source: [forum](https://forum.jmix.io/t/textfield-not-trimming-white-space-in-jmix-2/5132)
Similar to Classic UI - https://docs.jmix.io/1.x/jmix/1.5/ui/vcl/components/text-field.html#trim

KremnevDmitry:
`TextArea` component is also supported

glebfox:
@KremnevDmitry create doc issue

KremnevDmitry:
Jmix Studio issues: 
* [JST-5277](https://youtrack.jmix.io/issue/JST-5277/Support-application-property-aware-meta-property-default-value)
* [JST-5278](https://youtrack.jmix.io/issue/JST-5278/Migrator-to-24-should-add-trim-enabled-application-property)

Doc issue: https://github.com/jmix-framework/jmix-docs/issues/861

SergeiAksenov2:
Tested on:
1) Jmix version: 2.4.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6816-242
IntelliJ version: IntelliJ IDEA 2024.2.1 (Community Edition)
2) Jmix version: 2.4.999-SNAPSHOT
Jmix Studio plugin version: 2.3.SNAPSHOT6809-241
IntelliJ version: IntelliJ IDEA 2024.1.6 (Community Edition)

Trim is supported - Ok.

PR #3651, opened 2024-08-27
Support trim attribute for TextField jmix-framework/jmix#3437

KremnevDmitry:
See: #3437 

The studio issue will be created after the merging


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/textfield/TypedTextField.java
```diff
@@ -176,6 +182,10 @@ public void setValue(String value) {
 
     protected void setValueInternal(@Nullable V modelValue, String presentationValue, boolean fromClient) {
         try {
+            if (isTrimEnabled()) {
```
glebfox, line 185: Move to `convertToModel`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiComponentProperties.java
```diff
@@ -112,6 +114,11 @@ public class UiComponentProperties {
      */
     boolean showErrorMessageBelowField;
 
+    /**
+     * Whether to trim the entered string by default for {@link HasTrimming} components.
+     */
+    boolean defaultTrimEnabled;
```
glebfox, line 120: we discussed that this feature should be enabled by default and disabled during migration process

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiComponentProperties.java
KremnevDmitry: Sure, the default value is `true` by default (see the `UiComponentProperties` class constructor)
The issue for the studio will be created after the merging (as stated in the PR description)

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
```diff
@@ -2116,6 +2116,7 @@ public interface StudioComponents {
                     @StudioProperty(xmlAttribute = "required", category = StudioProperty.Category.VALIDATION, type = StudioPropertyType.BOOLEAN,
                             defaultValue = "false"),
                     @StudioProperty(xmlAttribute = "tabIndex", type = StudioPropertyType.INTEGER),
+                    @StudioProperty(xmlAttribute = "trimEnabled", type = StudioPropertyType.BOOLEAN),
```
glebfox, line 2119: what about default value?

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiComponentProperties.java
glebfox: Yeah, missed that :)

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/textfield/TypedTextField.java
KremnevDmitry: If I move it to `convertToModel`, this is what happens:
1. The value set from the server-side will not be trimmed:
   If you try:
   ```java
   textField.setTypedValue("   test    ");
   ```
   Then `modelValue` will not be `null`, so the conversion will not be processed.
   image
2. `typedValue` will not be equal `value`:
   image
   image

jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/component/StudioComponents.java
KremnevDmitry: This will be implemented later in the studio issue. Since the value depends on the application property, it is necessary to implement a new strategy at the meta-annotation level.

glebfox (approved):

Issue #3343, opened 2024-06-05
Builders in Dialogs bean should return Dialog instance when it is opened


Flaurite:
### Description
It would be great to have method that return opened dialog instance from `OptionDialogBuilder`, `MessageDialogBuilder`, etc.
The return dialog can be further configured and can be closed programmatically.

### Possible solutions
1) Provide method `build()` that will return `Dialog`.
2) Make `open()` method return `Dialog`.

PR #3663, opened 2024-08-30
Builders in Dialogs bean should return Dialog instance when it is opened jmix-framework/jmix#3343

KremnevDmitry:
See: #3343 

`build()` and `open()` methods have been added to all dialog builders.
Why do you need a build method?

After the `build()` method, you can customize the `dialog`: add listeners for closing/opening. And then manually open the `dialog` using the `Dialog#open()` method. If we'll provide only the `Builder#open()` method, it will be impossible to add a listener for opening.

----

Features of the implementation of the `BackgroundTaskDialogBuilder#build` method.

Here is the old implementation of the `BackgroundTaskDialogBuilder#open` method:
```java
        @Override
        public void open() {
            if (isIndeterminateMode()) {
                progressTextSpan.setVisible(false);
                progressBar.setIndeterminate(true);
            } else {
                progressTextSpan.setVisible(true);
            }
            updateProgress(0);

            dialog.open();

            startExecution();
    }
```

As you can see - the background task starts immediately using the `startExecution()` method after opening the dialog. The build method cannot be implemented in a this way.

In order for users to be able to customize the dialog and listeners after the `BackgroundTaskDialogBuilder#build`, a self-removing listener was created that runs the background task once after opening the dialog.

This way the dialog can be opened outside the builder, the background task will be starter in any case.

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
@@ -58,7 +58,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -793,7 +792,7 @@ public class DialogsImpl implements Dialogs {
         protected boolean showProgressInPercentage;
         protected boolean cancelAllowed = false;
 
-        protected AtomicReference<Registration> openedRegistration = new AtomicReference<>();
+        protected Registration openedRegistration;
 
         protected BackgroundTaskHandler<V> taskHandler;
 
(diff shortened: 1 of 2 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
```diff
@@ -613,8 +612,25 @@ public String getMaxHeight() {
         }
 
         @Override
-        public void open() {
+        public Dialog build() {
+            DialogAction okAction = new DialogAction(DialogAction.Type.OK);
```
glebfox, line 616: Message Dialog has only OK button which doesn't depend on any setting of the builder.  With current implementation, sequential calls of the `build()` method produce button duplicates:

```
@Subscribe(id = "testBtn", subject = "clickListener")
public void onTestBtnClick(final ClickEvent<JmixButton> event) {
    Dialogs.MessageDialogBuilder builder = dialogs.createMessageDialog()
            .withText("Test");
    Dialog dialog = builder.build();
    dialog = builder.open();
}
```

Screenshot 2024-09-02 at 11 49 37

For example, `OptionsDialogBuilder` removes all buttons in the `build()` and inits them again. But in case of Message Dialog, as mentioned above, the OK button is persistent.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
```diff
@@ -999,9 +1014,29 @@ public void open() {
             }
             updateProgress(0);
 
-            dialog.open();
+            AtomicReference<Registration> openedRegistration = new AtomicReference<>();
```
glebfox, line 1017: Same as for message dialog, it's possible to call `build()` several times. Let's convert this variable to class field, so it's re-set if build() is called again.

glebfox (changes requested):

glebfox (approved):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
```diff
@@ -999,9 +1018,33 @@ public void open() {
...
-            dialog.open();
+            openedRegistration.getAndUpdate(registration -> {
+                // removing the previous listener, makes sense in case of dialog rebuilding
+                if (registration != null) {
+                    registration.remove();
+                }
+
+                return dialog.addOpenedChangeListener(event -> {
+                    // self-remove
+                    if (openedRegistration.get() != null) {
```
glebfox, line 1029: Try `event.unregisterListener()`

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/impl/DialogsImpl.java
KremnevDmitry: Works great

Issue #3263, opened 2024-05-21
Excel export action should take into account grid columns visibility


gorbunkov:
Starting with Jmix 2.2 the `gridColumnVisibility` may be added to the DataGrid to control columns visibility. The `ExcelExportAction` exports all columns both visible and hidden. 

We should either make the action export only visible columns or support additional configuration for the action (export all columns, export visible columns).

It will also be useful if we provide a configuration property for the export action that will configure exact list of columns that must be exported. For example, it may be a comma-separated list of column names.

See [forum](https://forum.jmix.ru/t/eksport-dannyh-tabliczy-v-excel/1803)

KremnevDmitry:
Related studio migrator issue: [JST-5316](https://youtrack.jmix.io/issue/JST-5316/Migrator-to-24-should-add-default-columns-to-export-application-property)
Related docs issue: https://github.com/jmix-framework/jmix-docs/issues/869

# For QA:
Use temporary documentation for testing: https://github.com/jmix-framework/jmix/pull/3668

KremnevDmitry:
> columnsToExport has no effect on behaviour
> 
> ```
>     <action id="excelExportAction" type="grdexp_excelExport">
>                     <properties>
>                         <property name="columnKeysToExport" value="aaa, bbb"/>
>                         <property name="columnsToExport" value="VISIBLE_COLUMNS"/>
>                     </properties>
>                 </action>
> ```
> 
> ```
>  <columns resizable="true">
>                 <column property="aaa" visible="false"/>
>                 <column property="bbb"/>
>                 <column property="dddd"/>
>               </columns>
> ```
> 
[...]

PR #3668, opened 2024-09-02
Excel export action should take into account grid columns visibility jmix-framework/jmix#3263

KremnevDmitry:
See: #3263
A migration issue for the Studio will be created after the merging. The migration should add an application to preserve the old export behavior.
```properties
jmix.gridexport.default-column-export-filter=ALL_COLUMNS
```
----

Added three options for export column filtering:

# Predefined column filter predicate (enumeration)

```java
    /**
     * Export all columns.
     */
    ALL_COLUMNS(column -> true),

    /**
     * Export only visible columns.
     */
    VISIBLE_COLUMNS(Grid.Column::isVisible);
``` 

You can set filtering options using `jmix.gridexport.default-column-export-filter` application property or using a `properteis` in the Component Inspector.
The default value for new projects: `VISIBLE_COLUMNS`
Migration will add the old value to the `application.properties`.

This filtering option **has the lowest priority**.

# Flexible filtering predicate (install point)

```java
    @Install(to = "usersDataGrid.excelExport", subject = "columnExportPredicate")
    private boolean usersDataGridExcelExportColumnExportPredicate(final Grid.Column<Object> column) {
        return column.getKey().contains("name") || column.getKey().contains("zone");
    }
```

Provides the same functionality as enumeration above.

This filtering options **has the secondary priority.**

# Specefying keys of columns for filtering (list of keys)

Using the Component Inspector, you can specify columns to export.

```xml
                <action id="excelExport" type="grdexp_excelExport">
                    <properties>
                        <property name="columnKeysToExport" value="username, firstName, lastName"/>
                    </properties>
                </action>
```

This filtering option **has the primary priority**.

### DIFF
--- a/jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/action/ExportAction.java
+++ b/jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/action/ExportAction.java
@@ -25,22 +25,27 @@ import io.jmix.flowui.action.DialogAction;
 import io.jmix.flowui.action.SecuredBaseAction;
 import io.jmix.flowui.action.list.ListDataComponentAction;
 import io.jmix.flowui.component.ListDataComponent;
+import io.jmix.flowui.component.grid.DataGrid;
 import io.jmix.flowui.data.ContainerDataUnit;
 import io.jmix.flowui.download.Downloader;
 import io.jmix.flowui.kit.action.Action;
 import io.jmix.flowui.kit.action.ActionVariant;
 import io.jmix.flowui.model.HasLoader;
 import io.jmix.gridexportflowui.GridExportProperties;
+import io.jmix.gridexportflowui.exporter.ColumnExportFilter;
 import io.jmix.gridexportflowui.exporter.DataGridExporter;
 import io.jmix.gridexportflowui.exporter.ExportMode;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.lang.Nullable;
+import org.springframework.util.CollectionUtils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  * Base action for export table content with defined exporter.
@@ -61,6 +66,10 @@ public class ExportAction extends ListDataComponentAction<ExportAction, Object> 
 
     protected DataGridExporter dataGridExporter;
     protected List<ExportMode> availableExportModes;
+
+    protected ColumnExportFilter columnExportFilter;
+    protected Predicate<Grid.Column<Object>> columnExportPredicate;
+    protected List<String> columnKeysToExport;
 
     public ExportAction() {
         this(ID);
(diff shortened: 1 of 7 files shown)


jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/action/ExportAction.java
```diff
@@ -95,6 +104,65 @@ public void setGridExportProperties(GridExportProperties gridExportProperties) {
...
+
+    /**
+     * Sets the {@link ColumnExportFilter} that is used to filter columns to export. This is a simple, predefined
+     * alternative to {@link #setColumnExportPredicate(Predicate)}. Has the lowest filtering priority.
+     * <p>
+     * The default value depends on {@link GridExportProperties#getDefaultColumnExportFilter()}.
+     *
+     * @param columnExportFilter column export filter to set
+     */
+    public void setColumnExportFilter(ColumnExportFilter columnExportFilter) {
```
glebfox, line 118: From users point of view, they set column export mode either **all columns** or **visible**. The fact that this enum stores filter predicates is hidden from users. Just read this: _set column export filter all rows_. It sounds like we need to filter all columns, but in fact, it's opposite. Let's rename Enum, app property and method.

jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/action/ExportAction.java
```diff
@@ -95,6 +104,65 @@ public void setGridExportProperties(GridExportProperties gridExportProperties) {
...
+        return this;
+    }
+
+    /**
+     * Sets the {@link Predicate} that is used to filter columns to export. This is a flexible alternative to
+     * {@link #setColumnExportFilter(ColumnExportFilter)}. Has secondary filtering priority.
+     *
+     * @param columnExportPredicate column export predicate to set
+     */
+    public void setColumnExportPredicate(Predicate<Grid.Column<Object>> columnExportPredicate) {
```
glebfox, line 137: Let's name it `setColumnExportFilter`. Currently, without reading JavaDoc, it's unclear what we exactly set.

glebfox (changes requested):

jmix-gridexport/gridexport-flowui/src/main/java/io/jmix/gridexportflowui/action/ExportAction.java
```diff
@@ -104,46 +104,46 @@ public void setGridExportProperties(GridExportProperties gridExportProperties) {
...
-     * Sets the {@link ColumnExportFilter} that is used to filter columns to export. This is a simple, predefined
-     * alternative to {@link #setColumnExportPredicate(Predicate)}. Has the lowest filtering priority.
+     * Sets the {@link ColumnsToExport} that is used to filter columns to export. This is a simple, predefined
+     * alternative to {@link #setColumnsExportFilter(Predicate)}. Has the lowest filtering priority.
      * <p>
-     * The default value depends on {@link GridExportProperties#getDefaultColumnExportFilter()}.
+     * The default value depends on {@link GridExportProperties#getDefaultColumnsToExport()}.
      *
-     * @param columnExportFilter column export filter to set
+     * @param columnsToExport column export filter to set
```
glebfox, line 116: Fix java doc)

glebfox (approved):

Issue #3585, opened 2024-08-14
Add application property to disable initial form validation


glebfox:
Source: [forum](https://forum.jmix.io/t/how-to-disable-initial-form-validation/5352)

Currently, when a field is bound to an entity attribute and this attribute is mandatory, validation is triggered from `setRequiredIndicatorVisible`, e.g.:

```
@Override
public void setRequiredIndicatorVisible(boolean requiredIndicatorVisible) {
    super.setRequiredIndicatorVisible(requiredIndicatorVisible);

    fieldDelegate.updateInvalidState();
}
```

As a result a view is opened with error messages which may scare users :-)

Add application property to disable initial form validation, so even required validation is triggered by form submitting.

rusiaikinat:
Kotlin, java
Jmix version: 2.4.999-SNAPSHOT
Jmix Studio plugin version: 2.4.SNAPSHOT6821-242
IntelliJ version: IntelliJ IDEA 2024.2.1 (Ultimate Edition)

PR #3670, opened 2024-09-02
Add application property to disable initial form validation jmix-framework/jmix#3585

KremnevDmitry:
See: #3585


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/delegate/AbstractFieldDelegate.java
```diff
@@ -252,6 +254,12 @@ public void setConversionInvalid(boolean conversionInvalid) {
     }
 
     public void updateInvalidState() {
+        // skip initial validation when the form is opened
```
glebfox, line 257: It's not actually true. Validation is performed when the required state is changed. In particular, due to data binding. What if the user disabled the required indicator and then enabled it again? Let's handle required state changes in general.

glebfox (changes requested):

glebfox (approved):

Issue #3620, opened 2024-08-19
Date datatype can't be formatted for Axes


KremnevDmitry:
### Environment

Jmix version: 2.3.999-SNAPSHOT
Forum topic: [link](https://forum.jmix.io/t/how-to-format-date-in-dd-mm-yy-format-in-chart-axis/5370)

### Bug Description

An attribute of `java.util.Date` can't be formatted by the axis formatter

ilushaaa:
Make sure that all Jmix data types related to date and time work after the changes.

SergeiAksenov2:
Tested on:
Jmix version: 2.3.999-SNAPSHOT
Jmix Studio plugin version: 2.4.SNAPSHOT6844-242
IntelliJ version: IntelliJ IDEA 2024.2.3 (Community Edition)

PR #3729, opened 2024-09-28
Date datatype can't be formatted for Axes 3620

ilushaaa:
#3620


jmix-charts/charts-flowui/src/main/java/io/jmix/chartsflowui/component/serialization/AbstractDataSerializer.java
```diff
@@ -39,7 +39,7 @@ public abstract class AbstractDataSerializer<T> extends AbstractSerializer<T> {
     protected static final String ITEM_KEY_PROPERTY_NAME = "$k";
 
     public static final String DEFAULT_DATE_FORMAT = "yyyy-MM-dd";
-    public static final String DEFAULT_TIME_FORMAT = "HH:mm:ss:S";
+    public static final String DEFAULT_TIME_FORMAT = "HH:mm:ss.S";
```
glebfox, line 42: Are you sure that `.S` is enough? usually we use `HH:mm:ss.SSS`

glebfox (changes requested):

glebfox (approved):

Issue #2404, opened 2023-10-26
Get rid of AcceptsTenant interface


gorbunkov:
The `io.jmix.multitenancy.core.AcceptsTenant` is used only in `io.jmix.multitenancy.core.impl.TenantProviderImpl#getCurrentUserTenantId` for getting a current user tenant and checking that a user class supports tenant. It seems that we can analyze the presence of `@TenantId` annotation on any class field instead of checking this interface.  In this case [configuring a user entity](https://docs.jmix.io/jmix/multitenancy/index.html#configuring-users) will be simplified.

KremnevDmitry:
# For QA

## Test case 1

1. Create a project with version 2.3
2. Set up multitenancy add-on according to [doc](https://docs.jmix.io/jmix/multitenancy/index.html#installation)
3. Check the functionality of the multitenancy add-on
4. Migrate the project version to 2.4
5. Check the functionality of the multitenancy add-on one more time

## Test case 2

1. Create a project with version 2.4
2. Set up multitenancy add-on in a new way (without using the `AcceptTenant` interface)
3. Check the functionality of the multitenancy add-on

SergeiAksenov2:
Tested on:
Jmix version: 2.4.999-SNAPSHOT
Jmix Studio plugin version: 2.4.SNAPSHOT6866-242
IntelliJ version: IntelliJ IDEA 2024.2.3 (Community Edition)

PR #3751, opened 2024-10-03
Get rid of AcceptsTenant interface jmix-framework/jmix#2404

KremnevDmitry:
See: #2404

### DIFF
--- a/jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantProviderImpl.java
+++ b/jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantProviderImpl.java
@@ -16,13 +16,15 @@
 
 package io.jmix.multitenancy.core.impl;
 
+import io.jmix.core.annotation.TenantId;
+import io.jmix.core.common.util.ReflectionHelper;
 import io.jmix.core.security.CurrentAuthentication;
 import io.jmix.multitenancy.core.TenantProvider;
-import io.jmix.multitenancy.core.AcceptsTenant;
+import org.springframework.lang.Nullable;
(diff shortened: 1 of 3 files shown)


jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantProviderImpl.java
```diff
@@ -46,14 +48,33 @@ public String getCurrentUserTenantId() {
...
+
         if (!(currentAuthentication.getAuthentication().getPrincipal() instanceof UserDetails)) {
             return TenantProvider.NO_TENANT;
         }
+
         UserDetails userDetails = currentAuthentication.getUser();
-        if (!(userDetails instanceof AcceptsTenant)) {
+        String tenantIdFieldName = getTenantIdFieldName(userDetails.getClass());
+
+        if (tenantIdFieldName == null) {
```
glebfox, line 59: Check `AcceptsTenant` presence, just in case

glebfox (changes requested):

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/impl/TenantProviderImpl.java
KremnevDmitry: As far as I understand from the issue description, we are trying to completely get rid of the use of the `AcceptTenants` interface.

That's why it's marked as `deprecated`.

During migration nothing will change for end-users, since now only the presence of the `@TenantId` annotation above the field is required. The interface check in this case, as it seems to me, will be redundant.

Issue #3430, opened 2024-06-28
Support for DTO entities in Generic REST


knstvk:
The REST add-on should work with DTO entities in the same way as with JPA.
The `/search` endpoint filter should be converted to the `Condition` tree.

Issue #1866, opened 2023-06-20
Restore sample-rest project with REST API tests


gorbunkov:
The project was disabled when migrated to Spring Boot 3.

#### Changes
`sample-rest` project partially restored.
Second part: https://github.com/jmix-framework/jmix/issues/3758

Issue #3698, opened 2024-09-14
Support for entity parameters in search conditions


knstvk:
The REST API `/search` endpoint should support entity parameters like this:

```
{
  "conditions": [
    {
      "property": "field1",
      "operator": "=",
      "value": {
        "_entityName": "Customer",
        "id": "00000000-0000-0000-0000-000000000001",
        "firstName": "John",
        "lastName": "Doe"
      }
    }
  ]
}
```

This is required for #3641, see for example `filter_builder.RestFilterBuilderTest#testQueryWithEntityParameter`.

PR #3757, opened 2024-10-05
Support for DTO entities in Generic REST, Restore sample-rest project

dtaimanov:
jmix-framework/jmix#1866,  jmix-framework/jmix#3430


jmix-rest/rest/src/main/java/io/jmix/rest/impl/service/EntitiesControllerManager.java
```diff
@@ -193,35 +201,33 @@ public EntitiesSearchResult searchEntities(String entityName,
...
-        }
+        LoadContext<?> ctx = new LoadContext<>(metadata.getClass(metaClass.getJavaClass()))
+                .setQuery(new LoadContext.Query(queryString)
+                        .setCondition(jmixCondition));
+        return dataManager.getCount(ctx);
+    }
 
-        return new EntitiesSearchResult(json, count);
+    protected static String createSimpleSelect(MetaClass metaClass){
+        return "select e from " + metaClass.getName() + " e";
```
knstvk, line 230: Return JPQL query only for JPA entities (`MetadataTools.isJpaEntity()`). Otherwise return empty string. See for example `io.jmix.restds.impl.RestDataStore.countAll` - it ignores the query text.

jmix-rest/rest/src/main/java/io/jmix/rest/impl/service/EntitiesControllerManager.java
```diff
@@ -287,20 +282,24 @@ public Long countSearchEntities(String entityName, String searchRequestBody) {
         return countSearchEntities(entityName, searchEntitiesRequest.getFilter().toString(), searchEntitiesRequest.getModelVersion());
     }
 
-    protected String _loadEntitiesList(String queryString,
+    protected String _loadEntitiesList(Condition condition,
```
knstvk, line 285: Do not use "_" in names. If you cannot invent a new name, use "Internal" suffix, like `loadEntitiesListInternal`.
But here `loadEntitiesJson()` would be a good name.

jmix-rest/rest/src/main/java/io/jmix/rest/impl/service/filter/RestFilterParser.java
```diff
@@ -227,21 +218,45 @@ protected RestFilterPropertyCondition parsePropertyCondition(JsonObject conditio
             } else {
                 value = parseValue(metaProperty, valueJsonElem.getAsString());
             }
-            condition.setValue(transformValue(value, op));
-            condition.setQueryParamName(generateQueryParamName());
+            condition.setParameterValue(value);
+        } else {
+            if ("isNull".equals(operator)) condition.setParameterValue(false);
```
knstvk, line 223: Avoid literals. Perhaps create a class with string constants for all supported REST operators. Or an enum with a second field - corresponding Condition operation.

jmix-rest/sample-rest/src/test/resources/application.properties
```diff
@@ -13,6 +13,40 @@
...
+spring.security.oauth2.authorizationserver.client.myclient.registration.authorization-grant-types=password
+# Client credentials must be passed in the Authorization header using the HTTP Basic authentication scheme
+spring.security.oauth2.authorizationserver.client.myclient.registration.client-authentication_methods=client_secret_basic
+# Use opaque tokens instead of JWT
+spring.security.oauth2.authorizationserver.client.myclient.token.access-token-format=reference
+
+# Define resource roles for the client
+#todo naming
+jmix.authserver.client.myclient.client-id = myclient
+jmix.authserver.client.myclient.resource-roles = rest-minimal, system-full-access
```
knstvk, line 47: Remove rest-minimal

jmix-rest/sample-rest/src/test/resources/application.properties
```diff
@@ -32,4 +66,6 @@ jmix.data.dbms-type-db1=hsql
 
 jmix.core.available-locales=en,ru
 
-main.liquibase.change-log=io/jmix/samples/rest/liquibase/test-changelog.xml
\ No newline at end of file
+#todo which one is correct? revert if removing unnecessary hsql file db will help
```
knstvk, line 69: What does it mean?

knstvk (changes requested):

jmix-rest/rest/src/main/java/io/jmix/rest/impl/service/EntitiesControllerManager.java
dtaimanov: Fixed, but null is used instead of empty string because it is more clear way to say that query is absent and `io.jmix.core.LoadContext.Query#getQueryString()` allow nulls now.

jmix-rest/sample-rest/src/test/resources/application.properties
dtaimanov: It looks like v1.6 did not use this changelog. At least `io/jmix/samples/rest/liquibase/test-changelog.xml` has been used instead which has no include for `io/jmix/samples/rest/changelog.xml`. 

Replaced unclear message with more clear one, reference to jmix-framework/jmix#3758 has been added to deal with it during work on this issue.

dtaimanov:
Support for entity parameters in search conditions jmix-framework/jmix#3698 changes have been added

knstvk (approved):
Please squash and merge it

Issue #3638, opened 2024-08-21
Default value of dynamic attribute is cleared when detail view opens


fractal3000:
### Environment

Jmix version: 2.3.2

### Bug Description

A description of what the bug is.

### Steps To Reproduce
Create dynamic attribute of String type
Set any default value for it
![image](https://github.com/user-attachments/assets/4794b9ab-d699-42b2-8f3b-8d18964e7223)
Save the attribute
![image](https://github.com/user-attachments/assets/0aa6ea4e-6ad5-4a03-9841-9de11962ac45)
Open editor screen of the attribute again

### Current Behavior

![image](https://github.com/user-attachments/assets/ed05b0be-6714-475b-bf99-89a83651d727)

### Expected Behavior

The being set up default value of the attribute have been saved.

SergeiAksenov2:
Tested on:
Jmix version: 2.4.999-SNAPSHOT
Jmix Studio plugin version: 2.4.SNAPSHOT6866-242
IntelliJ version: IntelliJ IDEA 2024.2.3 (Community Edition)

PR #3760, opened 2024-10-07
fix default value of dynamic attribute #3638

AmirShelli:
See issue #3638

**Problem:** In the `refreshAttributesUI()` method, both `defaultStringField` and `defaultEnumField` are cleared due to the following code:

```java
if (!visible && component instanceof HasValue) {
    ((HasValue<?, ?>) component).clear();
}
```

When the `defaultEnumField` component is processed, it gets cleared, which triggers a listener attached to it. This listener also clears the `defaultStringField`.

The listener is added in the `initDefaultEnumField()` method, which is called during every initialization:

```java
private void initDefaultEnumField() {
    //...
    defaultEnumField.addValueChangeListener(e -> getEditedEntity().setDefaultString(e.getValue()));
    if (StringUtils.isNotBlank(getEditedEntity().getDefaultString())) {
        defaultEnumField.setValue(getEditedEntity().getDefaultString());
    }
    //...
}
```

**Solution:** As a temporary fix until we refactor the dynamic attribute handling, I propose the following changes:

1. Call `initDefaultEnumField()` only when the component's data type is set to `ENUMERATION`. This prevents the listener from being attached when the data type is a string.

2. Modify the logic in `refreshAttributesUI()` to skip clearing the `defaultStringField` when dealing with an enumeration field. This change will prevent the `defaultEnumField` from being cleared.

### DIFF
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
@@ -267,9 +267,7 @@ public class CategoryAttributesDetailView extends StandardDetailView<CategoryAtt
     private boolean isRefreshing = false;
 
     @ViewComponent
-    protected TypedTextField<String> defaultStringField; //debug
-    @ViewComponent
-    protected TypedTextField<Integer> defaultIntField; //debug
+    protected TypedTextField<String> defaultStringField;
 
     private final List<String> defaultEnumValues = new ArrayList<>();
 
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
@@ -221,8 +221,6 @@ public class CategoryAttributesDetailView extends StandardDetailView<CategoryAtt
     @ViewComponent
     protected JmixComboBox<String> defaultEnumField;
     @ViewComponent
-    protected TypedTextField<String> defaultStringField;
-    @ViewComponent
     protected JmixComboBox<OptionsLoaderType> optionsLoaderTypeField;
     @ViewComponent
     protected JmixValuePicker<Object> defaultEntityIdField;
@@ -242,6 +240,8 @@ public class CategoryAttributesDetailView extends StandardDetailView<CategoryAtt
     protected JmixTabSheet tabSheet;
     @ViewComponent
     protected TypedTextField<String> codeField;
+    @ViewComponent
+    protected TypedTextField<String> defaultStringField;
     @ViewComponent
     protected TypedTextField<BigDecimal> defaultDecimalField;
     @ViewComponent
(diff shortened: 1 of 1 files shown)


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -278,7 +281,10 @@ protected void onInit(InitEvent event) {
     private void initDefaultEnumField() {
         defaultEnumValues.addAll(getEnumValues());
         defaultEnumField.setItems(defaultEnumValues);
-        defaultEnumField.addValueChangeListener(e -> getEditedEntity().setDefaultString(e.getValue()));
+        defaultEnumField.addValueChangeListener(e -> {
```
KartnDev, line 284: why?

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -302,7 +308,10 @@ protected List<String> getEnumValues() {
 
     @Subscribe
     protected void onAfterShow(BeforeShowEvent event) {
-        initDefaultEnumField();
+        if (getEditedEntity().getDataType() != null &&
```
KartnDev, line 311: as we discussed, yes this is right code, but we need to relocate this part into the method that will be named "initDefaultEnumFieldIfEnum"

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -684,6 +693,10 @@ protected void refreshAttributesUI() {
             component.setVisible(visible);
 
             if (!visible && component instanceof HasValue) {
+                if (component.equals(defaultStringField) && ENUMERATION.equals(attributeType) &&
```
KartnDev, line 696: If we continueing if condition from new line, lets split each condition from new line like this:
(cond1 &&
cond2 &&
cond3)

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
```diff
@@ -266,6 +266,9 @@ public class CategoryAttributesDetailView extends StandardDetailView<CategoryAtt
     private JmixButton editEnumerationBtn;
     private boolean isRefreshing = false;
 
+    @ViewComponent
```
KartnDev, line 269: Please sort fields - this field should be with other ViewComponents fields - TypedTextField nearby other text fields

jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/view/categoryattr/CategoryAttributesDetailView.java
KartnDev: don't bring outstaff code that not related with ticket

KartnDev: also if lambda can be written into one line without brackets - don't use brackets

KartnDev (changes requested):
Also, you forget name branch **feature/** 3638-default-value-of-dynamic-attribute

And about commit messages:

- instead of writing **#** 123123 - {comment text} use syntax like this: {commit message}  **#** 123123 ( several things fixed **#** 123, spaces between hashtags to avoid github trigger that links issues )
image

- when you will squash commits, use this syntax

- **DONT FORGET TO SQUASH!** :)

KartnDev (approved):

Issue #3788, opened 2024-10-17
Use gradle-idea-ext-plugin to declare excluded directories


Gasloff:
Consider using **gradle-idea-ext-plugin** in Jmix projects. 
Using this plugin we can declare excluded directories for IDEA.
See example below:

```
plugins {
    id 'io.jmix' version '2.3.4'
    id 'java'
    id 'org.jetbrains.gradle.plugin.idea-ext' version '1.1.9'
}

...

idea {
    module {
        excludeDirs.add(file("frontend/generated"))
        excludeDirs.add(file(".jmix"))
    }
}
```

See: https://github.com/JetBrains/gradle-idea-ext-plugin/wiki

---

+ node_modules
+ bundles

PR #3794, opened 2024-10-21
Use gradle-idea-ext-plugin to declare excluded directories jmix-framework/jmix#3788

KremnevDmitry:
See: #3788


KremnevDmitry (commented):
Do we need to create an issue for the Studio team to modify existing `build.gradle` files in projects during migration process?

knstvk:
> Do we need to create an issue for the Studio team to modify existing `build.gradle` files in projects during migration process?

No, it's only for new projects.

knstvk (approved):

Issue #3820, opened 2024-10-28
Implement `InitializingBean` in `PivotTableExporterImpl` to initialize `PivotTableExcelExporter`


ilushaaa:
### Environment

Jmix version: <2.4>

### Bug Description

`PivotTableExporterImpl ` gets `PivotTableExcelExporter` as a constructor parameter. This is incorrect, because `PivotTableExcelExporter` is a bean and should be initialized via `ApplicationContext`.

PR #3873, opened 2024-11-15
Implement InitializingBean in PivotTableExporterImpl to initialize PivotTableExcelExporter 3820

ilushaaa:
#3820

### DIFF
--- a/jmix-pivottable/pivottable-flowui/src/main/java/io/jmix/pivottableflowui/export/PivotTableExporterImpl.java
+++ b/jmix-pivottable/pivottable-flowui/src/main/java/io/jmix/pivottableflowui/export/PivotTableExporterImpl.java
@@ -56,6 +56,19 @@ public class PivotTableExporterImpl implements PivotTableExporter, ApplicationCo
         this.pivotTable = pivotTable;
     }
 
+    /**
+     * @deprecated use {@link #PivotTableExporterImpl(PivotTable)}
+     */
+    @Deprecated(since = "2.4.1", forRemoval = true)
+    public PivotTableExporterImpl(PivotTable<?> pivotTable, PivotTableExcelExporter exporter) {
+        Preconditions.checkNotNullArgument(pivotTable);
+        Preconditions.checkNotNullArgument(exporter);
+
+        this.pivotTable = pivotTable;
+        this.excelExporter = exporter;
+        this.excelExporter.init(pivotTable);
+    }
+
     @Override
     public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
         this.applicationContext = applicationContext;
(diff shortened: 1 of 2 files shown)


jmix-pivottable/pivottable-flowui/src/main/java/io/jmix/pivottableflowui/export/PivotTableExporterImpl.java
```diff
@@ -36,22 +40,30 @@
...
 
     public static final Set<Renderer> supportedRenderers = Collections.unmodifiableSet(Sets.newHashSet(
             Renderer.TABLE, Renderer.TABLE_BAR_CHART, Renderer.HEATMAP, Renderer.ROW_HEATMAP, Renderer.COL_HEATMAP));
 
+    protected ApplicationContext applicationContext;
     protected PivotTableExcelExporter excelExporter;
     protected PivotTable<?> pivotTable;
     protected String fileName;
 
-    public PivotTableExporterImpl(PivotTable<?> pivotTable, PivotTableExcelExporter exporter) {
```
KremnevDmitry, line 48: We need to keep this constructor and create a new one.

Since this API was not marked as `experimental`, and was not marked as `@Internal`, we do not have the right to change it in a bugfix release (even if no one extended it).

Unfortunately, we will have to keep it and mark it as deprecated, with a comment that we need to use a new constructor.

jmix-pivottable/pivottable-flowui/src/main/java/io/jmix/pivottableflowui/export/PivotTableExporterImpl.java
```diff
@@ -36,22 +40,30 @@
...
+    }
+
+    @Override
+    public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+        this.applicationContext = applicationContext;
+    }
+
+    @Override
+    public void afterPropertiesSet() {
+        this.excelExporter = applicationContext.getBean(PivotTableExcelExporter.class, pivotTable);
```
KremnevDmitry, line 66: According to the comment above, we should allow the exporter to be defined externally.
So here we should add a condition:
```java
if (this.excelExporter == null) {
    this.excelExporter = applicationContext.getBean(....);
}
```

KremnevDmitry: Also add a comment that this `condition` should be removed with **Release 3.0**.

KremnevDmitry (changes requested):

KremnevDmitry (approved):
Can be merged after fixes

Issue #3401, opened 2024-06-19
NPE on downloading PDF file in Chrome


gorbunkov:
### Environment

Jmix version: 2.2.3

### Bug Description

NullPointerException appears when users try to download the PDF file from the Chrome PDF viewer.

See [forum topic](https://forum.jmix.io/t/pdf-download-in-chrome/4973).

### Steps To Reproduce

Create a screen that downloads a PDF file using the `Downloader` bean, for example:

```java
    @Subscribe(id = "downloadPdfBtn", subject = "clickListener")
    public void onDownloadPdfBtnClick(final ClickEvent<JmixButton> event) {
        downloader.download(
                reportService.createPdf(),
                "report.pdf",
                DownloadFormat.PDF
        );
    }
```

Run the application, and initiate the downloading.

The PDF file will be opened in Chrome browser.

Try to download the file using the button in Chrome PDF viewer:

<img width="989" alt="Screenshot 2024-06-19 at 09 54 07" src="https://github.com/jmix-framework/jmix/assets/7888486/6c4ea299-a25c-49e5-8a2a-e85e1c6a3dd2">

### Current Behavior

An exception will be thrown and the file with 0 bytes size will be downloaded.

```
java.lang.NullPointerException: Cannot invoke "java.io.InputStream.read(byte[])" because "source" is null
	at com.vaadin.flow.server.StreamResource$Pipe.read(StreamResource.java:119) ~[flow-server-24.3.8.jar:24.3.8]
	at com.vaadin.flow.server.StreamResource$Pipe.copy(StreamResource.java:109) ~[flow-server-24.3.8.jar:24.3.8]
	at com.vaadin.flow.server.StreamResource$Pipe.accept(StreamResource.java:84) ~[flow-server-24.3.8.jar:24.3.8]
	at io.jmix.flowui.component.filedownloader.JmixFileDownloader.lambda$runCommand$18809115$1(JmixFileDownloader.java:121) ~[jmix-flowui-2.2.3.jar:na]
	at java.base/java.util.function.Consumer.lambda$andThen$0(Consumer.java:65) ~[na:na]
	at io.jmix.flowui.component.filedownloader.JmixFileDownloader.lambda$beforeClientResponseDownloadHandler$69016e48$1(JmixFileDownloader.java:158) ~[jmix-flowui-2.2.3.jar:na]
[...]

PR #3928, opened 2024-11-25
NPE on downloading PDF file in Chrome jmix-framework/jmix#3401

KremnevDmitry:
See: #3401

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
@@ -66,7 +66,7 @@ public class UiProperties {
      *
      * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives">Cache-Control HTTP | MDN</a>
      */
-    int fileDownloaderCacheMaxAge;
+    int fileDownloaderCacheMaxAgeSec;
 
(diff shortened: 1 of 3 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
```diff
@@ -59,6 +60,14 @@ public class UiProperties {
      */
     List<String> viewFileExtensions;
 
+    /**
+     * Sets the maximum time in seconds during which the file will be considered relevant.
+     * Makes sense for using the built-in PDF viewer in the Chrome browser.
+     *
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives">Cache-Control HTTP | MDN</a>
+     */
+    int fileDownloaderCacheMaxAge;
```
glebfox, line 69: If a variable stores a physical quantity, always mention the unit of change in the name (i.e. timeoutSec, timeoutMillis, distanceMiles)

See https://youtrack.haulmont.com/articles/PLAT-A-1433/Code-Best-Practices#Code-Structure

glebfox (changes requested):

glebfox (approved):

PR #3982, opened 2024-12-06
Refactor database-related dependencies

knstvk:
### Main changes

- `data` does not depend on EclipseLink library anymore.  Consequences:
	- Modules depending on `data` and having JPA entities (audit, appsettings, etc.)  now need explicit dependency on `eclipselink` module.
	- `ReferenceToEntity` entity was removed from `data` and added separately to `audit` and `dynattr`.
- `seurity-flowui` does not depend on `security-data` anymore.
- Introduced service interfaces:
	- `io.jmix.flowui.component.genericfilter.FilterConfigurationPersistence`
	- `io.jmix.security.role.assignment.RoleAssignmentPersistence`
	- `io.jmix.security.role.RolePersistence`
	- `io.jmix.security.usersubstitution.UserSubstitutionPersistence`
- `flowui-restds` module can be used in applications instead of `flowui-data` to store settings and filter configuration in a REST DataStore.
- Introduced `MetadataPostProcessor` interface which is invoked by `MetadataLoader` for final adjustments of metadata in add-ons and applications. In the framework, it's implemented in REST DataStore add-on to configure store for UI config entities (settings and filter configuration) based on application property.

### Usage examples

- [Frontend without DB (based on Jmix 2.5) by knstvk · Pull Request #1 · jmix-framework/jmix-separate-tiers-sample](https://github.com/jmix-framework/jmix-separate-tiers-sample/pull/1)

- [Sample Jmix Application Without Database](https://github.com/jmix-projects/sample-app-without-db)

### Breaking changes

- Removed `ReferenceToEntity` entity
- Moved `NotInstantiatedList`, `NotInstantiatedSet` to `io.jmix.eclipselink.lazyloading`
- Moved:
	- `io.jmix.flowuidata.accesscontext.UiGenericFilterModifyGlobalConfigurationContext` → `io.jmix.flowui.accesscontext`
	- `io.jmix.flowuidata.action.genericfilter.*` → `io.jmix.flowui.action.genericfilter`
	- `io.jmix.securityflowui.model.*` → `io.jmix.security.model.*`
- `UiDataGenericFilterSupport` was merged into `GenericFilterSupport` and removed


jmix-core/core/src/main/java/io/jmix/core/MetadataTools.java
```diff
@@ -1347,13 +1347,10 @@ private boolean isNotLoadedAttributeException(Throwable e) {
...
     /**
-     * Returns all additional properties which are providing classes implement MetadataExtension interface
-     *
-     * @param metaClass instance MetaClass for getting additional properties
-     * @return all additional properties
+     * Returns all additional properties of the given MetaClass provided by {@link MetadataExtension}s.
      */
     public Set<MetaProperty> getAdditionalProperties(MetaClass metaClass) {
-        if (metadataExtensions == null) {
+        if (metadataExtensions == null || metadataExtensions.isEmpty()) {
```
knstvk, line 1353: This field is not null (contains a proxy) because of lazy initialization.

jmix-core/core/src/main/java/io/jmix/core/impl/importexport/EntityImportExportImpl.java
```diff
@@ -752,6 +753,9 @@ protected Collection<Object> createNewCollection(MetaProperty metaProperty) {
     @Nullable
     protected Object findReferenceEntity(Object entity, EntityImportPlanProperty importPlanProperty, SaveContext saveContext,
                                          Set<Object> loadedEntities) {
+        if (!isStored(entity)) {
```
knstvk, line 756: Added because of `FilterConfigurationModel#rootCondition` field that causes an exception when deserializing in REST API.

jmix-security/security/src/main/java/io/jmix/security/model/RoleSourceEnum.java
```diff
@@ -14,21 +14,21 @@
...
-package io.jmix.securityflowui.model;
+package io.jmix.security.model;
 
 import io.jmix.core.metamodel.datatype.EnumClass;
 
 import org.springframework.lang.Nullable;
 
 
-public enum RoleSource implements EnumClass<String> {
+public enum RoleSourceEnum implements EnumClass<String> {
```
glebfox, line 24: Probably it's better to name it `RoleSourceType` because we usually don't use `Enum` in enum names.

glebfox (approved):

jmix-security/security/src/main/java/io/jmix/security/model/RoleSourceEnum.java
knstvk: Done

glebfox (approved):

Issue #3937, opened 2024-11-27
Provide a standard API for cloning entity instances for passing to services from UI


knstvk:
`DataContextImpl` now uses `StandardSerialization` bean when saving entities to `DataManager`. We need a similar API available to application developers.

PR #3986, opened 2024-12-10
Standard API for cloning entity instances

knstvk:
See #3937

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/Copier.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/Copier.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2024 Haulmont.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.jmix.core;
+
+/**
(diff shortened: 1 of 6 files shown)


jmix-core/core/src/main/java/io/jmix/core/impl/SerializingCopier.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.jmix.core.impl;
+
+import io.jmix.core.Copier;
+import org.springframework.stereotype.Component;
+
+@Component("core_SerializingCopier")
+public class SerializingCopier implements Copier {
+
+    private final StandardSerialization serialization;
```
glebfox, line 25: `protected`?

jmix-core/core/src/main/java/io/jmix/core/Copier.java
```diff
@@ -0,0 +1,28 @@
...
+
+package io.jmix.core;
+
+/**
+ * Creates copies of objects.
+ */
+public interface Copier {
+
+    /**
+     * Creates a deep copy of the passed object.
```
glebfox, line 25: 1. Better to add `@return` and `@param` to avoid warnings from `javaDoc` task.
2. Let's describe in more detail what "deep copy" means. At least, similar to `io.jmix.core.MetadataTools#deepCopy(T)`. Describe the differences from `MetadataTools#deepCopy(T)` as it may be unclear what to use and when.

glebfox (changes requested):

jmix-core/core/src/main/java/io/jmix/core/Copier.java
knstvk: Done

jmix-core/core/src/main/java/io/jmix/core/impl/SerializingCopier.java
knstvk: I don't think it makes sense for this trivial class. It's better to provide your own implementation of Copier interface instead of extending this class.

glebfox (approved):

PR #4002, opened 2024-12-15
Improve performance of EntityStates.isLoaded operation

knstvk:
Introduced `LoadedPropertiesInfo` interface that can be set for any entity instance in its `EntityEntry`. Different types of entities can have different implementations of this interface.

The `LoadedPropertiesInfo` of an entity instance, if its is set, is used by `EntityStates.isLoaded()` instead of direct call to `PersistentAttributesLoadChecker`.

For JPA entities, `CachingLoadedPropertiesInfo` is used by default. If the `jmix.core.disable-caching-loaded-properties` property is true in the project, the `CachingLoadedPropertiesInfo` is not created and behavior is identical to the previous.

JPA entities in managed state do not have `CachingLoadedPropertiesInfo`, because automatic loading of attributes is uncontrollable and cache invalidation is impossible. So the isLoaded check goes directly to `PersistentAttributesLoadChecker`.

For entities loaded by REST DataStore, `StaticLoadedPropertiesInfo` is used, which provides the same behavior as the set of strings introduced in Jmix 2.4.

An alternative solution was considered in https://github.com/jmix-framework/jmix/pull/4003

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/EntityEntry.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/EntityEntry.java
@@ -16,10 +16,7 @@
 
 package io.jmix.core;
 
-import io.jmix.core.entity.BaseEntityEntry;
-import io.jmix.core.entity.EntityPropertyChangeListener;
-import io.jmix.core.entity.NullableIdEntityEntry;
-import io.jmix.core.entity.SecurityState;
+import io.jmix.core.entity.*;
 import io.jmix.core.entity.annotation.JmixGeneratedValue;
 import io.jmix.core.entity.annotation.JmixId;
 import io.jmix.core.impl.EntityInternals;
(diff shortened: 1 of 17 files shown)


dtaimanov (approved):

dtaimanov:
It is a problem with back reference and replacing existing references in entities after value holder loads entity.
see `io.jmix.eclipselink.impl.lazyloading.AbstractSingleValueHolder#afterLoadValue` and `io.jmix.eclipselink.impl.lazyloading.CollectionValuePropertyHolder#afterLoadValue`

patch to reproduce the issue: 
[LoadedPropertiesInfo_should_be_updated_when_ValueHolder_replaces_back_reference_and_existi.patch](https://github.com/user-attachments/files/18183789/LoadedPropertiesInfo_should_be_updated_when_ValueHolder_replaces_back_reference_and_existi.patch)

Снимок экрана 2024-12-18 в 18 13 53

It looks like `io.jmix.eclipselink.impl.lazyloading.AbstractValueHolder#replaceToExistingReferences` should invoke `io.jmix.eclipselink.impl.lazyloading.AbstractValueHolder#registerLoadedProperty` too at approptiate places

dtaimanov (changes requested):
see https://github.com/jmix-framework/jmix/pull/4002#issuecomment-2551440172

knstvk:
> It looks like `io.jmix.eclipselink.impl.lazyloading.AbstractValueHolder#replaceToExistingReferences` should invoke `io.jmix.eclipselink.impl.lazyloading.AbstractValueHolder#registerLoadedProperty` too at approptiate places

Done

dtaimanov (approved):

Issue #4034, opened 2024-12-28
Support for substitution in the Audit add-on


NikitaShchienko:
Support for substitution in the Audit add-on, displaying the current and substitution user.

- Add a new column to the Audit tables
- Add a new field to the filter on the Audit screen with information about the replacement user.

mig-sam:
Tested on: 
Jmix version: 2.5.999-SNAPSHOT
Jmix Studio plugin version: 2.5.SNAPSHOT7172-243
IntelliJ version: IntelliJ IDEA 2024.3.1 (Community Edition)

PR #4035, opened 2024-12-28
Support for substitution in the Audit add-on #4034

m-orlova:
1. Add the "substitutedUsername" column to the EntityLogItem entity and EntityLogItem list view
2. Add combo box to filter by substituted user on EntityLogItem list view


NikitaShchienko (approved):

jmix-audit/audit/src/main/java/io/jmix/audit/impl/EntityLogImpl.java
```diff
@@ -443,6 +447,20 @@ protected String findUsername() {
         }
     }
 
+    protected String findSubstitutedUsername() {
+        UserDetails effectiveUser = currentUserSubstitution.getEffectiveUser();
+        if (effectiveUser != null) {
+            return effectiveUser.getUsername();
+        } else {
+            String username = properties.getSystemUsername();
```
NikitaShchienko, line 455: It is necessary to check how the new column works if changes are made by the system user, for example in the quartz add-on.

Gavrilov-Ivan (approved):

NikitaShchienko (approved):

PR #4070, opened 2025-01-16
NotificationIndicator does not update its counter if the user is substituted (4007)

Flaurite:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiEventPublisher.java
```diff
@@ -163,6 +167,28 @@ public void publishEventForUsers(ApplicationEvent event, @Nullable Collection<St
         clusterApplicationEventPublisher.publish(uiUserEvent);
     }
 
+    @Nullable
+    protected Collection<String> addCurrentUserIfNecessary(@Nullable Collection<String> usernames) {
```
glebfox, line 171: Let's add detailed JavaDoc that describes why we need this

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiEventPublisher.java
Flaurite: Fixed

glebfox (approved):

PR #4086, opened 2025-01-18
JmixCheckbox does not support required state and validation (4045)

Flaurite:


Flaurite:
The issue was moved to release 2.5.

PR #4134, opened 2025-01-30
Ability to clear "modified" state of a detail view

knstvk:
Introduced `io.jmix.flowui.model.DataContext#clearChanges` and  `io.jmix.flowui.view.ChangeTracker#clearChanges` methods.

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/DataContext.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/DataContext.java
@@ -24,9 +24,7 @@ import io.jmix.flowui.view.Subscribe;
 import org.springframework.lang.Nullable;
 
 import javax.annotation.CheckReturnValue;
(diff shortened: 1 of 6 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/DataContext.java
```diff
@@ -157,6 +155,14 @@ public interface DataContext {
      */
     boolean hasChanges();
 
+    /**
+     * Clears the lists of modified and removed entities. After calling this method, {@link #hasChanges()} returns false,
+     * {@link #getModified()} and {@link #getRemoved()} return empty sets.
+     */
+    default void clearChanges() {
+        throw new UnsupportedOperationException("Not implemented");
```
knstvk, line 163: Not a problem because this method is not called by the framework.

glebfox (approved):

PR #4141, opened 2025-01-31
Use UUIDv7 for entity identifiers

knstvk:
I have chosen https://github.com/f4b6a3/uuid-creator library because it can generate increasing UUIDs within a millisecond (see [wiki](https://github.com/f4b6a3/uuid-creator/wiki/1.7.-UUIDv7)). I couldn't do the same with the alternative https://github.com/cowtowncoder/java-uuid-generator.

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/entity/KeyValueEntity.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/entity/KeyValueEntity.java
@@ -97,7 +97,7 @@ public class KeyValueEntity
     }
 
     public KeyValueEntity() {
-        uuid = UuidProvider.createUuid();
+        uuid = UuidProvider.createUuidV7();
     }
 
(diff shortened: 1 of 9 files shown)


jmix-core/core/src/main/java/io/jmix/core/entity/KeyValueEntity.java
```diff
@@ -97,7 +97,7 @@ public void setAttributeValue(@NonNull String name, Object value, boolean checkE
     }
 
     public KeyValueEntity() {
-        uuid = UuidProvider.createUuid();
+        uuid = UuidProvider.createUuidV7();
```
knstvk, line 100: Not customizable, but shouldn't be a problem because KeyValueEntity is in-memory only.

Gavrilov-Ivan (approved):

PR #4150, opened 2025-02-04
Correct REST sessions and OAuth2 tokens behaviour jmix-framework/jmix_3915

dtaimanov:


jmix-sessions/sessions/src/main/java/io/jmix/sessions/SessionsConfiguration.java
```diff
@@ -65,4 +85,31 @@ public SessionRepositoryFilter<SessionRepositoryWrapper<S>.SessionWrapper> jmixS
...
     }
+
+    @Bean
+    public OAuth2TokenCustomizer<OAuth2TokenClaimsContext> tokenCustomizer(ObjectProvider<SessionData> sessionDataProvider) {
+        return (context) -> {
+            Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+            if (authentication != null
+                    && authentication.isAuthenticated()
+                    && authentication.getDetails() instanceof ClientDetails cd) {
+                /*SessionData sessionData = sessionDataProvider.getIfAvailable();
```
Gavrilov-Ivan, line 96: Do we need this comment block?
If this is a suggestion for the possible future changes - it's better to describe is with more details (what it is going to fix and why it's not done yet)
Remove otherwise.

jmix-sessions/sessions/src/main/java/io/jmix/sessions/resolver/OAuth2AndCookieSessionIdResolver.java
```diff
@@ -76,40 +80,45 @@ public void expireSession(HttpServletRequest request, HttpServletResponse respon
...
-            OAuth2RefreshToken token = tokenStore.readRefreshToken(getRefreshToken(request));
-            if (token != null) {
-                OAuth2Authentication authentication = tokenStore.readAuthenticationForRefreshToken(token);
-                if (authentication.getDetails() instanceof ClientDetails) {
-                    sessionId = ((ClientDetails) authentication.getDetails()).getSessionId();
-                }
+            auth = oAuth2AuthorizationService.findByToken(getRefreshToken(request), OAuth2TokenType.REFRESH_TOKEN);
+        }
+        if (auth != null) {
+            OAuth2Authorization.Token<OAuth2AccessToken> token = auth.getAccessToken();//is .getAccessToken() actually nullable?
```
Gavrilov-Ivan, line 90: According to the code access token can be null.
As a possible example - work with 'Authorization Code' grant type.
This flow includes getting Auth Code (OAuth2Authorization is created without Access Token value) and then exchange it for Access Token (Access Token value is initialized). 
It looks like before exchange `getAccessToken()` will return null.

jmix-sessions/sessions/src/main/java/io/jmix/sessions/resolver/OAuth2AndCookieSessionIdResolver.java
```diff
@@ -76,40 +80,45 @@ public void expireSession(HttpServletRequest request, HttpServletResponse respon
...
-            token = tokenStore.readAccessToken(tokenValue);
-            if (token != null) {
-                String originalSessionId = (String) token.getAdditionalInformation().get(SESSION_ID);
-                if (!Objects.equals(originalSessionId, sessionId)) {
-                    token.getAdditionalInformation().put(SESSION_ID, sessionId);
-                    tokenStore.storeAccessToken(token, tokenStore.readAuthentication(token));
+            OAuth2Authorization auth = oAuth2AuthorizationService.findByToken(tokenValue, OAuth2TokenType.ACCESS_TOKEN);
+
+            if (auth != null) {
+                token = auth.getAccessToken();
```
Gavrilov-Ivan, line 114: Case of Access token nullability

Gavrilov-Ivan (changes requested):

Gavrilov-Ivan (approved):

Issue #1678, opened 2023-05-05
CodeEditor suggestion


KremnevDmitry:
Implement suggestion API such as ClassicUI

trx0eth7:
Could u please tell in which release do you plan to include this feature?

SergeiAksenov2:
Tested on:
Jmix version: 2.6.999-SNAPSHOT
Jmix Studio Plugin Version: 2.6.SNAPSHOT7481-251
IntelliJ version: IntelliJ IDEA 2025.1.2 Preview (Ultimate Edition)

PR #4232, opened 2025-03-03
CodeEditor suggestions #1678

n-13:
#1678


KremnevDmitry:
Hi!

Thank you for your contribution!

Unfortunately, your solution doesn't fully cover the requirements for the task. 
The main point of the task is to implement a mechanism of suggestions that can be arbitrarily configured on the server-side. First of all, it is necessary to implement a `JpqlSuggester` that will suggest attributes of the used entity and placeholders for JPQL queries.

Also, your implementation lacks integration with the studio and implementation of declarative XML description loading.

We are starting to work on this task on our own in this release, your developments will be used as an additional source of information, thank you.

PR #4263, opened 2025-03-14
Fix KeyValueCollection multiple filters search

### LINKED ISSUE #3980: Error when filtering by more than one condition on a KeyValueCollection
### Environment

Jmix version: 1.x, 2.x

### Bug Description

On a KeyValueCollection view screen, if you filter by more than one property, then the where condition in the generated query is incorrect. The entity name is added to the property when it shouldn't be

### How to Replicate
Open the school.zip project and go to the classrooms page. There should be one class room with its associated student.

Add a filter for school id = 2. The filter works correctly.

Now add a second filter for student name contains "Jamie". An error is thrown due to bad sql.

### Current Behavior

When filtering by one or more properties, the generated sql is incorrect

### Expected Behavior

The entity name (classRoom) should not be added to the front of the property name in this case.

I have traced the bug to _io.jmix.data.impl.jpql.generator.PropertyConditionGenerator.generateWhere()_

### Project zip file

[school.zip](https://github.com/user-attachments/files/18038532/school.zip) (1.5.5)

image

jguthrie100:
Fixes #3980 

The [KeyValuePropertyConditionGenerator#L43](https://github.com/jmix-framework/jmix/blob/e2b937f0644ad0e4782741596426ccf7ad26d640/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/KeyValuePropertyConditionGenerator.java#L43) (required for filtering KeyValueCollections) checks that the context has getSelectedExpressions() != null, however, in [ConditionJpqlGenerator#L74](https://github.com/jmix-framework/jmix/blob/e2b937f0644ad0e4782741596426ccf7ad26d640/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionJpqlGenerator.java#L74) the selectedExpressions were not being copied to childContexts.

### DIFF
--- a/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionJpqlGenerator.java
+++ b/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionJpqlGenerator.java
@@ -71,9 +71,7 @@ public class ConditionJpqlGenerator {
     private void copyGenerationContext(ConditionGenerationContext generationContext) {
         for (Condition childCondition:generationContext.getChildContexts().keySet()){
             ConditionGenerationContext childContext = generationContext.getChildContexts().get(childCondition);
-            childContext.setEntityAlias(generationContext.getEntityAlias());
-            childContext.setEntityName(generationContext.getEntityName());
-            childContext.setValueProperties(generationContext.getValueProperties());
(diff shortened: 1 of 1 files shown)


jguthrie100:
@glebfox is there an official track to push this through? It would be very helpful to my firm to have this in a future release!

PR #4275, opened 2025-03-18
Checkbox should ignore the NotNull annotation and the Mandatory state (4251)

Flaurite:


Flaurite:
Should we include "validation" in the application property name? Perhaps it would be better to name the property like this:
```
jmix.ui.component.checkbox-required-state-initialization-enabled=false
```

glebfox (approved):

PR #4281, opened 2025-03-19
Séparateurs de milliers et décimal

skerain:
Séparateurs de milliers et décimal pour la France, norme ISO 80000-1:2022.


knstvk (changes requested):
The `.` and `,` symbols in the decimal format are actually placeholders for the decimal and thousand separators. So they should be presented in the formats as is. What you really need is to redefine the `numberDecimalSeparator` and `numberGroupingSeparator` messages in the same file. 

Look at how it's done for the Russian locale (it's the same as in French):
https://github.com/jmix-framework/jmix/blob/25886576152cb1866d544034ed65d7e6442518f7/jmix-translations/content/io/jmix/core/messages_ru.properties#L42

knstvk:
Closed in favor of https://github.com/jmix-framework/jmix/pull/4285

Issue #2142, opened 2023-08-14
Cannot put newLine symbol into report string arguments


KartnDev:
Source: https://forum.jmix.ru/t/perenos-stroki-v-shablone-docx/1172

m-orlova:
### Implementation notes
The `jmix.reports.multilineStringsProcessingEnabled` application property has been added. If property enabled (by default disabled), line breaks in the parameter values are written to the executed report document. Works only for DOCX report templates.

konyashkina:
Tested on
Jmix version: 2.5.999-SNAPSHOT
Jmix Studio Plugin Version: 2.5.SNAPSHOT7336-243
IntelliJ version: IntelliJ IDEA 2024.3.5

PR #4288, opened 2025-03-20
Cannot put newLine symbol into report string arguments #2142

m-orlova:
Add application property `jmix.reports.multilineStringsProcessingEnabled`. Default value: false. If enabled, line breaks in the parameter values are written to the executed report document.

### DIFF
--- a/jmix-reports/reports/src/main/java/io/jmix/reports/ReportsProperties.java
+++ b/jmix-reports/reports/src/main/java/io/jmix/reports/ReportsProperties.java
@@ -144,6 +144,11 @@ public class ReportsProperties {
      */
     boolean formulasPostProcessingEvaluationEnabled;
 
+    /**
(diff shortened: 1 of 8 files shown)


jmix-reports/reports/src/main/java/io/jmix/reports/ReportsProperties.java
```diff
@@ -144,6 +144,11 @@ public class ReportsProperties {
      */
     boolean formulasPostProcessingEvaluationEnabled;
 
+    /**
+     * If enabled - all string parameters with multi-line values will be written with preserved line breaks in the DOCX report.
+     */
+    boolean multilineStringsProcessingEnabled;
```
NikitaShchienko, line 150: Don't forget to create a documentation issue.

NikitaShchienko (approved):

PR #4290, opened 2025-03-21
Reasses sessions behaviour jmix-framework/jmix_4213

dtaimanov:


jmix-sessions/sessions/src/main/java/io/jmix/sessions/impl/JmixExpiringMapProperties.java
```diff
@@ -0,0 +1,56 @@
...
+
+package io.jmix.sessions.impl;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.DefaultValue;
+
+import java.time.Duration;
+
+@ConfigurationProperties(prefix = "jmix.sessions.expiring-map")
+public class JmixExpiringMapProperties {
```
Gavrilov-Ivan, line 25: Move this to root `SessionsProperties` and add these properties as root-level or create `expiringMap` sub-group via nested class.

Gavrilov-Ivan (changes requested):

Gavrilov-Ivan:
Add missed `package-info.java`

Gavrilov-Ivan (changes requested):
Add missed `package-info.java` files.

PR #4505, opened 2025-06-04
Ability to turn off lazy loading for the whole project. Part I 

dtaimanov:
jmix-framework/jmix#3936


jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/DisabledValueHolder.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package io.jmix.eclipselink.impl.lazyloading;
+
+import io.jmix.core.metamodel.model.MetaProperty;
+import org.eclipse.persistence.indirection.ValueHolderInterface;
+import org.springframework.beans.factory.BeanFactory;
+
+public class DisabledValueHolder extends AbstractValueHolder {
```
knstvk: I'd suggest renaming this class to `NonLoadingValueHolder`. It better explains what the class does.

Issue #4364, opened 2025-04-15
Remove HTML from default list of extensions downloaded with `Content-Disposition: inline`


knstvk:
Currently the `Content-Disposition: inline` is used for all downloads if the file extension is present in `jmix.ui.view-file-extensions` list. By default it includes HTML: `htm, html, jpg, png, jpeg, pdf`. This can lead to XSS vulnerabilities, see for example [forum topic](https://forum.jmix.io/t/xss-via-file-upload/6538).

Consider restricting the default value of the jmix.ui.view-file-extensions property to `jpg, png, jpeg, pdf`. But this  will lead to downloading HTML reports as attachments, which may be inconvenient. Perhaps we need a separate settings for downloading reports, as they can be considered as "safe HTML".

glebfox:
For future assignee: create `ReportDownloader` + app property similar to `jmix.ui.view-file-extensions`

SergeiAksenov2:
Tested on:
Jmix version: 2.6.999-SNAPSHOT
Jmix Studio Plugin Version: 2.6.SNAPSHOT7555-251
IntelliJ version: IntelliJ IDEA 2025.1.2 (Ultimate Edition)

PR #4516, opened 2025-06-09
Remove HTML from default list of extensions downloaded with Content-Disposition: inline #4364

Ki10V01T:
Necessary file extensions (htm, html) have been removed from UiProperties.java (and jmix.ui.view-file-extensions propety). 
Similar property (jmix.reports.view-file-extensions) has been added that allows you to separately configure the display of uploaded documents in the browser window for working with reports. By default, htm and html files is allowed for display inline while reports configuring.
A delegate has been added to all related classes in the reports and reports-flowui to configure the required behavior of the Downloader.
In addition, the Downloader interface and its implementation class - DownloaderImpl, were affected.

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/Downloader.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/Downloader.java
@@ -21,6 +21,8 @@ import io.jmix.core.FileStorage;
 
 import org.springframework.lang.Nullable;
 
+import java.util.function.Function;
+
 /**
  * Generic interface to download data from the system.
  */
@@ -45,6 +47,15 @@ public interface Downloader {
      * @param showNewWindow {@code true} if downloader opens new window, otherwise {@code false}
      */
     void setShowNewWindow(boolean showNewWindow);
+
+    /**
+     * Sets a delegate that checks if file should be opened or downloaded.
+     * It takes file extension as an input parameter and returns {@code true} if file is allowed to be opened or
+     * {@code false} if file should be downloaded.
+     *
+     * @param viewFileAllowanceDelegate delegate
+     */
+    void setViewFileAllowanceDelegate(Function<String, Boolean> viewFileAllowanceDelegate);
 
     /**
      * Downloads an arbitrary resource defined by a DownloadDataProvider.
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/DownloaderImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/DownloaderImpl.java
@@ -45,6 +45,7 @@ import org.springframework.lang.Nullable;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 import static jakarta.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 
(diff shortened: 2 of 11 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/Downloader.java
```diff
@@ -46,6 +48,15 @@ public interface Downloader {
      */
     void setShowNewWindow(boolean showNewWindow);
 
+    /**
+     * Sets a delegate that checks if file should be opened or downloaded.
```
Gavrilov-Ivan: Lets change "if file should be opened or downloaded" to "if file is allowed to be opened".
Even if it returns true it doesn't prevent file to be downloaded because of some another reason.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/DownloaderImpl.java
```diff
@@ -72,6 +73,9 @@ public class DownloaderImpl implements Downloader {
     // Use flags from app.properties for show/download files
     protected boolean useViewList;
 
+    // Delegate for file type property checking
+    protected Function<String, Boolean> inlineChecker = this::defaultInlineCheck;
```
Gavrilov-Ivan: Rename variable according to the name of delegate. Make default one as inline lambda

jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/runner/impl/RunMultipleReportsBackgroundTask.java
```diff
@@ -62,6 +68,18 @@ public void setReportZipUtils(ReportZipUtils reportZipUtils) {
     @Autowired
     public void setDownloader(Downloader downloader) {
         this.downloader = downloader;
+
+        setDownloaderViewFileAllowanceDelegate();
+    }
+
+    protected void setDownloaderViewFileAllowanceDelegate() {
```
Gavrilov-Ivan: This method with the same logic repeats multiple times.
Looks like we need to create some configurer as an entry point to apply the delegate.

jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/helper/ReportDownloaderConfigurer.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.jmix.flowui.download.Downloader;
+import io.jmix.reports.ReportsProperties;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
+
+@Component("report_ReportDownloaderConfigurer")
+public class ReportDownloaderConfigurer {
+
+    public void configureDownloader(Downloader downloader, ReportsProperties reportsProperties) {
```
Gavrilov-Ivan: ReportsProperties shouldn't be the method parameter

jmix-flowui/flowui/src/main/java/io/jmix/flowui/download/DownloaderImpl.java
```diff
@@ -73,8 +73,8 @@ public class DownloaderImpl implements Downloader {
     // Use flags from app.properties for show/download files
     protected boolean useViewList;
 
-    // Delegate for file type property checking
-    protected Function<String, Boolean> inlineChecker = this::defaultInlineCheck;
+    // Predicate for file type property checking
```
Gavrilov-Ivan: Predicate for open/download files check

Gavrilov-Ivan (approved):

Issue #4240, opened 2025-03-04
Tenant-specific user should be able to be substituted by users from the same tenant only


Gavrilov-Ivan:
### Environment

Jmix version: 2.5.0

### Bug Description

Currently admin can configure substitution of user from one tenant by user from another tenant. This is an invalid case.
It will also cause an exception on login as user being substituted.

### Steps To Reproduce

1. Create 2 tenants - `t1` and `t2`
2. Create 3 users: `user1` and `user2` in `t1` and user3 in `t2`
3. Login as admin and try to substitute user1

### Current Behavior

It's possible to select users from both tenants as substitution candidates (user2 and user3).

### Expected Behavior

It should be possible to select users only from the same tenant of the "original" user (user2 only).

### Sample Project
[Jmix250_multitenancy_substitution_selection.zip](https://github.com/user-attachments/files/19067366/Jmix250_multitenancy_substitution_selection.zip)

PR #4536, opened 2025-06-18
Tenant-specific user should be able to be substituted by users from the same tenant only

Gavrilov-Ivan:
See #4240

### DIFF
--- a/jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/util/UserSubstitutionCandidatePredicate.java
+++ b/jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/util/UserSubstitutionCandidatePredicate.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.jmix.securityflowui.view.usersubstitution;
+package io.jmix.securityflowui.util;
 
 import org.springframework.security.core.userdetails.UserDetails;
 
--- a/jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/view/usersubstitution/UserSubstitutionDetailView.java
+++ b/jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/view/usersubstitution/UserSubstitutionDetailView.java
@@ -26,6 +26,7 @@ import io.jmix.flowui.component.textfield.TypedTextField;
 import io.jmix.flowui.view.*;
 import io.jmix.security.usersubstitution.UserSubstitutionModel;
 import io.jmix.security.usersubstitution.UserSubstitutionPersistence;
+import io.jmix.securityflowui.util.UserSubstitutionCandidatePredicate;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.userdetails.UserDetails;
 


jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/view/usersubstitution/UserSubstitutionCandidatesFilter.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.springframework.security.core.userdetails.UserDetails;
+
+import java.util.function.BiPredicate;
+
+/**
+ * Interface to provide ability to filter user substitution candidates based on target user.
+ * Accepts the user to be substituted and another user as substitution candidate.
+ */
+public interface UserSubstitutionCandidatesFilter extends BiPredicate<UserDetails, UserDetails> {
```
knstvk: Why not `UserSubstitutionCandidatePredicate`?
When I see `Filter`, it's not immediately clear to me what should I return from `test()` to include the candidate.

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/TenantProvider.java
```diff
@@ -16,6 +16,8 @@
 
 package io.jmix.multitenancy.core;
 
+import org.springframework.security.core.userdetails.UserDetails;
+
 /**
  * Provides the current tenant id.
```
knstvk: The Javadoc should be changed, now this interface provides tenant id not only for the current user

jmix-multitenancy/multitenancy-flowui/src/main/java/io/jmix/multitenancyflowui/impl/SameTenantUserSubstitutionCandidatesFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.stereotype.Component;
+
+import java.util.Objects;
+
+/**
+ * Implementation of {@link UserSubstitutionCandidatesFilter}
+ * that allows user to be substituted by user from the same tenant only.
+ */
+@Component("mten_SameTenantUserSubstitutionCandidatesFilter")
```
knstvk: I think it will be better to instantiate this bean in autoconfiguration, with `@ConditionalOnClass(name = "io.jmix.securityflowui.view.usersubstitution.UserSubstitutionCandidatesFilter")`

In theory, `security-flowui` is not needed for `multitenancy-flowui` and may not be present. Also, `api` dependency in `.gradle` should be changed to `compileOnly`.

knstvk (changes requested):

jmix-security/security-flowui/src/main/java/io/jmix/securityflowui/view/usersubstitution/UserSubstitutionCandidatesFilter.java
Gavrilov-Ivan: Done

jmix-multitenancy/multitenancy/src/main/java/io/jmix/multitenancy/core/TenantProvider.java
Gavrilov-Ivan: Done

jmix-multitenancy/multitenancy-flowui/src/main/java/io/jmix/multitenancyflowui/impl/SameTenantUserSubstitutionCandidatesFilter.java
Gavrilov-Ivan: Done.
`compileOnly project(':security-flowui')` was added to both `multitenancy-flowui.gradle` and `multitenancy-flowui-starter.gradle`

knstvk (approved):

Issue #4400, opened 2025-05-06
Card component


Flaurite:
### Description
See https://vaadin.com/docs/latest/components/card

Note that component is available under the feature flag.

KremnevDmitry:
Related studio issue: https://youtrack.jmix.io/issue/JST-6095/Support-card-component

SergeiAksenov2:
Tested on:
Jmix version: 2.7.999-SNAPSHOT
Jmix Studio Plugin Version: 2.6.SNAPSHOT7765-252
IntelliJ version: IntelliJ IDEA 2025.2.1 (Community Edition)

PR #4580, opened 2025-07-02
Card component

KremnevDmitry:
See: #4400

Expected sudio issues:
1. Inspection for priority of `title` & `subtitle` nested elements over attributes value
2. Inspection for ignoring `title` and `subtitle` nested elements in the presence of `header` element
3. Inspection of the range of the `titleHeadingLevel` value (must be in the range 1-6)
   From Vaadin web-components source code:
   > Setting values outside the range [1, 6] can cause accessibility issues
   
Usage example:
```xml
        <card id="card" width="50em" themeNames="stretch-media horizontal">
            <media>
                <image resource="icons/icon.png" minHeight="5em"/>
            </media>
            <headerPrefix>
                <avatar name="Jmix"/>
            </headerPrefix>
            <header>
                <div classNames="flex flex-col-reverse leading-xs">
                    <h2 text="Jmix"/>
                    <div text="Modern Dev platform" classNames="uppercase text-xs text-secondary"/>
                </div>
            </header>
            <headerSuffix>
                <span text="Fullstack" themeNames="badge success"/>
            </headerSuffix>
            <content>
                <p text="Jmix is a high-level full-stack framework for enterprise web applications. It comes with advanced tooling and a rich set of functional modules."/>
                <p text="Jmix is based on Spring Boot."/>
            </content>
            <footer>
                <button id="visitButton" text="Visit site"/>
                <button id="learnModeButton" text="Learn more"/>
            </footer>
        </card>
```


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/UiComponentUtils.java
```diff
@@ -105,13 +105,18 @@ public static Component getComponent(View<?> view, String id) {
      * or an empty {@link Optional}
      */
     public static Optional<Component> findComponent(Component component, String id) {
-        if (component instanceof View<?> view) {
+        BiPredicate<Component, String> idComparator = findFragment(component) == null
+                ? UiComponentUtils::sameId
+                : FragmentUtils::sameId;
+
+        if (idComparator.test(component, id)) {
```
glebfox: This method searches for **inner** components and doesn't declare that it may return parent, i.e. container itself (and should not).

glebfox (changes requested):

Issue #4475, opened 2025-05-28
Inconsistent handling of file extensions in FileStorage implementations


knstvk:
### Environment

Jmix version: 2.5.1

### Bug Description

When using Local and AWS file storage, the following operation successfully saves the file and creates a FileRef:
```
FileRef fileRef = fileStorage.saveStream("xx yy.aa bb", inputStream)
```

But the resulting FileRef cannot be used in an entity attribute because `FileRef.fromString(fileRefStr)` method throws an exception (because of the space symbol in URL path):
```
java.net.URISyntaxException: Illegal character in path at index 59: testFs://2025/05/28/0a136c28-4c90-4e89-2bba-6e4f4a14eb5d.aa bb?name=xx+yy.aa+bb
```

### Steps To Reproduce

```
def "test"() {
    when:
    byte[] a = 'Test output'.getBytes()
    FileRef fileRef = fileStorage.saveStream("xx yy.aa bb", new ByteArrayInputStream(a))
    String fileRefStr = fileRef.toString()

    then:
    fileRef.getFileName() == 'xx yy.aa bb'
    fileRef.getContentType() == FileTypesHelper.DEFAULT_MIME_TYPE
    // fails
    FileRef.fromString(fileRefStr) == fileRef

    cleanup:
    fileStorage.removeFile(fileRef)
}
```

SergeiAksenov2:
Tested on:
Jmix version: 2.6.999-SNAPSHOT
Jmix Studio Plugin Version: 2.6.SNAPSHOT7681-251
IntelliJ version: IntelliJ IDEA 2025.1.3 (Community Edition)

PR #4598, opened 2025-07-10
Inconsistent handling of file extensions in FileStorage implementations #4475

Ki10V01T:
Fixes: #4475

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/FileRef.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/FileRef.java
@@ -16,6 +16,7 @@
 
 package io.jmix.core;
 
+import com.google.common.net.UrlEscapers;
 import io.jmix.core.common.util.URLEncodeUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -177,23 +178,12 @@ public class FileRef implements Serializable {
         }
 
         StringBuilder result = new StringBuilder();
-        String[][] characterMatchingTable = {
-                {" ", "(", ")", ",", "/", ":", ";", "<", "=", ">", "?", "@", "[", "\\", "]", "{", "}"},
-                {"%20", "%28", "%29", "%2C", "%2F", "%3A", "%3B", "%3C", "%3D", "%3E", "%3F", "%40", "%5B", "%5C", "%5D", "%7B", "%7C"}
-        };
+        String escapedExtension = UrlEscapers.urlPathSegmentEscaper().escape(splittedPath[fileExtensionId]);
 
-        for (int i = 0; i < characterMatchingTable[0].length; i++) {
-            if (splittedPath[fileExtensionId].contains(characterMatchingTable[0][i])) {
-                splittedPath[fileExtensionId] = splittedPath[fileExtensionId]
-                        .replace(characterMatchingTable[0][i], characterMatchingTable[1][i]);
-            }
-        }
-
-        result.append(splittedPath[0])
+        return result.append(splittedPath[0])
                 .append(".")
-                .append(splittedPath[fileExtensionId]);
-
-        return result.toString();
+                .append(escapedExtension)
+                .toString();
     }
 
     /**
--- a/jmix-core/core/src/main/java/io/jmix/core/FileRef.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/FileRef.java
@@ -176,13 +176,9 @@ public class FileRef implements Serializable {
             return fileName;
         }
 
-        StringBuilder result = new StringBuilder();
         String escapedExtension = UrlEscapers.urlPathSegmentEscaper().escape(fileExtension);
 
-        return result.append(fileName)
-                .append(".")
-                .append(escapedExtension)
-                .toString();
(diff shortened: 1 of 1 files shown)


jmix-core/core/src/main/java/io/jmix/core/FileRef.java
```diff
@@ -157,6 +160,32 @@ public String toString() {
...
 
+    /**
+     * Transforms file extension in path in according to the RFC 7230 standard if presents. <br>
+     * Expects the {@link FileRef#path} to have an extension at the end or not. <br>
+     * Example with file extension:
+     * <pre>
+     *     1970/01/01/12345678-1234-5678-0000-123456xxxxxx.extension
+     * </pre>
+     */
+    private String transformFileExtensionToUri() {
```
Gavrilov-Ivan: "encodePath". Adjust javadoc accordingly

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
```diff
@@ -157,6 +160,32 @@ public String toString() {
         return uriStringBuilder.toString();
     }
 
+    /**
+     * Transforms file extension in path in according to the RFC 7230 standard if presents. <br>
+     * Expects the {@link FileRef#path} to have an extension at the end or not. <br>
```
Gavrilov-Ivan: What is the purpose of this line?

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
Gavrilov-Ivan: Provide path as argument

Gavrilov-Ivan: Make it protected

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
```diff
@@ -157,6 +160,32 @@ public String toString() {
...
+
+        // checks for missing file extension
+        if (fileExtensionId == 0) {
+            return splittedPath[0];
+        }
+
+        StringBuilder result = new StringBuilder();
+        String escapedExtension = UrlEscapers.urlPathSegmentEscaper().escape(splittedPath[fileExtensionId]);
+
+        return result.append(splittedPath[0])
```
Gavrilov-Ivan: Original path is split by dot. Then the **last** element is escaped (if it's not the only one). Then the **first** element and the **last** element (extension) are joined.
With UUID name it shouldn'd be a problem, but formally it's not correct.
It should be either join of **all** elements (except the last one) together with escaped extension or original path should be always split into 2 parts maximum - by the **last** dot. Then escape the second one if exists and join them back.

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
```diff
@@ -157,6 +160,32 @@ public String toString() {
...
+     * Transforms file extension in path in according to the RFC 7230 standard if presents. <br>
+     * Expects the {@link FileRef#path} to have an extension at the end or not. <br>
+     * Example with file extension:
+     * <pre>
+     *     1970/01/01/12345678-1234-5678-0000-123456xxxxxx.extension
+     * </pre>
+     */
+    private String transformFileExtensionToUri() {
+        String[] splittedPath = path.split("\\.");
+        int fileExtensionId = splittedPath.length-1;
```
Gavrilov-Ivan: Spaces

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
Ki10V01T: This is a leftover from the previous version of the method. Removed

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
Ki10V01T: Done

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
Ki10V01T: Done

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
Ki10V01T: Has been rewrite to using apache.commons.io.FilenameUtils

jmix-core/core/src/main/java/io/jmix/core/FileRef.java
```diff
@@ -157,6 +160,31 @@ public String toString() {
...
+
+        // checks for missing file extension
+        if (StringUtils.isEmpty(fileExtension)) {
+            return fileName;
+        }
+
+        StringBuilder result = new StringBuilder();
+        String escapedExtension = UrlEscapers.urlPathSegmentEscaper().escape(fileExtension);
+
+        return result.append(fileName)
```
Gavrilov-Ivan: Do via simple concatenation instead of StringBuilder. Concatenation of fixed amount of Strings will be optimized and shorter

Ki10V01T: Done

PR #4606, opened 2025-07-15
Unfetched attribute error when loading an instance name in case of entities hierarchy jmix-framework/jmix_2860

dtaimanov:


jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
```diff
@@ -106,6 +106,11 @@ public class CoreProperties {
      */
     boolean skipNullOrEmptyConditionsByDefault;
 
+    /**
+     * Whether to use ancestor's @InstanceName definition in case of unfetched attributes.
```
Gavrilov-Ivan, line 110: Lets add some description for outcomes of true/false values

jmix-core/core/src/main/java/io/jmix/core/InstanceNameProvider.java
```diff
@@ -45,6 +45,26 @@ public interface InstanceNameProvider {
      */
     String getInstanceName(Object instance);
 
+    /**
+     * Gets entity instance name defined by {@link InstanceName} annotation in {@code metaClass} or its ancestor.
+     *
+     * @param instance   instance
+     * @param metaClass   ancestor to get string representation from
```
Gavrilov-Ivan: According to general description this class acts like a starting point to check for instance name - provided class itself and it's ancestors. And it can be the _actual class of provided instance_.

But param description explicitly states that it is _ancestor_ (doesn't include the _actual class of provided instance_). 

Description should be more clear about it - it can be any entity class within hierarchy started from the class of the provided instance.

Also, should we mention here that ancestors check will be performed only if fallback is enabled?

(Same for the other method).

jmix-core/core/src/main/java/io/jmix/core/impl/InstanceNameProviderImpl.java
```diff
@@ -46,6 +47,7 @@
 
 @Component("core_InstanceNameProvider")
 public class InstanceNameProviderImpl implements InstanceNameProvider {
+    public static final String UNFETCHED_EXCEPTION_MESSAGE_PREFIX = "Cannot get unfetched attribute [";
```
Gavrilov-Ivan, line 51: Empty line at the start of class body

jmix-core/core/src/main/java/io/jmix/core/impl/InstanceNameProviderImpl.java
```diff
@@ -102,6 +107,30 @@ public InstanceNameRec(String format, @Nullable Method method, MetaProperty[] na
         }
     }
 
+    protected static class EvaluationResult {
+        public final String instanceName;
```
Gavrilov-Ivan, line 113: Empty line

jmix-core/core/src/main/java/io/jmix/core/impl/InstanceNameProviderImpl.java
```diff
@@ -102,6 +107,30 @@ public InstanceNameRec(String format, @Nullable Method method, MetaProperty[] na
...
+            this.instanceName = "";
+            this.exception = exception;
+        }
+
+        public boolean isSuccessful() {
+            return exception == null;
+        }
+
+        public String getUnfetchedAttributeName() {
+            String message = Objects.requireNonNull(exception).getMessage();
```
Gavrilov-Ivan: Should it throw exception if it was called on "successful" result?
Maybe just return null?

jmix-core/core/src/main/java/io/jmix/core/CoreProperties.java
```diff
@@ -106,6 +106,11 @@ public class CoreProperties {
...
+     */
+    boolean allowInstanceNameFallback;
```
Gavrilov-Ivan: Maybe rename to "xxxEnabled" variant like many other properties?

Gavrilov-Ivan (changes requested):

Issue #4563, opened 2025-06-26
ZipProject task includes unnecessary large folders


alexbudarov:
### Environment

Jmix version: 2.5.3

### Bug Description

ZipProject task creates a zip archive which is too big.
Archive includes generated frontend files in the autogenerated `src/main/frontend/generated` directory.

In a small sample project total content size is 3.4 MB, and 3.3 MB is from this unnecessary folder.

Looks like in some recent Vaadin version files location has been changed.

### Steps To Reproduce

- Create new project using Jmix 2.5
- Launch project, wait until frontend bundle is built
- Run zipProject
- Examine contents of the zip file

### Current Behavior

Archive includes generated frontend files in the autogenerated `src/main/frontend/generated` directory.
Archive size is big (1 MB)

### Expected Behavior

Archive size is small.

### Sample Project

[codeid.zip](https://github.com/user-attachments/files/20929203/codeid.zip)

Ki10V01T:
@glebfox There are some features that prevent the gitignore file from being archived.
In Gradle versions below 8, it was possible to change excluded files by default (including gitignore) inside the plugin during the build. Starting with version 8, this feature stopped working (the build fails). Now, by default, you need to move this logic to the settings.gradle file [(github-gradle-link)](https://github.com/gradle/gradle/issues/27225#issuecomment-1859282996) [(second-link-with-same-problem-description)](https://github.com/gradle/gradle/issues/26600). The same is written in the [official documentation](https://docs.gradle.org/current/userguide/working_with_files.html#sec:file_trees). But this approach only works if you remove .gitignore from the exceptions at the working project level, and not at the level of our plugin. In this case, this file will be archived, as expected. But if you do not do this, the settings from settings.gradle of our plugin will not be applied and .gitignore will be hidden (default behavior)
Perhaps, you should add a script (or amendments to the migration?), changing the settings.gradle file in a certain way when creating a new project.
Anyway, I added a rule inside our plugin, and if the user explicitly removes gitignore from the excluded files - the file will be added to the archive

PR #4620, opened 2025-07-21
ZipProject task includes unnecessary large folders #4563

Ki10V01T:


jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/ZipProject.groovy
```diff
@@ -53,7 +53,7 @@ class ZipProject extends DefaultTask {
         includeToZip += this.includeToZip
 
         def excludeFromZip = [
-                project.rootDir.name + '/frontend/generated',
```
Gavrilov-Ivan, line 56: Keep the old one, just in case.
This directory was moved, but maybe some old projects still have it.

Gavrilov-Ivan (changes requested):

jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/ZipProject.groovy
Ki10V01T: Done.
I also added to the issue discussion a description of why it is not possible to add a gitignore file from the plugin code.

PR #4641, opened 2025-07-29
Authorization server: invalidate token on logout

Gavrilov-Ivan:


jmix-authserver/authserver/src/main/java/io/jmix/authserver/authentication/TokenRevocationLogoutHandler.java
```diff
@@ -0,0 +1,73 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.authorization.OAuth2Authorization;
+import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationService;
+import org.springframework.security.oauth2.server.authorization.OAuth2TokenType;
+import org.springframework.security.web.authentication.logout.LogoutHandler;
+
+/**
+ * {@link LogoutHandler} that tries to get bearer token request and remove it from token storage.
```
dtaimanov: Sounds like request will be removed from storage instead of the token. Maybe rephrase to clarify?

dtaimanov (approved):

jmix-authserver/authserver/src/main/java/io/jmix/authserver/authentication/TokenRevocationLogoutHandler.java
Gavrilov-Ivan: "Get **token** and remove it from **token** storage" - so it doesn't look like removing request.
But there is typo - it should be "...from the request...".
Added more details.

PR #4676, opened 2025-08-25
Lazy loaded reference returns wrong result for nested back references jmix-framework/jmix_3507

dtaimanov:
When cache enabled.

### DIFF
--- a/jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/JpaLazyLoadingListener.java
+++ b/jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/JpaLazyLoadingListener.java
@@ -268,7 +268,7 @@ public class JpaLazyLoadingListener implements DataStoreEventListener {
 
         StringBuilder builder = new StringBuilder().append(originalValueHolder.getClass().getSimpleName());
         Object currentValueHolder = originalValueHolder;
-        int iteration = 0;
+        int iteration = 0; //Just in case. Prevents potential cyclic reference chains (e.g., VH1->VH2, VH2->VH3, VH3->VH1)
         while (currentValueHolder instanceof WrappingValueHolder<?> wrappingValueHolder
                 && wrappingValueHolder.getWrappedValueHolder() != null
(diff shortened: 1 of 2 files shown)


jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/ValueHoldersSupport.java
```diff
@@ -127,6 +145,17 @@ public static QueryBasedValueHolder unwrapToQueryBasedValueHolder(Object valueHo
         return null;
     }
 
+    @Nullable
+    public static Object unwrapToValueIfInstantiated(Object valueHolder) {
+        if (valueHolder instanceof WrappingValueHolder<?> wrappingValueHolder && wrappingValueHolder.getWrappedValueHolder() != null) {
+            ValueHolderInterface<?> wrappedHolder = wrappingValueHolder.getWrappedValueHolder();
+            if (wrappedHolder != null && wrappedHolder.isInstantiated() && !(wrappedHolder instanceof NoValueCollection)) {
```
Gavrilov-Ivan: Do we need this null-check here? It was already checked in outer if-statement.

jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/JpaLazyLoadingListener.java
```diff
@@ -224,10 +246,42 @@ protected void processManyToOneValueHolder(Object owner, MetaProperty property,
...
         }
     }
 
+    protected void logUnprocessedValueHolder(Object owner, MetaProperty property, Object originalValueHolder) {
+        final int MAX_ITERATION = 5;
+
+        StringBuilder builder = new StringBuilder().append(originalValueHolder.getClass().getSimpleName());
+        Object currentValueHolder = originalValueHolder;
+        int iteration = 0;
+        while (currentValueHolder instanceof WrappingValueHolder<?> wrappingValueHolder
```
Gavrilov-Ivan, line 272: Add some description about cases when we need these iterations

PR #4709, opened 2025-09-09
Error creating UUID from database value when trying log in application (mariaDB)

dtaimanov:
### Changes:
1. Liquibase version is locked to 4.25.0 in the `bom.gradle` which makes it generate char uuid columns as before.
2. UUID column generation and support can be enabled by:
 - setting `jmix.eclipselink.use-mariadb-uuid-id` to `true`
 - adding `implementation 'org.liquibase:liquibase-core:4.26.0!!` (4.25.**1** or later) to `build.gradle`
 - adding `-` to all uuid values in changelog
In such case the database should be new or ID and reference columns should be migrated to `UUID` types


jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/mapping/UuidMappingProcessor.java
```diff
@@ -93,6 +93,10 @@ private void setDatabaseFieldParameters(Session session, DatabaseField field) {
             field.setSqlType(Types.VARCHAR);
             field.setType(String.class);
             field.setColumnDefinition("varchar(32)");
+        } else if (session.getPlatform() instanceof MariaDBPlatform) {
```
knstvk, line 96: So VARCHAR will be used for both JmixMariaDBPlatform and JmixMariaDB107Platform. 
Is this OK?

dtaimanov: As far as I can see, this method (`io.jmix.eclipselink.impl.mapping.UuidMappingProcessor#setDatabaseFieldParameters`) is a legacy method, used as a backup option if DatabasePlatform not instanceof UuidMappingInfo, i.e. not from Jmix. For JmixMariaDBPlatform and JmixMariaDB107Platform their `io.jmix.eclipselink.impl.dbms.UuidMappingInfo#getUuidSqlType` will be used instead.

PR #4734, opened 2025-09-19
Dialog view cannot be opened from the done method of BackgroundTask (4664)

Flaurite:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/backgroundtask/ThreadLocalVaadinRequest.java
```diff
@@ -0,0 +1,52 @@
...
+ * request to newly spawned threads that do not inherently have access to the data of the original Vaadin request.
+ * <p>
+ * This class mainly used to provide access to Vaadin request within {@link UI#access(Command)} closure in background
+ * tasks. It is not recommended to use this class in the application code.
+ * <p>
+ * Ensure that you clear the thread-local storage after use. Use {@link #clear()} method once the thread-local data is
+ * no longer needed.
+ */
+@Internal
+public class ThreadLocalVaadinRequest {
```
glebfox, line 36: Probably we need t rename it to ThreadLocalVaadinRequestStorage/Holder/Wrapper as doc declares, because it's not a request itself

glebfox (changes requested):

Issue #4184, opened 2025-02-14
Rework facet registration and loading mechanism


Flaurite:
### Description
Current mechanism of registration and loading facets has several disadvantages:
1) `FacetProvider` interface combines two purpose: registration and loading facet from XML.
2) Hardcoded facet class in `FacetProvider` that makes impossible to reuse XML loading logic.

It is required to rework this mechanism to make it more flexible like "Component Registration".

mig-sam:
Tested on:
Jmix version: 2.8.999-SNAPSHOT
Jmix Studio Plugin Version: 2.8.SNAPSHOT8056-252
IntelliJ version: IntelliJ IDEA 2025.2 (Community Edition)

PR #4895, opened 2025-11-20
Rework facet registration and loading mechanism jmix-framework/jmix#4184

KremnevDmitry:
See: #4184

Facet registration now contains three main components: an implementation, a loader, and a tag. It is possible to register new combinations of these elements using `FacetRegistrationBuilder`.

For example:
```java
  @Configuration
  public class FacetConfiguration {
 
      @Bean
      public FacetRegistration extTimerFacet() {
          return FacetRegistrationBuilder.create(ExtTimerFacetImpl.class)
                  .replaceFacet(Timer.class)
                  .withFacetLoader("timer", ExtTimerFacetLoader.class)
                  .build();
      }
  }
```

The facet provider mechanism is marked as `@Deprecated`. When upgrading, standard Jmix facet overrides will no longer work for end-users, but custom facets will continue to work due to backward compatibility.

### DIFF
--- a/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/xml/facet/loader/DynAttrFacetLoader.java
+++ b/jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/xml/facet/loader/DynAttrFacetLoader.java
@@ -19,6 +19,8 @@ package io.jmix.dynattrflowui.xml.facet.loader;
 import io.jmix.dynattrflowui.DynAttrEmbeddingStrategies;
 import io.jmix.dynattrflowui.facet.DynAttrFacet;
 import io.jmix.flowui.component.UiComponentUtils;
+import io.jmix.flowui.impl.FacetsImpl;
+import io.jmix.flowui.xml.facet.FacetProvider;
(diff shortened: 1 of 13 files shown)


jmix-dynattr/dynattr-flowui/src/main/java/io/jmix/dynattrflowui/xml/facet/loader/DynAttrFacetLoader.java
```diff
@@ -0,0 +1,48 @@
...
+import io.jmix.flowui.component.UiComponentUtils;
+import io.jmix.flowui.xml.facet.loader.AbstractFacetLoader;
+
+public class DynAttrFacetLoader extends AbstractFacetLoader<DynAttrFacet> {
+
+    protected DynAttrEmbeddingStrategies embeddingStrategies;
+
+    @Override
+    protected DynAttrFacet createFacet() {
+        return facets.create(DynAttrFacet.class);
```
glebfox, line 32: `DynAttrFacetProvider` invokes `setAttributeDefaultValues` when creating a facet

glebfox (changes requested):

PR #4905, opened 2025-11-24
Support for H2 database

knstvk:
See also https://github.com/jmix-framework/jmix-db-tests-3 project

### DIFF
--- a/jmix-messagetemplates/messagetemplates/src/main/resources/io/jmix/messagetemplates/liquibase/changelog/001-message-templates.xml
+++ b/jmix-messagetemplates/messagetemplates/src/main/resources/io/jmix/messagetemplates/liquibase/changelog/001-message-templates.xml
@@ -151,7 +151,7 @@
                                  referencedColumnNames="ID"/>
     </changeSet>
 
-    <changeSet id="2" author="message-templates" dbms="mssql, hsqldb, oracle">
+    <changeSet id="2" author="message-templates" dbms="mssql, hsqldb, h2, oracle">
         <createIndex indexName="IDX_MSGTMP_MESSAGE_TEMPLATE_GROUP_UNQ_NAME"
                      tableName="MSGTMP_MESSAGE_TEMPLATE_GROUP"
                      unique="true">
--- a/jmix-multitenancy/multitenancy/src/main/resources/io/jmix/multitenancy/liquibase/changelog/001-multitenancy.xml
+++ b/jmix-multitenancy/multitenancy/src/main/resources/io/jmix/multitenancy/liquibase/changelog/001-multitenancy.xml
@@ -83,7 +83,7 @@
         </rollback>
     </changeSet>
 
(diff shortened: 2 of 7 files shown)


jmix-data/data/src/main/java/io/jmix/data/impl/dbms/H2DbTypeConverter.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.sql.*;
+import java.util.Date;
+import java.util.UUID;
+
+@Component("h2DbTypeConverter")
+public class H2DbTypeConverter implements DbTypeConverter {
+
+    @Override
+    public Object getJavaObject(ResultSet resultSet, int columnIndex) {
```
Gavrilov-Ivan, line 34: Minor. Return value in `DbTypeConverter` interface is marked as `Nullable`, but methods in implementations are marked as `NonNull` via `package-info` what makes IDE thinks they are effectively `NonNull`. That adds a warning where it shouldn't be (implementation itself) and doesn't add it where it should be (any calling code).

But we don't call this method directly (and it doesn't look we will), so it's not a big problem.

It's actual for all `DbTypeConverter` implementations.

jmix-authserver/authserver/src/main/resources/io/jmix/authserver/liquibase/changelog.xml
```diff
@@ -39,6 +39,9 @@
     <property name="authserver.data.type"
               value="BLOB"
               dbms="hsqldb"/>
+    <property name="authserver.data.type"
```
Gavrilov-Ivan, line 42: There are more modules with dbms-specific liquibase scripts\properties: message-templates, multitenancy, quartz, reports, sample-rest-service, securitydata

jmix-data/data/src/main/java/io/jmix/data/impl/dbms/H2DbTypeConverter.java
knstvk: Fixed

jmix-authserver/authserver/src/main/resources/io/jmix/authserver/liquibase/changelog.xml
knstvk: Fixed for Quartz and sample-rest-service. Others have only oracle/!oracle conditions.

Gavrilov-Ivan: Scripts `001-message-templates.xml`, `001-multitenancy.xml`, `001-reports.xml`, `003-reports.xml`, `004-reports.xml` and `001-security-data.xml` has changesets with explicitly specified databases.

`jmix-reports/reports/src/main/resources/io/jmix/reports/liquibase/changelog.xml` also has database-specific property `byte_array.type`

knstvk: Done

PR #4911, opened 2025-11-25
The instance container loader delegate is not called in the standard view

### LINKED ISSUE #4908: Instance container loader delegate is not called in standard view
### Environment

Jmix version: 2.x

### Bug Description

The delegate is not invoked in the instance container loader in the Standard view.

### Steps To Reproduce

1. Create a standard view.
2. Add the Instance container to the Standard  view descriptor in the loader without specifying a query string.
3. Add a Form layout to the Standard  view to display values ​​from the instance container.
4. In the Standard view controller, add a delegate call to set the entity instance to the instance container.
5. Run the application.
6. Open the created standard view.
7. The fields form layout will remain empty.

### Current Behavior

In the `needLoad()` method, `io.jmix.flowui.model.impl.InstanceLoaderImpl` only checks for the presence of an entityId or queryString and does not check for the presence of a delegate to obtain an entity instance to install into the container. As a result, the installed delegate is not invoked when the Standard view is opened.
To work around this, you can specify a querystring in the instance container loader. However, when using a delegate, this is likely not necessary.
In JMix 1.x, obtaining an entity instance via a delegate in the standard view worked without issue. The `needLoad()` method was called after checking for the presence of the delegate.

### Expected Behavior

When a delegate is set, the entity instance must be set to the instance container in the standard view, regardless of entityId or query.
This is how it was done in JMix 1.x.

kxb-pro:
1. The load method in InstanceLoaderImpl has been reworked to allow calling a delegate if it is specified and the entityId and query are not specified.
2. The needLoad method has been renamed to skipLoading. The method is now called without constantly inverting the result.
3. The resolveFetchPlan method has been reworked for greater readability. The @Nullable annotation has been added.
4. Removed double inversion of the sendPreLoadEvent method result. CollectionLoaderImpl
1. The resolveFetchPlan method has been reworked for greater readability. The @Nullable annotation has been added.
2. Removed double inversion of the sendPreLoadEvent method result. JmixImageTestView
1. Replaced the call to the deprecated org.apache.commons.lang3.RandomUtils.nextBytes method.

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/CollectionLoaderImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/CollectionLoaderImpl.java
@@ -114,7 +114,7 @@ public class CollectionLoaderImpl<E> implements CollectionLoader<E> {
 
         LoadContext<E> loadContext = createLoadContext();
 
-        if (!sendPreLoadEvent(loadContext)) {
+        if (sendPreLoadEvent(loadContext)) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


knstvk (changes requested):
@kxb-pro - thank you for the PR.

Please split it into two - one for the linked issue with instance container loader delegate and second for other minor issues with resolveFetchPlan and JmixImageTestView.

Also:

> The needLoad method has been renamed to skipLoading.
> Removed double inversion of the sendPreLoadEvent method result

Don't change protected methods if it's not necessary for your task. We try to avoid potentially breaking changes without strong motivation.

knstvk:
Closing in favor of focused fix in https://github.com/jmix-framework/jmix/commit/fe642ee64bce3e0ba86f62488d35d182bb2c82f4

PR #4930, opened 2025-12-03
Support for ElementCollection and transient collections

knstvk:


jmix-flowui/flowui/src/main/java/io/jmix/flowui/data/value/ContainerValueSource.java
```diff
@@ -372,4 +376,17 @@ protected boolean isEqualCollectionValue(@Nullable V a) {
         //noinspection ConstantConditions
         return CollectionUtils.isEqualCollection(newValue, oldValue);
     }
+
+    protected void setCollectionValue(E item, @Nullable V value) {
```
knstvk, line 380: Before this change, MultiValuePicker and similar components could not be bound to a Set attribute.

jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/AbstractValueHolder.java
```diff
@@ -105,10 +105,8 @@ public boolean shouldAllowInstantiationDeferral() {
 
     @Override
     public DatabaseMapping getMapping() {
-        if (LazyLoadingContext.isDisabled()) {
-            return originalValueHolder.getMapping();
-        }
-        throw new RuntimeException("Unsupported by lazy loading");
+        // Called e.g. by not-instantiated IndirectList.add()
+        return originalValueHolder.getMapping();
```
knstvk, line 109: Before this change, calling `add()` on a lazily loaded collection thrown an exception.

dtaimanov (approved):

Issue #5000, opened 2025-12-25
UnconstrainedDataManager.save method JavaDoc


fractal3000:
`UnconstrainedDataManager#save(io.jmix.core.SaveContext)` method JavaDoc doesn't contain information about reloading data. Especially, it is important to show that each entity is reloaded separately with its own fetch plan.

Additionally, it is good to put such information into documentation and to the studio's performance tips.

PR #5032, opened 2026-01-23
UnconstrainedDataManager.save method JavaDoc #5000

fractal3000:
#5000

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/UnconstrainedDataManager.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/UnconstrainedDataManager.java
@@ -64,8 +64,8 @@ public interface UnconstrainedDataManager {
 
     /**
      * Saves entity instances to their data stores.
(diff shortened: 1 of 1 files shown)


jmix-core/core/src/main/java/io/jmix/core/UnconstrainedDataManager.java
```diff
@@ -64,18 +64,25 @@ public interface UnconstrainedDataManager {
 
     /**
      * Saves entity instances to their data stores.
+     * <p><b>Note:</b> Each object is reloaded separately with its own fetch plan.
```
knstvk: Remove `<b>Note:</b>` here and below. 
Also, change the sentence a little: 
`Each entity is reloaded separately with its own fetch plan. If the fetch plan is not specified in {@link SaveContext}, it's created using the current state of the entity.`

fractal3000: Fixed

Issue #5029, opened 2026-01-22
The non-obvious order of actions when creating a Band with the DataSet type - Entity


Ki10V01T:
It's not obvious that before creating a Band with the `DataSet` -> `Entity` type, you first need to set the `Input parameters` in the next tab to enable entity selection ability in the `Select` field. I suggest adding a `tooltip` in this field reminding you that for the values ​​to appear, you must first set the `Input parameter` in the next tab, then go back and select a value in the `Entity parameter name` field.

PR #5042, opened 2026-01-28
The non-obvious order of actions when creating a Band with the DataSet type - Entity #5029

Ki10V01T:
See: #5029


jmix-reports/reports-flowui/src/main/java/io/jmix/reportsflowui/view/report/ReportDetailView.java
```diff
@@ -409,6 +413,28 @@ protected void onBandsTreeDataGridRemove(ActionPerformedEvent event) {
         bandsTreeDataGrid.focus();
     }
 
+    protected void initEntitiesParamFieldTooltip() {
```
Gavrilov-Ivan, line 416: It still works as a hover on any part of comboBox so there is no benefit of having a helper button.
Also helper button as a prefix doesn't look good. But it can't be added as a suffix to comboBox.

As a variant - add helper button outside of comboBox, on the right side of it

PR #5058, opened 2026-02-05
Add Auto-Instrumentation for distributed tracing

KremnevDmitry:
Added custom spans for the following cases:
- Core
   - DataManager loading
   - DataManager saving
- UI
   - Views lifecycle
   - Actions execution

Implemented observation context propagation for background tasks.

The following application properties have been added:
- `jmix.core.data-observation-enabled` with `false` default value
- `jmix.core.use-user-info-for-observation` with `true` default value
- `jmix.ui.ui-observation-enabled` with `false` default value

The value of the properties controls the creation of custom spans in the application.

[observability-test.zip](https://github.com/user-attachments/files/25119063/observability-test.zip)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/View.java
```diff
@@ -143,7 +145,8 @@ public void afterNavigation(AfterNavigationEvent event) {
         afterNavigationProcessed = false;
 
         Sample sample = start(meterRegistry);
-        fireEvent(new ReadyEvent(this));
+        getUiObservationSupport().createViewEventObservation(this, ReadyEvent.class.getSimpleName())
+                .observe(() -> fireEvent(new ReadyEvent(this)));
         stopViewTimerSample(sample, meterRegistry, READY, getId().orElse(null));
```
fractal3000, line 151: Why do we need to use Timer and Observation at the same time? As I know, Observation contains Timer functions inside.

jmix-core/core/src/main/java/io/jmix/core/impl/UnconstrainedDataManagerImpl.java
```diff
@@ -94,7 +98,8 @@ public <E> E load(LoadContext<E> context) {
         context.setAccessConstraints(mergeConstraints(context.getAccessConstraints()));
 
         @SuppressWarnings("unchecked")
-        E entity = (E) storage.load(context);
+        E entity = (E) observationSupport.createEntityLoadObservation(metaClass, context.getQuery())
```
fractal3000, line 101: Can we use it not only in the DataManager, but also in the DataLoader(s) (when we currently use Timers)?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/action/ObservableBaseAction.java
```diff
@@ -0,0 +1,124 @@
...
+
+import java.util.function.Consumer;
+
+
+/**
+ * Base implementation of {@link Action} that introduces execution observation support.
+ *
+ * @see UiObservationSupport
+ */
+public class ObservableBaseAction extends BaseAction {
```
glebfox: Let's make it typed while we can. Overriding all fluent API methods to change the return type is a nightmare. 

Take a look at the `io.jmix.flowui.action.usermenu.UserMenuAction` as an example.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/action/ObservableBaseAction.java
```diff
@@ -0,0 +1,124 @@
...
+            return uiObservationSupport;
+        }
+
+        // try to instantiate bean in case of action created by a constructor
+        UI ui = UI.getCurrent();
+        if (ui != null) {
+            try {
+                uiObservationSupport = Instantiator.get(ui)
+                        .getOrCreate(UiObservationSupport.class);
+            } catch (Exception e) {
```
glebfox: IDE should highlight an exception variable with warning, that it's unused. Change name to `ignored`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/backgroundtask/impl/BackgroundWorkerImpl.java
```diff
@@ -81,6 +85,11 @@ public void setProperties(UiBackgroundTaskProperties properties) {
         createThreadPoolExecutor();
     }
 
+    @Autowired(required = false)
+    public void setObservationRegistry(ObservationRegistry observationRegistry) {
```
glebfox: Field injection

jmix-flowui/flowui/src/main/java/io/jmix/flowui/backgroundtask/impl/BackgroundWorkerImpl.java
```diff
@@ -397,9 +406,20 @@ public final BackgroundTask<T, V> getTask() {
         @ExecutedOnUIThread
         @Override
         public final void startExecution() {
+            Observation parentObservation = observationRegistry == null
```
glebfox: Could you please flip it xD

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/View.java
glebfox: We cannot remove already existing monitoring functionality. But we can deprecated it. @KremnevDmitry let's discuss this.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/View.java
```diff
@@ -177,7 +180,8 @@ public void beforeEnter(BeforeEnterEvent event) {
         fireEvent(new QueryParametersChangeEvent(this, event.getLocation().getQueryParameters()));
 
         Sample sample = startTimerSample(meterRegistry);
-        fireEvent(new BeforeShowEvent(this));
+        getUiObservationSupport().createViewEventObservation(this, BeforeShowEvent.class.getSimpleName())
```
glebfox: What about Fragments? They also have lifecycle events

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiObservationSupport.java
```diff
@@ -0,0 +1,85 @@
...
+ * @see UiProperties#isUiObservationEnabled()
+ */
+@Internal
+@Component("flowui_UiObservationSupport")
+public class UiObservationSupport {
+
+    @Autowired(required = false)
+    protected ObservationRegistry observationRegistry;
+
+    @Value("${jmix.ui.ui-observation-enabled:false}")
```
glebfox: `UiProperties` defines the default value. Better to keep it in the single place.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiObservationSupport.java
```diff
@@ -0,0 +1,85 @@
...
+    protected Boolean observationEnabled;
```
glebfox: `boolean`?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiObservationSupport.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        return Observation.createNotStarted(VIEW_OBSERVATION_NAME, observationRegistry)
+                .contextualName("fire view event")
+                .lowCardinalityKeyValue("event.name", eventName)
+                .lowCardinalityKeyValue("view.id", view.getId().orElse(""));
+    }
+
+    // the method must be static to avoid NPE in the case where the bean object can't be instantiated
+    public static Observation createActionExeutionObservation(Action action,
```
glebfox: Image

we need to talk....

jmix-flowui/flowui/src/main/java/io/jmix/flowui/action/ObservableBaseAction.java
glebfox: Upd: `return null;`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/View.java
KremnevDmitry: These approaches are slightly different.

#### Observation
Yes, observations creates a timer function, but it's used to track the execution time of a specific request. Furthermore, this function may not be created depending on a particular observability configuration.

Also production applications use request sampling: not all requests create a trace, which means not all requests will create a timer. This does not fit with the concept of the collected timer metric as such.

#### Timer metrics
Metrics are used to track application statistics and are collected continuously. Creating a timer is a much lighter operation then creating an observation.

Both approaches can be used either together or separately. I see no point in replacing one approach with the other, as they serve different purposes.

fractal3000: #### Micrometer Observation vs. Manual Timer

The argument for using both `Observation` and `Timer` is conceptually and technically flawed for a modern observability stack. Here is the breakdown:

1. **The Power of Correlation (TraceId):** Unlike a standalone `Timer`, an `Observation` acts as a "golden thread" linking the three pillars of observability. It automatically injects a unified **TraceId** into your **Metrics** (via Exemplars), **Tracing** (spans), and **Logs** (via MDC). This allows for seamless navigation: you can click on a latency spike in Prometheus and jump directly to the specific trace and corresponding log entries. Using a manual `Timer` breaks this chain, leaving your metrics isolated and making root-cause analysis significantly slower.
2. **Integrated Lifecycle (Single Point of Truth):** Micrometer Observation is a unified facade. Using a manual `Timer` alongside an `Observation` creates redundant instrumentation, violating the [**"Instrument Once"**](https://docs.micrometer.io) principle and increasing maintenance debt.
3. **Incorrect Sampling Assumption:** Metrics in Micrometer are **not affected by trace sampling**. The `DefaultMeterObservationHandler` records data for **100% of events**, even if the trace is sampled out. This ensures reliable statistics without the overhead of full tracing ([Documentation Reference](https://docs.micrometer.io#usage-examples)).
4. **Operational Risks of Duplication:** Simultaneous usage leads to **Metric Cardinality Explosion** (doubling storage costs) and **Data Inconsistency**. Divergent tags between a manual `Timer` and an `Observation` make cross-system correlation nearly impossible.
5. **Legacy Status of Direct Meter API:** In the Spring Boot 3 ecosystem, direct `Timer` usage is considered a **legacy approach**. Modern frameworks [automatically disable internal timers](https://docs.spring.io) when `Observation` is active to prevent data duplication.

**Conclusion:** For any product pursuing "Full Observability," direct `Timer` usage is obsolete. `Observation` provides the essential correlation (TraceId) that a standalone `Timer` fundamentally cannot offer.

jmix-core/core/src/main/java/io/jmix/core/impl/UnconstrainedDataManagerImpl.java
KremnevDmitry: This might be a good idea. We can implement this in future iterations. For the initial implementation, I think using a data manager will be sufficient.

jmix-core/core/src/main/java/io/jmix/core/impl/UnconstrainedDataManagerImpl.java
```diff
@@ -267,21 +275,30 @@ protected TransactionDefinition createTransactionDefinition(boolean isJoinTransa
...
+        return observationSupport.createEntitiesSaveObservation(context, store)
+                .observe(() -> dataStore.save(context));
     }
 
     @Override
     public List<KeyValueEntity> loadValues(ValueLoadContext context) {
         DataStore store = dataStoreFactory.get(getStoreName(context.getStoreName()));
         context.setAccessConstraints(mergeConstraints(context.getAccessConstraints()));
-        return store.loadValues(context);
+
```
Gavrilov-Ivan, line 287: Extra line

jmix-core/core/src/main/java/io/jmix/core/observation/DataObservationSupport.java
```diff
@@ -0,0 +1,172 @@
...
+    public DataObservationSupport(CoreProperties coreProperties) {
+        this.observationEnabled = coreProperties.isDataObservationEnabled();
+    }
+
+    public Observation createEntityLoadObservation(MetaClass metaClass, @Nullable LoadContext.Query query) {
+        if (!observationEnabled) {
+            return Observation.NOOP;
+        }
+
+        Observation observation = Observation.createNotStarted(DATA_BASE_LOAD_NAME, observationRegistry)
```
Gavrilov-Ivan, line 66: `observationRegistry` is optional.
`Observation.createNotStarted` is called with `null` `observationRegistry` leads to `NOOP` Observation.
Maybe introduce method like `isObservationEnabled` which checks both Jmix property and existance of `observationRegistry` and call it at the begining?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -0,0 +1,153 @@
...
+                .lowCardinalityKeyValue("fragment.class", observationInfo.fragmentClass());
+
+        if (!Strings.isNullOrEmpty(observationInfo.fragmentId())) {
+            observation.lowCardinalityKeyValue("fragment.id", observationInfo.fragmentId());
+        }
+
+        return observation;
+    }
+
+    public Observation createActionExeutionObservation(Action action) {
```
Gavrilov-Ivan: Typo

glebfox (approved):

PR #5072, opened 2026-02-16
Test refactoring

knstvk:
1. Added a new internal module `jmix-core/framework-test-support` with shared Spring configuration classes. The `*TestConfiguration` classes import these shared configurations instead of duplicating code. 

2. Fixed the order of configuration classes in `@SpringBootTest` in some test classes: `DialogTest`, `BackwardNavigationUiTest`, `NotificationTest`.

After merging this PR, we need to refactor premium add-ons to use `framework-test-support`.


jmix-core/framework-test-support/src/main/java/io/jmix/testsupport/config/CommonCoreTestBeans.java
```diff
@@ -0,0 +1,47 @@
...
+import io.jmix.core.cluster.LocalApplicationEventChannelSupplier;
+import io.jmix.core.impl.JmixMessageSource;
+import org.springframework.cache.CacheManager;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.MessageSource;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CommonCoreTestBeans {
```
Gavrilov-Ivan: Maybe add `Configuration` suffix to all `*Beans` configuration classes (or replace `Beans` with `Configuration`)?

jmix-core/framework-test-support/src/main/java/io/jmix/testsupport/config/LiquibaseTestConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.sql.DataSource;
+
+@Configuration
+public class LiquibaseTestBeans {
+
+    @Bean("liquibase")
+    public SpringLiquibase liquibase(DataSource dataSource) {
+        SpringLiquibase liquibase = new SpringLiquibase();
+        liquibase.setDataSource(dataSource);
+        liquibase.setChangeLog("test_support/liquibase/test-changelog.xml");
```
Gavrilov-Ivan, line 32: `Search` modules have different name of the root changelog.xml. Do they keep their own configurations intact without refactoring?

jmix-core/framework-test-support/src/main/java/io/jmix/testsupport/config/CommonCoreTestBeans.java
knstvk: Renamed to `*Configuration`

jmix-core/framework-test-support/src/main/java/io/jmix/testsupport/config/LiquibaseTestConfiguration.java
knstvk: Applied the new test configurations to the Search modules too.

Issue #4842, opened 2025-10-27
Login to UI via OIDC doesn't provide info about security scope


Gavrilov-Ivan:
### Environment

Jmix version: 2.7

### Bug Description
Login to UI via OIDC doesn't provide info about security scope, so both scopes (UI and API) are considered as available and all roles are applied.

See main description of https://github.com/jmix-framework/jmix/issues/4425 (`Case 1`)

Also other external IdP should be taken into acount.

Issue #4425, opened 2025-05-15
Resource role scopes don't work if we use OpenID authorization


fractal3000:
### Environment

Jmix version: 1.6.2

### Bug Description

If we use OpenID authorization the org.springframework.security.core.Authentication object (that can be accessed via io.jmix.core.security.CurrentAuthentication) doesn't contain an io.jmix.core.security.ClientDetails object(that can be accessed via org.springframework.security.core.Authentication#getDetails) but org.springframework.security.web.authentication.WebAuthenticationDetails.

The org.springframework.security.web.authentication.WebAuthenticationDetails doesn't contain any information about the scopes(in meaning Jmix scopes io.jmix.core.security.ClientDetails#getScope).

As a consecuence of the absence information about the scope the security engine stops taking into account the resource roles scopes and take all the resource roles as applicable.

So that means if we use OpenID the dividing resource roles by scope doesn't work.

### Steps To Reproduce

1. Create project with an entity with name TestEntity
2. Create resource role with "API" scope for the TestEntity
3. Create a user with the created resource role
4. Configure OpenID authorization for the project
5. Login with the created user via UI

### Current Behavior

The user can get access to the TestEntity

### Expected Behavior

The user can't get access to the TestEntity

glebfox:
For future assignee: make a port issue

Gavrilov-Ivan:
It's actual for Authorization Server too.

PR #5094, opened 2026-02-24
Security scopes fix

Gavrilov-Ivan:
See #4842 (and also #4425).

Provides an ability to create Security filter chain customizers which can apply some modifications to specific chain(s) (if exists).
Each module is responsible for its own necessary modifications.

Customizers have been added to AuthServer and OIDC modules to set proper `AuthenticationDetailsSource`

### AuthServer
Module provides `AuthServerResourceServerSecurityFilterChainCustomizer` that modifies `BearerTokenAuthenticationFilter` and force API scope.

#### Properties:

- `jmix.authserver.filter-chain.api-scope-security-filter-chain-names` - list of chains to modify (default value `authsr_ResourceServerSecurityFilterChain`, defines in `module.properties`)
- `jmix.authserver.filter-chain.force-api-scope-enabled` - feature toggle (default value `true`)

### OIDC
Module provides `OidcResourceServerSecurityFilterChainCustomizer` that modifies `BearerTokenAuthenticationFilter` and force API scope.
Also provides `OidcVaadinSecurityFilterChainCustomizer` that modifies `OAuth2LoginAuthenticationFilter` and force UI scope

#### Properties:

- `jmix.oidc.filter-chain.api-scope-security-filter-chain-names` - chains to modify for API scope (default value `oidc_JwtSecurityFilterChain`, defines in `module.properties`)
- `jmix.oidc.filter-chain.ui-scope-security-filter-chain-names` - chains to modify for UI scope (default value `VaadinSecurityFilterChainBean`, defines in `module.properties`)
- `jmix.oidc.filter-chain.force-api-scope-enabled` - feature toggle for API scope (default value `true`)
- `jmix.oidc.filter-chain.force-ui-scope-enabled` - feature toggle for UI scope (default value `true`)

Resource Server only properties (`jmix.resource-server.force-api-security-scope.enabled`, `jmix.resource-server.force-api-security-scope.security-filter-chain-names`) has been removed.

### DIFF
--- a/jmix-authserver/authserver/src/main/java/io/jmix/authserver/filter/AuthServerResourceServerSecurityFilterChainCustomizer.java
+++ b/jmix-authserver/authserver/src/main/java/io/jmix/authserver/filter/AuthServerResourceServerSecurityFilterChainCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Haulmont.
+ * Copyright 2026 Haulmont.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/jmix-oidc/oidc/src/main/java/io/jmix/oidc/filter/OidcResourceServerSecurityFilterChainCustomizer.java
+++ b/jmix-oidc/oidc/src/main/java/io/jmix/oidc/filter/OidcResourceServerSecurityFilterChainCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Haulmont.
+ * Copyright 2026 Haulmont.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/jmix-oidc/oidc/src/main/java/io/jmix/oidc/filter/OidcVaadinSecurityFilterChainCustomizer.java
+++ b/jmix-oidc/oidc/src/main/java/io/jmix/oidc/filter/OidcVaadinSecurityFilterChainCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Haulmont.
+ * Copyright 2026 Haulmont.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/jmix-security/security-resource-server/src/main/java/io/jmix/securityresourceserver/authentication/BaseResourceServerSecurityFilterChainCustomizer.java
+++ b/jmix-security/security-resource-server/src/main/java/io/jmix/securityresourceserver/authentication/BaseResourceServerSecurityFilterChainCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Haulmont.
+ * Copyright 2026 Haulmont.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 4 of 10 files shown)


jmix-authserver/authserver/src/main/java/io/jmix/authserver/filter/AuthServerResourceServerSecurityFilterChainCustomizer.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2025 Haulmont.
```
knstvk: 2026?

jmix-authserver/authserver/src/main/java/io/jmix/authserver/AuthServerProperties.java
```diff
@@ -206,4 +218,39 @@ public List<String> getRowLevelRoles() {
...
+         * @see #forceApiScopeEnabled
+         */
+        public boolean isForceApiScopeEnabled() {
+            return forceApiScopeEnabled;
+        }
+
+        /**
+         * @see #apiScopeSecurityFilterChainNames
+         */
+        public List<String> getApiScopeSecurityFilterChainNames() {
```
knstvk, line 252: Why do we need this property?
How do you expect it to be defined in a project?

If this is really needed, maybe use `jmixModules.getPropertyValues()` to enable additive values collected from all add-ons?  

The same question is for OIDC property.

Gavrilov-Ivan: #### Why
Default value of this property (and the similar one from OIDC) relates to the name of the chain from _default_ resource server configuration (within both AuthServer and OIDC).
We already provide an option to disable those default configurations and implement the custom one on the project side. They can have different names - so we need the way to tell what chain should be customized.

The UI one from OIDC relates to the name of the chain defined by Vaadin itself (VaadinSecurityFilterChainBean). I'm not sure if this specific chain will be overridden with different name, but I want to provide the option to provide the actual name.

#### Collecting values
Something similar was before then it works only for API scope within Resource Server. And the Resource Server was responsible for that management.

Now specific customizer belongs to specific module. And each of them should be applied to specific chain only the module is responsible for.

So in the best case collecting values from all modules will lead to repeatable modification of the same chain. In the worst - something will break.

jmix-authserver/authserver/src/main/java/io/jmix/authserver/filter/AuthServerResourceServerSecurityFilterChainCustomizer.java
Gavrilov-Ivan: Updated for new files

jmix-authserver/authserver/src/main/java/io/jmix/authserver/AuthServerProperties.java
Gavrilov-Ivan: Additional clarification about collecting chain names for resource server.
Did you suggest to move partial responsibility for chains related to Resource Server back to Resource Server module?

- Property for chain names is provided by Resource Server module.
- `getChainBeanNames` implementation is moved to `BaseResourceServerSecurityFilterChainCustomizer` (in Resource Server module).
- Values are collected from all modules with this property.
- Specific customizer beans in AuthServer and OIDC are removed (`BaseResourceServerSecurityFilterChainCustomizer` is the final one).
- AuthServer and OIDC only provides default chain names in their module.properties
- Feature toggle property also go back to Resource Server.

So basically restore the functionality of `ForceApiSecurityScopePropertiesProvider`?

OIDC + UI case is intact.

knstvk: I'm not sure. Will it be simpler for understanding and configuring (say for 90% usages)? 
Also, what about the new SAML add-on and its usage in REST?

knstvk: As we have module responsibility for configuring Vaadin (in OIDC and perhaps SAML), looks like it's more logical to keep also their own responsibility for configuring Resource Server.

Issue #3389, opened 2024-06-13
Support in-memory sorting for DataGrid


glebfox:
Origin: [forum](https://forum.cuba-platform.com/t/sorting-generated-column/17447)

**Problem**

If more than one page of data is loaded, in-memory sorting for transient attributes is not applied. This may confuse users because behavior differs for one page and several.

**Proposed solution**

1. Make columns now associated with jpa attributes non-sortable by default.
2. Provide explicit in-memory sorting rule, e.g. using Vaadin Grid API - https://vaadin.com/docs/next/components/grid#specifying-sort-property (`Column:: setComparator`)
3. CollectionLoader should take into account this in-memory sorter

glebfox:
Relates to https://github.com/jmix-framework/jmix/issues/2093

Flaurite:
### QA
Check new feature for `DataGrid`/`TreeDataGrid` and `GroupDataGrid`.

Refer to PR for feature description: https://github.com/jmix-framework/jmix/pull/5141

SergeiAksenov2:
Found: DESC sorting error.

SergeiAksenov2:
Tested on:
Jmix version: 3.0.999-SNAPSHOT
Jmix Studio Plugin Version: 3.0.SNAPSHOT8333-261
IntelliJ version: IntelliJ IDEA 2026.1.2 Preview

PR #5141, opened 2026-03-13
Support in-memory sorting for DataGrid (3389)

Flaurite:
Fixes for: https://github.com/jmix-framework/jmix/issues/3389

### Sort.ExpressionOrder
Added new type of sort order- `ExpressionOrder`.

It is supported by `SortJpqlGenerator`. The generator checks `ExpressionOrder` and place it as is to `sortExpressions`. It means that `ExpressionOrder` skips checks for `MetaPropertyPath` and does not pass to `JpqlSortExpressionProvider`.

`ExpressionOrder` can be used when it is need to sort by specific expression. For instance, for JPQL, the expressions can be:

- `"function('calc_total_sum', {E}.id)"`;
- `"length({E}.firstName)"`;
- `"{E}.firstName"`.

#### ExpressionOrder and KeyValueEntity query
When `ExpressionOrder` is used in query which result is `KeyValueEntity`, the `ExpressionOrder` requires to use the same entity alias as in a query. 

**For instance:**
Query: `select e.name, e.startTime, e.endTime from CarService e`
The expression: `length(e.name)`

----

### Comparator in DataGridColumn
Now, when `DataGridColumn` has custom comparator, it will be used when sorting is `in-memory`.
Note that `in-memory` sorting occurs:
- when loader is not set to CollecitionContainer
- when loaded first page (e.g. due to `SimplePagination`) and number of loaded items is less than page size.

Example of adding comparator:
```
DataGridColumn<Customer> fullNameColumn = dataGrid.getColumnByKey("fullName");
if (fullNameColumn != null) {
    fullNameColumn.setComparator(
            Comparator.comparing(Customer::getFirstName)
                    .thenComparing(Customer::getLastName));
}
```

----

### Sort builder delegate in DataGrid
Added the extension point to customize sort in `DataGrid`. For instance:

```
dataGrid.setSortBuilderDelegate(context ->
        DataGridSortBuilder.create(context)
                .replaceSort("firstName", "length({E}.firstName)")
                .build());
```
When this feature may be used:
 1) You need to customize persistent sorting or in-memory sorting when the user sorts by a column bound with entity property;
 2) You have a column that is not bound with entity property, but there is a way to sort it in-memory and in persistent sorting.
 3) You need to sort transient entity property in persistent sorting.
 
 #### Sample 1. Override sorting for JPA entity property
 Let's imagine that some entity property with String type stores integer values.
 
 By default, this property is sorted as String values:
 - Original order: "5", "56", "1", "12", "43"
-  Sort asc: "1", "12", "43", "5", "56"
-  Sort: desc: "56", "5", "43", "12", "1"

Let's make sorting to sort by number values as follows:
 - Original order: "5", "56", "1", "12", "43"
-  Sort asc: "1", "5", "12", "43", "56"
-  Sort: desc: "56", "43", "12", "5", "1"

```
dataGrid.setSortBuilderDelegate(context ->
        DataGridSortBuilder.create(context)
                .replaceSort("loyaltyPointsStr", "cast({E}.loyaltyPointsStr integer)",
                        (o1, o2) -> {
                            int calc1 = Integer.parseInt(o1.getLoyaltyPointsStr());
                            int calc2 = Integer.parseInt(o2.getLoyaltyPointsStr());
                            return Integer.compare(calc1, calc2);
                        })
                .build());
```
We have provided sorting for `in-memory` and for persistent (JPQL).

#### Sample 2. Add sorting for column without property.
Let's say we have a column "Full name". It renders concatenated `firstName` and `lastName` properties.

```
DataGridColumn<Customer> fullNameColumn = dataGrid.getColumnByKey("fullName");
if (fullNameColumn != null) {
    fullNameColumn.setRenderer(new TextRenderer<>(item -> item.getFirstName() + " " + item.getLastName()));
}
```

In the sort builder delegate we can do the following:
```
dataGrid.setSortBuilderDelegate(context ->
        DataGridSortBuilder.create(context)
                .replaceSort("fullName", List.of("{E}.firstName", "{E}.lastName"),
                        Comparator.comparing(Customer::getFirstName).thenComparing(Customer::getLastName))
                .build());
```

### Sort Builder Delegate vs CollectionContainerSorter and JpqlSortExpressionProvider
Use **Sort Builder Delegate** when sorting must be customized for a specific `DataGrid`:
- one place to define both persistent sort (JPQL expressions) and in-memory sort (Comparator);
- convenient for columns without bound entity property, transient/computed fields, or per-screen rules;
- best choice for local UI behavior without global side effects.

Use `CollectionContainerSorter` when you need to customize container-level **in-memory** sorting mechanics:
- affects sorting of the whole `CollectionContainer` (not only one grid column mapping);
- useful for custom comparator creation logic in sorter internals;
- if this behavior should be global/system-wide, override `SorterFactory` (`@Primary` custom factory).

Use custom `JpqlSortExpressionProvider` when you need global customization of **DB-level** ORDER BY expression generation for entity properties:
- applies to persistent sorting for entity property-based orders across loaders/views;
- good for DB-specific functions/casts/null handling as a default policy;
- does not handle in-memory sorting and is bypassed when using `Sort.ExpressionOrder` (expression is inserted as-is).

In short:
- view/column-specific + both persistent/in-memory -> **Sort Builder Delegate**
- global/container in-memory sorting strategy -> `CollectionContainerSorter` (+ `SorterFactory` for global rollout)
- global JPQL expression policy for property sorting -> `JpqlSortExpressionProvider`.


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/InMemorySortInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Represents sorting information for in-memory sorting operations.
+ */
+public interface InMemorySortInfo extends SortInfo {
+
+    /**
+     * @return comparator to be used for in-memory sorting, or {@code null} if not set
+     */
+    @Nullable Comparator<?> getComparator();
```
glebfox: Let's put `@Nullable` above field as we always do

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/SortInfo.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return the MetaPropertyPath object representing the property path, or {@code null} if not set
+     */
+    @Nullable MetaPropertyPath getMetaPropertyPath();
+
+    /**
+     * @return the property name to sort by
+     */
+    String getProperty();
```
glebfox: Unclear why for property path we use we use `MetaPropertyPath`, but for property - string isntead of `MetaProperty`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/InMemorySortInfoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class InMemorySortInfoImpl implements InMemorySortInfo {
+
+    protected final @Nullable MetaPropertyPath metaPropertyPath;
+    protected final String property;
+    protected final boolean ascending;
+
+    protected @Nullable Comparator<?> comparator;
+
+    public InMemorySortInfoImpl(@Nullable MetaPropertyPath metaPropertyPath,
```
glebfox: looks odd having a nullable parameter at the first place

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/InMemorySortInfoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+package io.jmix.flowui.component.grid.sort.impl;
+
+import io.jmix.core.metamodel.model.MetaPropertyPath;
+import io.jmix.flowui.component.grid.sort.InMemorySortInfo;
+import org.jspecify.annotations.Nullable;
+
+import java.util.Comparator;
+
+public class InMemorySortInfoImpl implements InMemorySortInfo {
```
glebfox, line 32: JavaDoc

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/PersistentSortInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import io.jmix.core.common.util.Preconditions;
+import io.jmix.core.metamodel.model.MetaPropertyPath;
+import io.jmix.flowui.component.grid.sort.PersistentSortInfo;
+import org.jspecify.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class PersistentSortInfoImpl implements PersistentSortInfo {
```
glebfox, line 35: JavaDoc

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfo.java
```diff
@@ -0,0 +1,45 @@
...
+
+package io.jmix.flowui.component.grid;
+
+import io.jmix.core.annotation.Experimental;
+import io.jmix.core.metamodel.model.MetaPropertyPath;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Represents sorting information for a specific column.
+ * @param <E> entity type
```
glebfox, line 26: format the file

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfo.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@Experimental
+public interface ColumnSortInfo<E> {
+
+    /**
+     * @return the column associated with this sort info
```
glebfox: don't be lazy, add description to JavaDoc, not only `@return`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.jmix.core.annotation.Experimental;
+import io.jmix.core.annotation.Internal;
+import io.jmix.core.common.util.Preconditions;
+import io.jmix.core.metamodel.model.MetaPropertyPath;
+import org.jspecify.annotations.Nullable;
+
+@Experimental
+@Internal
+public class ColumnSortInfoImpl<T> implements ColumnSortInfo<T> {
```
glebfox, line 33: !!!

jmix-flowui/flowui/src/main/java/io/jmix/flowui/data/grid/ContainerDataGridItems.java
```diff
@@ -121,19 +122,65 @@ public boolean containsItem(T item) {
 
     @Override
     public void sort(Object[] propertyId, boolean[] ascending) {
+        sortInternal(createSort(propertyId, ascending));
+    }
+
+    @Override
+    public void sort(DataGridSort sort) {
+        if (container.getSorter() instanceof BaseContainerSorter sorter) {
```
glebfox: looks like we can extract a method for this code

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/BaseContainerSorter.java
```diff
@@ -80,6 +92,11 @@ protected void sortInMemory(Sort sort) {
     protected abstract void setItemsToContainer(List<?> list);
 
     protected Comparator<?> createComparator(Sort.Order sortOrder, MetaClass metaClass) {
+        Comparator<?> customComparator = propertyComparators.get(sortOrder.getProperty());
```
glebfox: I think that it's better to check if `propertyComparators` is `null` here rather than init it with an empty then check passed parameter.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/BaseContainerSorter.java
```diff
@@ -52,6 +53,17 @@ public CollectionContainer<?> getContainer() {
...
 
+    /**
+     * Sets property comparators. Comparators are used in in-memory sorting.
+     * <p>
+     * The key of the map is a property name and value is comparator.
+     *
+     * @param propertyComparators the map of property comparators
+     */
+    public void setPropertyComparators(@Nullable Map<String, Comparator<?>> propertyComparators) {
+        this.propertyComparators = propertyComparators != null ? propertyComparators : Collections.emptyMap();
```
glebfox: it'd better to copy passed map to prevent issues if it will be changed in the outside code.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/DataGridSortContext.java
```diff
@@ -0,0 +1,59 @@
...
+    public DataGridSortContext(Grid<E> grid, List<ColumnSortInfo<E>> columnSortInfos) {
+        Preconditions.checkNotNullArgument(grid);
+        Preconditions.checkNotNullArgument(columnSortInfos);
+
+        this.grid = grid;
+        this.columnSortInfos = new ArrayList<>(columnSortInfos);
+    }
+
+    /**
+     * @return the grid
```
glebfox: !!!

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/DataGridSort.java
```diff
@@ -0,0 +1,117 @@
...
+public class DataGridSort {
+
+    private static final Logger log = LoggerFactory.getLogger(DataGridSort.class);
+
+    protected List<PersistentSortInfo> persistentSortInfos;
+    protected List<InMemorySortInfo> inMemorySortInfos;
+
+    protected DataGridSort(List<PersistentSortInfo> persistentSortInfos,
+                           List<InMemorySortInfo> inMemorySortInfos) {
+        this.persistentSortInfos = new ArrayList<>(persistentSortInfos);
```
glebfox: looks like we don't change these lists after initialization. Replace ArrayList with unmodifiableList

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/DataGridSort.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/DataGridSortContext.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/BaseContainerSorter.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/BaseContainerSorter.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/data/grid/ContainerDataGridItems.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfoImpl.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfo.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/ColumnSortInfo.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/PersistentSortInfoImpl.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/InMemorySortInfoImpl.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/impl/InMemorySortInfoImpl.java
Flaurite: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/SortInfo.java
Flaurite: Fixed by renaming to `getSortKey()`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/grid/sort/InMemorySortInfo.java
Flaurite: Fixed

PR #5163, opened 2026-03-23
UnitOfWorkQueryValueHolder is not wrapped properly jmix-framework/jmix_2461

### LINKED ISSUE #2461: UnitOfWorkQueryValueHolder is not wrapped properly
See forum [topic](https://forum.jmix.io/t/onetoone-plan-fetch/3266).
### Steps:
1. Download and run project: [unfetch.zip](https://github.com/jmix-framework/jmix/files/13343070/unfetch.zip)
2. Open Blank Screen
3. Press upper button, then bottom button
**ER:** No exceptions
**AR:**
```java
java.lang.IllegalStateException: Cannot get unfetched attribute [infoEntity] from detached object com.company.jmix_unfetch.entity.LastEntity-92a4bf42-6755-3b2a-c5ba-2badd96853f2 [detached].
	at org.eclipse.persistence.internal.indirection.UnitOfWorkValueHolder.throwUnfetchedAttributeException(UnitOfWorkValueHolder.java:326)
	at org.eclipse.persistence.internal.indirection.DatabaseValueHolder.getValue(DatabaseValueHolder.java:106)
	at com.company.jmix_unfetch.entity.LastEntity._persistence_get_infoEntity(LastEntity.java)
	at com.company.jmix_unfetch.entity.LastEntity.getInfoEntity(LastEntity.java:25)
	at com.company.jmix_unfetch.screen.blank.BlankScreen.onReadClick(BlankScreen.java:49)
	at io.jmix.core.common.event.EventHub.publish(EventHub.java:170)
```
### Cause: 
UnitOfWorkQueryValueHolder is not wrapped for `infoEntity` property in deeply nested entity `LastEntity`, because it is already loaded despite absence in fetch plan.

see `io.jmix.eclipselink.impl.lazyloading.JpaLazyLoadingListener#processValueHolders`

### Possible solutions:
a. Prevent excessive loading
b. Modify wrapping logic
c. leave as is if there are no "good" solutions

dtaimanov:


jmix-core/core/src/main/java/io/jmix/core/entity/LoadedPropertiesInfo.java
```diff
@@ -31,6 +32,8 @@ public interface LoadedPropertiesInfo extends Serializable {
 
     boolean isLoaded(Object entity, String property, PersistentAttributesLoadChecker checker);
 
+    PropertyLoadedState isLoadedSafe(Object entity, String property, PersistentAttributesLoadChecker checker);
```
Gavrilov-Ivan, line 46: Naming. `isXXX` usually tells that it is a boolean outcome. But `isLoadedSafe` returns some enum, not boolean (in contrast to existing `isLoaded`).
But it has similar naming to other non-boolean methods in this context...

Gavrilov-Ivan, line 46: Add javadoc for both `isLoaded` and `isLoadedSafe` to understand the difference

jmix-core/core/src/main/java/io/jmix/core/impl/CorePersistentAttributesLoadChecker.java
```diff
@@ -40,81 +41,90 @@ public class CorePersistentAttributesLoadChecker implements PersistentAttributes
...
-        NO,
-        UNKNOWN
-    }
 
     @Override
     public boolean isLoaded(Object entity, String property) {
+        return YES.equals(isLoadedInternal(entity, property, false));
+    }
+
+    public PropertyLoadedState isLoadedInternal(Object entity, String property, boolean safe) {
```
Gavrilov-Ivan, line 57: Javadoc for 'safe' parameter (here and everywhere else) or rename it.

jmix-data/eclipselink/src/main/java/io/jmix/eclipselink/impl/lazyloading/JpaLazyLoadingListener.java
```diff
@@ -134,6 +145,79 @@ public void processValueHolders(Object entity, LoadContext<?> loadContext, Fetch
...
+                wrapExcessivelyLoadedPropertyValueHolders(excessivelyLoadedProperty.getKey(), property, loadOptions, processedEntities);
+            }
+        }
+    }
+
+    /**
+     * Organizes value holder wrapping for excessively loaded properties according to the property's cardinality.
+     * Skips processing of system entities according to the {@code isSkipExcessiveLoadedProcessing} method.
+     */
+    private void wrapExcessivelyLoadedPropertyValueHolders(Object entity, MetaProperty property, LoadOptions loadOptions, Set<Object> processedEntities) {
```
Gavrilov-Ivan: Make private methods protected

Gavrilov-Ivan (changes requested):

jmix-core/core/src/main/java/io/jmix/core/entity/LoadedPropertiesInfo.java
dtaimanov: As we discussed during our last meetings, it may be useful to keep `isLoadedSafe` because its return type is close to `boolean` and similar to `Boolean` (`YES/NO/UNKNOWN` vs. `TRUE/FALSE/null`), and this name has several benefits:

1. It is similar to and consistent with the old widely used method, making it easier to find through autocomplete and text search.
2. `isLoadedSafe` is shorter than something like `getLoadedStateSafe` or `getPropertyLoadedStateSafe`, which is important because this check may often be used in complex `if` clauses.
 
I would prefer to keep `isLoadedSafe` method name.

PR #5180, opened 2026-03-28
Move processParameters logic from ParameterJpqlGenerator to ConditionGenerator implementations

### LINKED ISSUE #5179: Ability to extend ParameterJpqlGenerator without overriding
Currently `processParameters` method in `io.jmix.data.impl.jpql.generator.ParameterJpqlGenerator` is hardcoded to process `PropertyCondition` and `JpqlCondition` parameters. So when you create a custom `io.jmix.core.querycondition.Condition` implementation, the only way to add new process logic is to override `ParameterJpqlGenerator` and mark it `@Primary` in the app.

It would be better to be able to implement parameters processing in extension beans like `io.jmix.data.impl.jpql.generator.ConditionGenerator` implementations that allow to implement query and parameter value generation logic for each type of condition. 

This might be useful to implement custom condition types in addons.

godlewski-dm:
Fixes #5179

Additionally added `isCaseInsensitiveOperation` convenience method to `PropertyConditionUtils`

### DIFF
--- a/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionGenerator.java
+++ b/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionGenerator.java
@@ -23,8 +23,10 @@ import java.util.Map;
 
 /**
  * Modifies parts of JPQL query
+ *
+ * @param <T> condition type
  */
-public interface ConditionGenerator {
+public interface ConditionGenerator<T extends Condition> {
 
     /**
      * Checks whether the condition generator supports the given {@code context}.
@@ -60,7 +62,7 @@ public interface ConditionGenerator {
      */
     Map<String, Object> processParameters(Map<String, Object> parameters,
                                           Map<String, Object> queryParameters,
-                                          Condition condition,
+                                          T condition,
                                           @Nullable String entityName);
 
     /**
--- a/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/JpqlConditionGenerator.java
+++ b/jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/JpqlConditionGenerator.java
@@ -32,7 +32,7 @@ import java.util.regex.Pattern;
 
 @Component("data_JpqlConditionGenerator")
 @Order(JmixOrder.LOWEST_PRECEDENCE)
-public class JpqlConditionGenerator implements ConditionGenerator {
+public class JpqlConditionGenerator implements ConditionGenerator<JpqlCondition> {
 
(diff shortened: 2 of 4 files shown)


jmix-core/core/src/main/java/io/jmix/core/querycondition/PropertyConditionUtils.java
```diff
@@ -71,6 +71,20 @@ public static boolean isMemberOfCollectionOperation(PropertyCondition propertyCo
                 || PropertyCondition.Operation.NOT_MEMBER_OF_COLLECTION.equals(operation);
     }
 
+    /**
+     * Checks if the property condition's operation is case-insensitive.
+     * @param condition property condition
+     * @return true if the operation is case-insensitive, otherwise - false
+     */
+    public static boolean isCaseInsensitiveOperation(PropertyCondition condition) {
```
KremnevDmitry, line 79: Perhaps the name doesn't quite capture the essence of the method. With the `string` property type, the `equals` operation is also a case-sensitive. Is `isStringBasedOperation` a more appropriate name?

jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionGenerator.java
KremnevDmitry, line 80: It appears this method is no longer part of the public API because specific method implementations are used directly in interface implementations. Let's deprecate this method and mark it as not part of the public API.

`@Deprecated(since = "2.8", forRemoval = true)`

Also note in the javadoc that implementations of the `ConditionGenerator` interface use internal implementations of this method.

KremnevDmitry: It looks like the interface now needs to be typed to the `Condition` type, as does the `processParameters` method.

jmix-core/core/src/main/java/io/jmix/core/querycondition/PropertyConditionUtils.java
godlewski-dm: My purpose here is to indicate exactly case-insensitive operations for property conditions. As I can see case-insensitive marker is applied only for these four operations. Also double checked `equals` operation and it does consider character case.

Image

KremnevDmitry: Okay, accepted

jmix-data/data/src/main/java/io/jmix/data/impl/jpql/generator/ConditionGenerator.java
godlewski-dm: Marked as deprecated

godlewski-dm: Done

KremnevDmitry:
Merged, thanks for your contribution

PR #5200, opened 2026-04-09
Support for application.yml files in the "cleanConf" task

Gavrilov-Ivan:


jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/JmixPlugin.groovy
```diff
@@ -217,17 +217,10 @@ class JmixPlugin implements Plugin<Project> {
...
-                            appPropsFile.withInputStream { mainProperties.load(it) }
-                        } else {
-                            project.logger.lifecycle("File src/main/resources/application.properties not found")
-                        }
-
+                        def mainProperties = loadProperties(project)
                         def confDir = resolveConfDir(project, mainProperties)
 
-                        project.logger.lifecycle("Delete directory: {}", confDir)
+                        project.logger.lifecycle("Delete directory: ${confDir}")
```
dtaimanov: I would prefer the previous version, because this code will do string substitution even if logs are disabled. It may not affect performance in this place, but generally it is considered as a bad practice.

dtaimanov (approved):

jmix-gradle-plugin/src/main/groovy/io/jmix/gradle/JmixPlugin.groovy
Gavrilov-Ivan: During testing I faced an issue with slf4j style in newly added logs (due to nullable value and signature conflict). Switched to variable resolution but did it too massive.
Reverted to slf4j style everywhere besides lines with possible conflict

PR #5225, opened 2026-04-25
Runtime view templates

### LINKED ISSUE #5226: Runtime view templates
Ability to generate standard list and detail views at runtime from templates and entity metadata.

knstvk:
## Summary

Adds UI View Templates, allowing standard list and detail views to be generated at runtime from templates declared on entity metadata instead of being created manually at design time. 
## Usage

- Public API is exposed through `io.jmix.flowui.view.template.ListViewTemplate` and `io.jmix.flowui.view.template.DetailViewTemplate` annotations. The minimal example:
  ```
  @ListViewTemplate(parentMenu = "application")
  @DetailViewTemplate
  public class Customer {
  }
  ```
- Shared annotation attributes:
  - `path`: resource path to a Freemarker XML descriptor template
  - `templateParams`: JSON object with additional template parameters
  - `parentMenu`: parent menu id for the generated menu item
  - `viewId`: generated view id
  - `viewRoute`: generated route path
  - `viewTitle`: generated page title
- `ListViewTemplate` also supports `lookupComponentId`, which defaults to `dataGrid`.
- `DetailViewTemplate` also supports `editedEntityContainerId`, which defaults to `entityDc`.
- Default values:
  - list view id: `<entityName>.list`
  - detail view id: `<entityName>.detail`
  - list view title: `<entityName> list`
  - detail view title: `<entityName>`
- Routing rules:
  - list `viewRoute` is used as-is
  - detail `viewRoute` is treated as a base route and always becomes `.../:id`
  - if `viewRoute` is empty, it is derived from `viewId`
- `templateParams` is parsed as JSON and merged into the Freemarker model.
- Built-in templates additionally support `includeProperties` and `excludeProperties` in `templateParams`.
- Built-in property filtering supports only direct single-value datatype, enum, association, and composition properties.
- Built-in property filtering does not support collection-valued properties, embedded properties, or dotted property paths such as `customer.name`.
- Built-in templates exclude system, secret, id/version/generated, and audit/soft-delete properties by default.
- `includeProperties` can restore excluded supported direct properties; `excludeProperties` is applied last and wins.
- In the following example, the list view route is `customers` and the detail view route is `customers/:id`:

  ```java
  @ListViewTemplate(
          path = "io/jmix/flowui/view/template/list-view.ftl",
          viewId = "demo_Customer.list",
          viewRoute = "customers",
          viewTitle = "Customers",
          parentMenu = "application",
          templateParams = """
                  {"includeProperties":["createdBy","createdDate"],"excludeProperties":["secretToken"]}
                  """
  )
  @DetailViewTemplate(
          path = "io/jmix/flowui/view/template/detail-template.ftl",
          viewId = "demo_Customer.detail",
          viewRoute = "customers",
          viewTitle = "Customer",
          editedEntityContainerId = "entityDc"
  )
  public class Customer {
  }
  ```

- If `parentMenu` is empty, no menu item is created.
- If `parentMenu` points to a missing menu item, a new root menu item is created automatically.

## Known issues

- To fix before release:
  - Entity properties are sorted randomly. This issue will be addressed in a separate PR.

- To consider later based on user feedback:
  - `ComponentXmlFactory` doesn't respect `jmix.ui.component.entity-field-*` properties. Maybe introduce also entity annotations like CUBA's [@Lookup](https://doc.cuba-platform.com/manual-latest/entity_attr_annotations.html#lookup_annotation).
  - Composition collections are not supported. We could provide a detail view template with a TabSheet like in EntityInspector.
  - Embedded attributes are not supported.

## Implementation details

- `ViewTemplateDefinitions` scans entity metadata, reads template annotations, resolves defaults, validates routes, parses `templateParams`, renders the Freemarker descriptor, and caches the resulting definitions.
- `ViewTemplateDescriptorRegistry` stores generated XML descriptors in memory under synthetic `view-template:` paths.
- `ViewXmlLoader` recognizes those synthetic paths and loads the descriptor from the registry instead of the classpath.
- `ViewTemplateHelper` and `DefaultViewTemplateHelper` provide the built-in property filtering logic used by templates.
- `ComponentXmlFactory` produces parseable XML fragments for supported property types and default field/action mappings.
- `TemplateListView` and `TemplateDetailView` remain generic base classes.
- `ViewTemplateControllerClassFactory` uses ByteBuddy to generate a dedicated controller subclass for each template view, adds runtime `@ViewController`, `@ViewDescriptor`, and `@Route` annotations, and overrides base-class accessors for `lookupComponentId` or `editedEntityContainerId`.
- `ViewRegistry` registers generated views as normal `ViewInfo` entries and also as primary list/detail views for the entity.
- `MenuConfig` appends template-generated menu items after loading standard XML menu definitions.
- Main coverage is in `ViewTemplateIntegrationTest`, `ComponentXmlFactoryTest`, and `DefaultViewTemplateHelperTest`.


jmix-flowui/feature.md
```diff
@@ -0,0 +1,166 @@
+# View Templates Feature Context
```
glebfox: Remove before merge?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/template/impl/DefaultViewTemplateHelper.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Default {@link ViewTemplateHelper} implementation used by built-in view templates.
+ *
+ * @see #getProperties(MetaClass, List, List)
+ */
+@Component("flowui_DefaultViewTemplateHelper")
+public class DefaultViewTemplateHelper implements ViewTemplateHelper {
```
glebfox, line 48: We agreed that utility beans should be called `FooSupport`

glebfox (changes requested):

jmix-flowui/feature.md
knstvk: Yes, I'll remove it right before merge.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/template/impl/DefaultViewTemplateHelper.java
knstvk: This bean is called Helper intentionally because it's used by the templates, not by the application or framework code. If we need a `ViewTemplateSupport` for some internal work in the future, it will be a different bean.

glebfox (approved):

PR #5261, opened 2026-05-07
Application state information 4657

### LINKED ISSUE #4657: Application state information
See Studio issue:
https://youtrack.jmix.io/issue/JST-5606

Description:
Now Studio cannot correct determine the actual state of the application (is application running now or not, on which port, etc.). 
This is affects hot deploy functionality and some actions like start/stop application, open url, etc

Possible solution:
When application started we can create json file with some information about current application state and place this file in conf directory.
Also, in shutdown hook or spring context destroy listener we can update/delete this file.

Gavrilov-Ivan:
See #4657

- Added compile only dependency `org.springframework.boot:spring-boot-web-server` to be able to get effective port number in case of randomly assigned one.
- Consider to also add logging to output application info to console on startup instead of logging in project side in main application class (created from template) which is not accurate rignt now.

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
@@ -20,6 +20,7 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
+import io.jmix.core.annotation.Internal;
 import jakarta.annotation.PostConstruct;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -42,6 +43,7 @@ import java.util.List;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
+@Internal
 @Component("core_ApplicationInfoProvider")
 @ConditionalOnClass(WebServerApplicationContext.class)
 public class ApplicationInfoProvider {
--- a/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
@@ -67,7 +67,9 @@ public class ApplicationInfoProvider {
     @EventListener
     public void onApplicationReady(ApplicationReadyEvent event) {
         initValues(event.getApplicationContext());
-        createAppInfoFile();
+        if (coreProperties.isApplicationInfoFileEnabled()) {
+            createAppInfoFile();
+        }
         log.debug("ApplicationInfoProvider initialization is completed");
     }
 
--- a/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright 2026 Haulmont.
+ *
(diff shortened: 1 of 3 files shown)


jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
```diff
@@ -0,0 +1,237 @@
...
+            );
+
+            log.info("Application info stored to file: {}", fileLocation.toAbsolutePath());
+        } catch (IOException e) {
+            log.warn("Failed to store application info to file: {}", fileLocation.toAbsolutePath(), e);
+        }
+    }
+
+    protected Path resolveAppInfoFileLocation() {
+        String confDir = coreProperties.getConfDir();
```
knstvk, line 208: Are you using conf dir intentionally, to be cleaned on startup?

jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
```diff
@@ -0,0 +1,237 @@
...
+        general.add("activeProfiles", createActiveProfilesJsonArray());
+        general.addProperty("startupTime", startupTime != null ? startupTime.toString() : null);
+
+        JsonObject root = new JsonObject();
+        root.add("general", general);
+
+        return root;
+    }
+
+    protected void writeFile(Path fileLocation, JsonObject json) {
```
knstvk, line 182: As far as I understand, in the current implementation the file will be written even in production when using an embedded web server. I think we need an application property to disable writing the file. For example `jmix.application-info-file.enabled`

knstvk (changes requested):

jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
Gavrilov-Ivan: Added `jmix.core.application-info-file-enabled` property to fit current core properties structure. `true` by default.

jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
Gavrilov-Ivan: File is stored to `<conf>/app-info/app-info.json` according to request from Studio.
This file is explicitly removed on shutdown via shutdown hook - `cleanConf` task can be disabled.

jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
```diff
@@ -0,0 +1,239 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.time.OffsetDateTime;
+import java.util.List;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+@Component("core_ApplicationInfoProvider")
+@ConditionalOnClass(WebServerApplicationContext.class)
+public class ApplicationInfoProvider {
```
knstvk, line 49: If it's a public class, add some Javadocs. Or maybe move it to `impl` or make internal otherwise? Does it make sense to make it a public API?

knstvk (changes requested):

jmix-core/core/src/main/java/io/jmix/core/ApplicationInfoProvider.java
Gavrilov-Ivan: Made it internal.
Don't see it as a public service for now.

knstvk (approved):

PR #5274, opened 2026-05-12
Support multitenancy in application settings

### LINKED ISSUE #5144: Support multitenancy in application settings
### Environment

Jmix version: 2.7.6

### Bug Description

Unable to save application settings for a tenant if they are previously saved for other tenant.

### Steps To Reproduce

1. Create a Jmix project
2. Add the Application settings and Multitenanct add-ons
3. Add an entity for the custom application settings, e.g. `CustomerSettings`
4. Add a tenant field to the custom application settings entity
5. Login as a user with one tenant -> Change some property for the created application setting 
6. Login as a user with one tenant -> Change the same property for the created application setting  and try to save

### Current Behavior

An exception occurs:
Image

### Expected Behavior

Application settings are saved without errors.

### Sample Project

[appsettingsmultitenancy.zip](https://github.com/user-attachments/files/26015414/appsettingsmultitenancy.zip)

KremnevDmitry:
See: https://github.com/jmix-framework/jmix/issues/5144#issuecomment-4430578570

### DIFF
--- a/jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
+++ b/jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
@@ -406,7 +406,7 @@ public class AppSettingsEntityView extends StandardView {
     protected InstanceContainer initInstanceContainerWithDbEntity() {
         InstanceContainer container = dataComponents.createInstanceContainer(currentMetaClass.getJavaClass());
         entityToEdit = appSettingsTools.loadAppSettingsEntityFromDataStore(currentMetaClass.getJavaClass(),
-                AppSettingsEntityLoadMode.FOR_READ, false);
+                AppSettingsEntityLoadMode.FOR_READ);
         entityToEdit = dataContext.merge(entityToEdit);
         return container;
     }
--- a/jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
+++ b/jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
@@ -28,6 +28,7 @@ import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
 import com.vaadin.flow.component.orderedlayout.Scroller;
 import com.vaadin.flow.router.*;
 import io.jmix.appsettings.AppSettings;
+import io.jmix.appsettings.AppSettingsEntityLoadMode;
 import io.jmix.appsettings.AppSettingsTools;
 import io.jmix.appsettings.entity.AppSettingsEntity;
 import io.jmix.appsettingsflowui.AppSettingsUiProperties;
@@ -36,7 +37,6 @@ import io.jmix.core.metamodel.model.MetaClass;
 import io.jmix.core.metamodel.model.MetaProperty;
 import io.jmix.core.metamodel.model.MetadataObject;
 import io.jmix.core.metamodel.model.Range;
-import io.jmix.data.PersistenceHints;
 import io.jmix.flowui.Notifications;
 import io.jmix.flowui.UiComponents;
 import io.jmix.flowui.accesscontext.UiEntityAttributeContext;
@@ -55,7 +55,6 @@ import io.jmix.flowui.util.OperationResult;
 import io.jmix.flowui.util.UnknownOperationResult;
 import io.jmix.flowui.view.*;
 import io.jmix.flowui.view.navigation.RouteSupport;
-import org.jspecify.annotations.NonNull;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import jakarta.persistence.Convert;
@@ -71,7 +70,6 @@ import java.util.stream.Collectors;
 public class AppSettingsEntityView extends StandardView {
 
     protected static final int MAX_CAPTION_LENGTH = 50;
-    protected static final String SELECT_APP_SETTINGS_ENTITY_QUERY = "select e from %s e where e.id = 1";
     protected static final String QUERY_PARAM_SETTINGS_CLASS = "settings";
     protected static final int FORM_COLUMNS = 3;
 
@@ -91,8 +89,6 @@ public class AppSettingsEntityView extends StandardView {
     @Autowired
     protected EntityStates entityStates;
     @Autowired
-    protected UnconstrainedDataManager dataManager;
-    @Autowired
     protected MetadataTools metadataTools;
     @Autowired
     protected AccessManager accessManager;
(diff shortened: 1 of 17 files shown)


jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
```diff
@@ -411,30 +405,12 @@ protected String getPropertyCaption(MetaClass metaClass, MetaProperty metaProper
...
-                .fetchPlan(createFetchPlan())
-                .hint(PersistenceHints.SOFT_DELETION, false)
-                .optional()
-                .orElse(null);
-        if (entityToEdit == null) {
-            entityToEdit = dataContext.create(currentMetaClass.getJavaClass());
-        } else {
-            entityToEdit = dataContext.merge(entityToEdit);
-        }
+        entityToEdit = appSettingsTools.loadAppSettingsEntityFromDataStore(currentMetaClass.getJavaClass(),
```
Gavrilov-Ivan, line 408: Is it correct to use `FOR_READ` mode to load entityTo**Edit**?

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/AppSettingsTenantProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.jmix.appsettings;
+
+/**
+ * Provides the current tenant id for app settings support.
+ */
+public interface AppSettingsTenantProvider {
```
Gavrilov-Ivan, line 22: Why do we need separate service to get current tenant? What about existing `TenantProvider`?

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/impl/AppSettingsToolsImpl.java
```diff
@@ -135,6 +179,85 @@ public <T extends AppSettingsEntity> List<String> getPropertyNames(Class<T> claz
         return metadata.getClass(clazz).getProperties().stream()
                 .filter(metaProperty -> !metadataTools.isSystem(metaProperty))
                 .map(MetadataObject::getName)
+                .filter(name -> !TENANT_ID_PROPERTY.equals(name))
                 .collect(Collectors.toList());
     }
+
+    @Nullable
+    protected <T extends AppSettingsEntity> T loadCurrentTenantAppSettingsEntity(Class<T> clazz,
```
Gavrilov-Ivan: `loadCurrentTenantAppSettingsEntity` accepts tenantId as parameter. So it's not current tenant - it's some tenant.
Change naming or set of parameters.

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/impl/AppSettingsToolsImpl.java
```diff
@@ -135,6 +179,85 @@ public <T extends AppSettingsEntity> List<String> getPropertyNames(Class<T> claz
...
+                                                                                 boolean softDeletion) {
+        List<T> entities = getDataManagerForAppSettingsEntity().load(clazz)
+                .condition(PropertyCondition.equal(TENANT_ID_PROPERTY, tenantId))
+                .hint(PersistenceHints.SOFT_DELETION, softDeletion)
+                .fetchPlan(createFetchPlan(clazz))
+                .list();
+
+        if (entities.size() > 1) {
+            throw new IllegalStateException(String.format(
+                    "More than one tenant settings record exists for entity %s",
```
Gavrilov-Ivan: Is it a user-visible exception? If not maybe also add tenant id to the message?

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/impl/AppSettingsToolsImpl.java
```diff
@@ -135,6 +179,85 @@ public <T extends AppSettingsEntity> List<String> getPropertyNames(Class<T> claz
...
+
+    @Nullable
+    protected <T extends AppSettingsEntity> T loadGlobalAppSettingsEntity(Class<T> clazz, boolean softDeletion) {
+        MetaClass metaClass = metadata.getClass(clazz);
+        String storeName = metaClass.getStore().getName();
+
+        TransactionTemplate transactionTemplate = storeAwareLocator.getTransactionTemplate(storeName);
+        return transactionTemplate.execute(status -> {
+            EntityManager entityManager = storeAwareLocator.getEntityManager(storeName);
+            boolean softDeletionBefore = PersistenceHints.isSoftDeletion(entityManager);
```
Gavrilov-Ivan: Minor: use similar naming style for backup variables.
Personal preference: don't use "paired" labels like  'before' and 'previous' because it implies having values 'after' (after what?) and 'next' (sort of sequence?) which can be confusing in this specific context.

"Unary" label like 'backup', 'origin' or something like this should be more clear for backup-restore case.

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/impl/AppSettingsToolsImpl.java
```diff
@@ -135,6 +179,85 @@ public <T extends AppSettingsEntity> List<String> getPropertyNames(Class<T> claz
...
+        return metadata.create(clazz, LEGACY_GLOBAL_ENTITY_ID);
+    }
+
+    protected <T extends AppSettingsEntity> T createTenantAppSettingsEntity(Class<T> clazz, String tenantId) {
+        T entity = metadata.create(clazz, Math.toIntExact(generateTenantEntityId(clazz)));
+        entity.setTenantId(tenantId);
+        return entity;
+    }
+
+    protected <T extends AppSettingsEntity> long generateTenantEntityId(Class<T> clazz) {
```
Gavrilov-Ivan: Maybe add some javadoc or more detailed name to make functionality of this method more clear?

Gavrilov-Ivan (changes requested):

jmix-appsettings/appsettings-flowui/src/main/java/io/jmix/appsettingsflowui/view/appsettings/AppSettingsEntityView.java
KremnevDmitry: Yes, because this entity is used for display and editing on the view. After that it is saved through the AppSettings service, where it is reloaded with the `FOR_SAVE` flag.

jmix-appsettings/appsettings/src/main/java/io/jmix/appsettings/AppSettingsTenantProvider.java
KremnevDmitry: We need our own `AppSettingsTenantProvider` as a local SPI, thus reducing coupling and dependency on `multitenancy` add-on. This means we only have a compile-time dependency in `appsettings-starter`.

We could use the original `TenantProvider` from the `multitenancy`, but that would increase coupling between modules. Is this what we want to do?

Gavrilov-Ivan (approved):

Issue #5284, opened 2026-05-14
LocalFileStorage: add extra path validation (port to 1.7)


Gavrilov-Ivan:
Jmix version: 1.7

Add extra path validation to write/delete/exists operations.

SergeiAksenov2:
Tested on:
Jmix version: 1.7.999-SNAPSHOT
Jmix Studio Plugin Version: 3.0.SNAPSHOT8375-261
IntelliJ version: IntelliJ IDEA 2026.1.2

PR #5285, opened 2026-05-14
LocalFileStorage: add extra path validation

Gavrilov-Ivan:
See #5284

- Added extra check to save/remove/exists operations.
- Refactored full absolute path evaluation.

### DIFF
--- a/jmix-localfs/localfs/src/main/java/io/jmix/localfs/LocalFileStorage.java
+++ b/jmix-localfs/localfs/src/main/java/io/jmix/localfs/LocalFileStorage.java
@@ -145,7 +145,8 @@ public class LocalFileStorage implements FileStorage {
         checkStorageDefined(roots, fileRef.getFileName());
         checkPrimaryStorageAccessible(roots, fileRef.getFileName());
 
-        Path path = roots[0].resolve(relativePath);
+        Path root = roots[0];
+        Path path = root.resolve(relativePath);
         Path parentPath = path.getParent();
         if (parentPath == null) {
             throw new FileStorageException(FileStorageException.Type.IO_EXCEPTION,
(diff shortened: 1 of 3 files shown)


jmix-localfs/localfs/src/main/java/io/jmix/localfs/LocalFileStorage.java
```diff
@@ -270,6 +326,11 @@ public void removeFile(FileRef reference) {
             Path filePath = root.resolve(relativePath);
             File file = filePath.toFile();
             if (file.exists()) {
+                if (isInvalidPath(filePath, root)) {
```
knstvk: I think this check should be done before `file.exists()`

jmix-localfs/localfs/src/main/java/io/jmix/localfs/LocalFileStorage.java
```diff
@@ -286,6 +347,10 @@ public boolean fileExists(FileRef reference) {
         for (Path root : roots) {
             Path filePath = root.resolve(relativePath);
             if (filePath.toFile().exists()) {
+                if (isInvalidPath(filePath, root)) {
```
knstvk: This check should be done before `file.exists()`. Otherwise we allow probing for existence of files outside storage dir.

knstvk (changes requested):

jmix-localfs/localfs/src/main/java/io/jmix/localfs/LocalFileStorage.java
Gavrilov-Ivan: In these specific 2 cases (this and above) there is no info go outside besides log message.
But it's relevant to 'saveStream' case - there are different exception.

However - moved path check before existence check for all cases. Previously order matters because path check required physical existence of the file - now it's not necessary.

knstvk (approved):

PR #5286, opened 2026-05-14
UI observability instrumentation enhancements: data loaders, view lifecycle, fragments #5265

### LINKED ISSUE #5265: UI observability instrumentation enhancements: data loaders, view lifecycle, fragments
#4283  introduced `UiObservationSupport` and Micrometer Observation API instrumentation for actions, view runtime lifecycle (READY, BEFORE_SHOW, BEFORE_CLOSE, AFTER_CLOSE), and fragments. Alongside it, the legacy `Timer`-based `jmix.ui.views` and `jmix.ui.data` metrics from #1704 are still emitted for back-compat with existing dashboards.

Coverage of the modern Observation path is incomplete in several places:
- **Data loaders** are still observed only through the legacy `jmix.ui.data` Timer — no Observation, no spans, and `dataManager.load*` Observations live as orphan root spans rather than as children of the corresponding loader phase.
- **View early lifecycle phases** (CREATE / LOAD / INJECT / INIT in `ViewSupport.initView`) are still measured only by the legacy `jmix.ui.views` Timer, even though the `ViewLifecycle` enum already declares all 8 phases. This leaves a diagnostic blind spot for per-request slowness in XML loading, autowire, and the user-defined `InitEvent` handlers.
- **`view.id` tag is missing** from `jmix.ui.actions` Observation in many real scenarios.
- **Legacy `Timer` instrumentation is duplicated inline** at every call site in loaders and views, with no shared abstraction — making the eventual removal of the legacy path painful and error-prone.

The goal is to reach **FullObservability** for UI primitives — every lifecycle phase that produces a metric also produces a span when Observation is enabled — and to introduce a single project-wide flag that lets users opt out of the legacy Timer paths once their dashboards migrate to the modern tag schema. The solution must also be structured so that the legacy `Timer` paths can be removed in a future release as a localised change  without touching call sites in loaders, views, or fragments.

fractal3000:
#5265 

## Summary

Full UI Observation coverage for Views, Fragments, Actions and DataLoaders.

- **`view.id`** added to every modern observation, with the `"N/A"` sentinel for the anonymous case.
- **`fragment.id`** added to `jmix.ui.actions` and `jmix.ui.data` for actions/loaders living inside a fragment.
- **`full_loader_id`** (high-cardinality) for generated loaders: `loader.id="<generated>"` aggregates, the original id is preserved on the span for trace search.
- **Legacy `view` tag** (`jmix.ui.data` Timer) keeps its previous semantics for fragment-owned loaders — existing dashboards do not break.
- **View close paths** via `view.close()` and `StandardDetailView.navigate()` are now observed (previously bypassed router listeners).
- **Legacy Timer bridge** extracted into `@Deprecated(forRemoval=true)` `LegacyUiTimerSupport`.
- Pre-existing bug fixed: `view.id` for a loader inside a fragment used to contain the fragment id.
- 19 commits across three review waves; will be squashed before merge.

A paired PR in `jmix-premium` (tabbedmode Views + dynmodel-flowui adaptation) is required for tabbedmode/dynmodel users.

## Public API changes (vs master)

All public-API surface lives in the `@Experimental` package `io.jmix.flowui.observation`. Additions:

- `UiObservationSupport`: new methods `observeViewLifecycle`, `observeFragmentLifecycle`, `observeDataLoader`, and `createActionExecutionObservation(Action, @Nullable Component invocationSource)`. The `create`/`observe` duality from earlier iterations was removed — `observeXxx` is the only public way to observe lifecycle phases; `createActionExecutionObservation` stays as the single action entry point.
- `LegacyUiTimerSupport` — new `@Internal @Deprecated(forRemoval=true)` bridge. Owns `MeterRegistry` for legacy `Timer` recording; installs a `MeterFilter` that suppresses modern `Observation`-derived meters under `jmix.ui.*` when legacy monitoring is on (avoids Prometheus tag-key collisions).
- 13 `public static final` constants on `UiObservationSupport` (`OBSERVATION_NAME_PREFIX`, `VIEW_OBSERVATION_NAME`, …, `LIFECYCLE_NAME_TAG`, `VIEW_ID_TAG`, …, `NOT_AVAILABLE`) for tag-schema consistency across the modern stack and the legacy filter.
- `FragmentLifecycle` enum.
- `FragmentDataImpl` receives the host `Fragment` via constructor (no `setHostFragment` on the interface — contract not extended for observation).
- `UiProperties.legacyMonitoringEnabled` (`@Deprecated(forRemoval=true)`; property key `jmix.ui.legacy-monitoring-enabled`).
- `UiProperties.uiObservationEnabled` (property key `jmix.ui.ui-observation-enabled`).
- `DataLoader.setObservationInfoProvider` / `getObservationInfoProvider` with `DataLoaderObservationInfo(viewId, loaderId, fragmentId)`. Legacy `setMonitoringInfoProvider` / `getMonitoringInfoProvider` kept and `@Deprecated(forRemoval=true)`; the legacy setter delegates to the new one (wraps `DataLoaderMonitoringInfo(viewId, loaderId)` → `DataLoaderObservationInfo(viewId, loaderId, fragmentId=null)`).
- `DataLoaderObservationInfo` — new record, lives in `io.jmix.flowui.observation`.
- `FragmentObservationInfo` — new record (replaces the early-iteration `FragmentLifecycleObservationInfo`, which was removed; package is `@Experimental`).
- `ViewControllerUtils.fireEvent` auto-observes view lifecycle events (private `event → ViewLifecycle` mapping), so drivers like `ViewSupport.fireViewInitEvent`, `DialogWindow`, premium tabbedmode `Views` get observability for free.

## Breaking changes (vs master)

1. **Canonical record constructor** of `DataLoaderMonitoringInfo` is preserved (legacy 2-tuple `(viewId, loaderId)`). The new `DataLoaderObservationInfo` carries the 3-tuple; setters/getters on `DataLoader` were renamed accordingly. Existing legacy provider implementations keep working through the wrapper.
2. **Modern `jmix.ui.data` `view.id` tag for a fragment-owned loader** now carries the enclosing view id (it used to carry the fragment id because of a pre-existing bug). The legacy `jmix.ui.data` Timer `view` tag keeps the old value via a back-compat lookup — existing dashboards keep matching.
3. **Generated loaders** now emit observation events with `loader.id="<generated>"` plus a high-cardinality `full_loader_id` (previously they were dropped entirely as NOOP). Fresh spans/metrics will appear in Jaeger/Prometheus.
4. **`ResourceRoleModelDetailView` create-policy actions** now use a stable id `createPolicy_<ProviderViewClass>` instead of a random per-instance id.

## Rebases onto master

- **2026-06-01** — integrated `jmix#5139` (Taymanov, `View that uses DataRepository doesn't set hint DynAttrQueryHints.LOAD_DYN_ATTR`). Parallel change in `InstanceLoaderImpl#load()` was merged: our `observeDataLoader` wrapper preserved, the new `JmixDataRepositoryContext.builder().plan(...).hints(hints).build()` call adopted inside the lambda. Fix amended into the original `Add data loader lifecycle observation` commit via `git rebase --autosquash`; verified by Taymanov's `ViewDataTest#"repository instance loader receives context"` (green).
- **2026-06-02** — integrated `jmix#5287` (PlantUML). No conflicts.

## QA checklist

- [ ] `/actuator/prometheus` exposes `jmix_ui_views_*`, `jmix_ui_fragments_*`, `jmix_ui_actions_*`, `jmix_ui_data_*`.
- [ ] Opening / closing a View → spans with `view.id` / `view.class`. Closing via `view.close()` also produces a span.
- [ ] Action inside a fragment → span has `fragment.id`. Action outside any fragment → no `fragment.id`.
- [ ] Loader inside a fragment → modern `view.id` = enclosing view id, `fragment.id` = fragment id; legacy `view` tag = fragment id (back-compat).
- [ ] Generated loader → `loader.id="<generated>"` (single bucket in Prometheus), `full_loader_id` = original id (visible in the trace).
- [ ] View without an id → `view.id="N/A"`, drill-down via `view.class`.
- [ ] `:flowui:test` green.
- [ ] `jmix.ui.ui-observation-enabled=false` → zero overhead, modern metrics are not emitted.
- [ ] `jmix.ui.legacy-timer-enabled=true` → legacy `jmix.ui.data` Timer behaves as before.

## Diagnostic queries

```promql
# View classes without an id
topk(20, sum by (view_class) (rate(jmix_ui_views_seconds_count{view_id="N/A"}[5m])))

# Anonymous data loads grouped by View
sum by (view_id) (rate(jmix_ui_data_seconds_count{loader_id="<generated>"}[5m]))
```


jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/LegacyUiTimerSupport.java
```diff
@@ -0,0 +1,82 @@
...
+ * ({@code jmix.ui.data}, {@code jmix.ui.views}) when
+ * {@link UiProperties#isLegacyTimerEnabled()} is on.
+ * <p>
+ * Exists solely for back-compat with dashboards built on the legacy tag schema. The whole bean
+ * is expected to be removed in a future release once consumers migrate to the modern Observation
+ * tag schema; at that point this class — together with the property — can be deleted with no
+ * impact on call sites in {@code UiObservationSupport}.
+ */
+@Internal
+@Deprecated(since = "2.9", forRemoval = true)
```
KremnevDmitry: 2.9 will never be released
This class will be deprecated since 3.0

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/LegacyUiTimerSupport.java
```diff
@@ -0,0 +1,82 @@
...
+@SuppressWarnings("removal")
```
KremnevDmitry: This warning shouldn't be suppressed, I guess

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
```diff
@@ -110,6 +110,16 @@ public class UiProperties {
...
 
+    /**
+     * Whether to keep recording the legacy {@link io.micrometer.core.instrument.Timer}-based
+     * UI metrics with their original tag schema. Currently controls {@code jmix.ui.data}
+     * (tags {@code dataLoader}/{@code view}/{@code lifeCycle}) and {@code jmix.ui.views}
+     * (tags {@code view}/{@code lifeCycle}). The same metric names are also produced by the modern
+     * Observation path with a different tag schema; this flag is intended for back-compat with
+     * dashboards built on the legacy tag set and is expected to be turned off as dashboards migrate.
+     */
+    boolean legacyTimerEnabled;
```
KremnevDmitry: 1. The property's name currently doesn't reflect its purpose. 
Enabling a 'legacy timer' doesn't clearly refer to timer metrics. A `timer` can refer to a visual component or a facet. The property needs to be renamed to better reflect its purpose.

# Important
2. We really need to remove this functionality in the future. What if I don't want to build an observability environment, but just want to use the basic metrics from the micrometer? This seems like a different approach to metrics, not necessarily replaceable by observability.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
```diff
@@ -97,13 +97,37 @@ public void registerContainer(String id, InstanceContainer<?> container) {
...
+    /**
+     * @return id of the enclosing view, or {@code null} if not resolvable. Default delegates to
+     * {@link #getOwnerId()} so that subclasses overriding {@code getOwnerId()} keep populating
+     * {@code view.id} with no migration. Override this in new code when the enclosing view id
+     * cannot be derived from {@code getOwnerId()}.
+     */
+    @Nullable
+    protected String resolveViewId() {
+        return getOwnerId();
+    }
```
KremnevDmitry: By default, you shouldn't return `ownerId` because not all inheritor types are views

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
```diff
@@ -97,13 +97,37 @@ public void registerContainer(String id, InstanceContainer<?> container) {
...
         getLoaders().put(id, loader);
 
-        DataLoaderMonitoringInfo monitoringInfo = new DataLoaderMonitoringInfo(getOwnerId(), id);
-        loader.setMonitoringInfoProvider(dl -> monitoringInfo);
+        // Lazy provider: viewId and fragmentId are resolved on each monitoring event, not at
+        // registration time. This is required for fragment-owned loaders — at the moment of
+        // registration the fragment is not yet attached to its host, so the enclosing view is
+        // not yet reachable via the parent chain.
+        loader.setMonitoringInfoProvider(dl ->
+                new DataLoaderMonitoringInfo(resolveViewId(), id, resolveFragmentId()));
```
KremnevDmitry: I don't think we should implement such logic in an abstract class. We don't know what our implementations will be.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
```diff
@@ -97,13 +97,37 @@ public void registerContainer(String id, InstanceContainer<?> container) {
...
+
+    /**
+     * @return id of the enclosing fragment when the loader is owned by a fragment, {@code null}
+     * otherwise. Default is {@code null}; override when the holder is fragment-scoped.
+     */
+    @Nullable
+    protected String resolveFragmentId() {
+        return null;
+    }
+
```
KremnevDmitry: The same moment such as `resolveViewId`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/DataLoaderMonitoringInfo.java
```diff
@@ -16,18 +16,24 @@
...
  *
- * @param viewId   id of the view containing target data loader
- * @param loaderId id of the target data loader
+ * @param viewId     id of the enclosing {@link View}, or {@code null} if not resolvable
+ * @param loaderId   id of the target data loader
+ * @param fragmentId id of the enclosing {@link Fragment} when the loader lives inside a fragment,
+ *                   {@code null} for loaders attached directly to a view
  */
-public record DataLoaderMonitoringInfo(@Nullable String viewId, @Nullable String loaderId) {
+public record DataLoaderMonitoringInfo(@Nullable String viewId,
```
KremnevDmitry: Do we really need to change `DataLoaderMonitoringInfo` signature? Why we can't create a new `DataLoaderObservationInfo` for example? 

This way we won't have to make any mandatory changes to `UiMonitoring#stopDataLoaderTimerSample`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/fragment/FragmentData.java
```diff
@@ -36,4 +36,12 @@ public interface FragmentData extends HasDataComponents {
...
      */
     void setFragmentId(@Nullable String fragmentId);
+
+    /**
+     * Sets the host fragment so the data context can resolve the enclosing view lazily for
+     * monitoring purposes. Called by {@link Fragment} when this data context is assigned.
+     */
+    default void setHostFragment(Fragment<?> hostFragment) {
+        // default no-op
+    }
```
KremnevDmitry: `FragmentData` doesn't require this method in its contract. So, we're extending the contract solely for observation purposes, which isn't good.

I think we need to introduce new `ObservableData` interface for this? It will be great to have an internal utility class to set fragment.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/UiMonitoring.java
```diff
@@ -27,7 +27,13 @@
 /**
  * Provides utility methods and constants to facilitate UI monitoring, including measuring and recording the
  * durations of various UI-related operations and lifecycle events.
+ *
+ * @deprecated Legacy {@link Timer}-based instrumentation, kept for back-compat with dashboards built on the
+ * old tag schema. New code should use {@link io.jmix.flowui.observation.UiObservationSupport}, which
+ * provides modern Observation-based instrumentation. The internal bridge to this class —
+ * {@link LegacyUiTimerSupport} — is also marked for removal.
  */
+@Deprecated(since = "2.9", forRemoval = true)
```
KremnevDmitry: 2.9 will never be released

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/ViewSupport.java
```diff
@@ -77,73 +74,63 @@ public ViewSupport(ApplicationContext applicationContext,
...
-
-        ViewActions actions = applicationContext.getBean(ViewActions.class, view);
-        ViewControllerUtils.setViewActions(view, actions);
-
-        ViewControllerUtils.setViewFacets(view, applicationContext.getBean(ViewFacets.class, view));
-
-        stopViewTimerSample(createSample, meterRegistry, ViewLifeCycle.CREATE, viewId);
+        ViewActions actions = uiObservationSupport.observeViewLifecycle(view, ViewLifecycle.CREATE, () -> {
+            ViewControllerUtils.setViewData(view, applicationContext.getBean(ViewData.class));
+            ViewActions a = applicationContext.getBean(ViewActions.class, view);
```
KremnevDmitry: a?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -94,7 +124,9 @@ public Observation createFragmentLifecycleObservation(FragmentLifecycleObservati
         Observation observation = Observation.createNotStarted(FRAGMENT_OBSERVATION_NAME, observationRegistry)
                 .contextualName("fragment lifecycle")
                 .lowCardinalityKeyValue("lifecycle.name", lifecycle.getName())
-                .lowCardinalityKeyValue("fragment.class", observationInfo.fragmentClass());
+                .lowCardinalityKeyValue("fragment.class", observationInfo.fragmentClass())
+                .lowCardinalityKeyValue("view.id",
+                        Strings.isNullOrEmpty(observationInfo.viewId()) ? MISSING_VIEW_ID : observationInfo.viewId());
```
KremnevDmitry: Why do we set a missing view ID? I don't think we should set a cardinality if its value is undefined.

For example `fragment.id` below

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -103,7 +135,121 @@ public Observation createFragmentLifecycleObservation(FragmentLifecycleObservati
...
+    /**
+     * Records monitoring data for a value-returning {@link DataLoader} lifecycle phase.
+     *
+     * @see #observeDataLoader(DataLoader, DataLoaderLifeCycle, Runnable)
+     */
+    public <T> T observeDataLoader(DataLoader loader, DataLoaderLifeCycle phase, Supplier<T> action) {
+        Observation observation = createDataLoaderObservation(loader, phase);
+        return legacyUiTimerSupport.recordDataLoaderTimer(loader, phase,
+                () -> observation.observe(action));
+    }
```
KremnevDmitry: First we create an `observation`, and then we might not use anything at all.
Why we can't check ui property here?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -103,7 +135,121 @@ public Observation createFragmentLifecycleObservation(FragmentLifecycleObservati
...
+
+    public Observation createDataLoaderObservation(DataLoaderMonitoringInfo info, DataLoaderLifeCycle lifecycle) {
+        if (!isObservationAvailable()) {
+            return Observation.NOOP;
+        }
+
+        String loaderId = info.loaderId();
+        if (StringUtils.isBlank(loaderId)) {
+            return Observation.NOOP;
+        }
```
KremnevDmitry: How many `if` conditions returning `NOOP` are needed to squash them into one? :)

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -103,7 +135,121 @@ public Observation createFragmentLifecycleObservation(FragmentLifecycleObservati
...
+
+        return observation;
+    }
+
     public Observation createActionExecutionObservation(Action action) {
+        return createActionExecutionObservation(action, null);
+    }
+
+    public Observation createActionExecutionObservation(Action action, @Nullable Component triggerComponent) {
```
KremnevDmitry: Why should we pass the target component if it is calculated for components that have it?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -103,7 +135,121 @@ public Observation createFragmentLifecycleObservation(FragmentLifecycleObservati
...
+    /**
+     * Records monitoring data for a void {@link Fragment} lifecycle phase via the modern Observation path.
+     * <p>
+     * Note: unlike data loaders and views, fragments never had a legacy {@link io.micrometer.core.instrument.Timer}-based metric, so
+     * {@link UiProperties#isLegacyTimerEnabled()} has no effect here. No new legacy metric is introduced.
+     */
+    public void observeFragmentLifecycle(Fragment<?> fragment, FragmentLifecycle phase, Runnable action) {
+        observeFragmentLifecycle(fragment, phase, () -> {
+            action.run();
+            return null;
```
KremnevDmitry, line 239: 1. Looks not good, to many overloads. Why we don't use native `.observe()` method without returning the value?
2. Moreover, we now have two options for interaction: creating an observation using `create...Observation` and performing an observation using `observe...`. This seems inconsistent, and it's very easy to get confused about what's happening where.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/ViewSupport.java
```diff
@@ -77,73 +74,63 @@ public ViewSupport(ApplicationContext applicationContext,
...
 
         // perform injection for the nested fragments
         componentLoaderContext.executeAutowireTasks();
 
-        Timer.Sample initSample = startTimerSample(meterRegistry);
-
-        fireViewInitEvent(view);
-
-        stopViewTimerSample(initSample, meterRegistry, ViewLifeCycle.INIT, viewId);
+        uiObservationSupport.observeViewLifecycle(view, ViewLifecycle.INIT, () -> fireViewInitEvent(view));
```
KremnevDmitry: This is where we start the `observation`, but if we dig deeper...
-> `fireViewInitEvent` -> `ViewControllerUtils.fireEvent` * and here we create the observation again. I'm sure this isn't the only place where this happens.

Relates to not consistent API of `UiObservationSupport`

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/LegacyUiTimerSupport.java
fractal3000: Sorry. Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/LegacyUiTimerSupport.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
fractal3000: Implemented as discussed. Renamed to `legacyMonitoringEnabled` . Spring property key updated to `jmix.ui.legacy-monitoring-enabled`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
fractal3000: Fixed: the abstract class declares
`implements ObservableDataHolder` but provides **no** default implementation
of `getObservableViewId()` / `getObservableFragmentId()`. Each concrete
subclass (`FragmentDataImpl`, `ViewData impl`) declares its observability
context explicitly.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
fractal3000: Fixed. `registerLoader` now delegates the wiring step to
the external utility `LoaderObservabilityWiring.install(this, loader, id)`
instead of building the provider lambda in-place.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/AbstractDataComponentsHolder.java
fractal3000: Fixed. `resolveViewId` no longer exists in the abstract class. Same fix as in previous comment: the read-side getters (`getObservable*Id`) live on concrete subclasses; the wiring is in `LoaderObservabilityWiring`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/DataLoaderMonitoringInfo.java
fractal3000: Agreed and adopted:

- New record `DataLoaderObservationInfo(viewId, loaderId, fragmentId)` for the modern Observation path.
- Existing `DataLoaderMonitoringInfo(viewId, loaderId)` kept on master signature, marked @Deprecated(forRemoval = true)`, lives as long as `UiMonitoring` does.
- `DataLoader` interface gains `setObservationInfoProvider()`/`getObservationInfoProvider()`; existing `set/getMonitoringInfoProvider` deprecated.
- `LoaderObservabilityWiring.install` installs both providers and reproduces the historical `view`-tag-carries-fragmentId masquerade at the producer site, so legacy `jmix_ui_data` dashboards keep working bit-for-bit.

Further refinement:
- `DataLoaderObservationInfo` moved from `io.jmix.flowui.monitoring` (legacy package leftover) to `io.jmix.flowui.observation` — same package as the rest of the modern Observation API.
- `FragmentLifecycleObservationInfo` renamed to `FragmentObservationInfo` for symmetric naming with `DataLoaderObservationInfo` (no "Lifecycle" infix).

jmix-flowui/flowui/src/main/java/io/jmix/flowui/fragment/FragmentData.java
fractal3000: Fully agreed. Landed on a stricter resolution than the proposed `ObservableData`-interface route:

- `setHostFragment()` removed from `FragmentData` entirely.
- `FragmentDataImpl` receives the host fragment via its **constructor**; the `hostFragment` field is `final`. No setter on the impl at all.
- `FragmentsImpl.init` creates the bean with `getBean(FragmentData.class, fragment)` — identical to the existing
  pattern for `FragmentActions` / `FragmentFacets`.
- No new public interface, no utility class. The framework's existing Spring prototype-with-arg mechanism does the work.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/monitoring/UiMonitoring.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/ViewSupport.java
fractal3000: fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: The sentinel (`N/A`) is required by `PrometheusMeterRegistry`: it locks the tag-key set on the first registration of a given metric name, and rejects later registrations with a different set (warn + silent drop). Omitting `view.id` (or `fragment.id`) conditionally would lose one slice depending on which registration wins the race. Sentinel cost: one extra series per missing slot; bounded. Rationale documented next to `MISSING_VIEW_ID` / `MISSING_FRAGMENT_ID` / `MISSING_TARGET_ID` in `UiObservationSupport`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed: `isObservationAvailable()` short-circuit runs *before* `loader.getObservationInfoProvider().apply(loader)` (which walks the UI tree for fragment-owned loaders). Regression test asserts the provider is never invoked when observation is disabled while the supplied work still runs.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Redundant single-arg overload dropped. The remaining `createActionExecutionObservation(Action, @Nullable Component triggerComponent)` is the sole entry point; for `TargetAction` the view is derived from the target, and `triggerComponent` is only used when the action is not a `TargetAction`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Agreed and resolved across API cleaning up — the create/observe duality is gone. `observeXxx` is the only public way to observe view, fragment, and data-loader phases; the previous `createXxxLifecycleObservation` overloads
were inlined into `private buildXxxObservation` helpers. Action keeps a single public `createActionExecutionObservation` by design (no `observe` variant — per R14, the `Optional<UiObservationSupport>` handling lives on the caller side because actions may be constructed outside Spring).
Resulting public surface:

```
observeViewLifecycle(View, ViewLifecycle, Runnable | Supplier<T>)
observeFragmentLifecycle(Fragment, FragmentLifecycle, Runnable | Supplier<T>)
observeFragmentLifecycle(FragmentObservationInfo, FragmentLifecycle, Supplier<T>)  // for CREATE: Fragment doesn't exist yet
observeDataLoader(DataLoader, DataLoaderLifeCycle, Runnable | Supplier<T>)
createActionExecutionObservation(Action, @Nullable Component)
```

`Runnable` variants delegate to `Supplier` (`return null`) — Java ergonomics, not new surface. The legacy-timer wrapping is now contained inside the protected `buildXxx` helpers, not leaked to callers.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/sys/ViewSupport.java
fractal3000: Fixed structurally: `ViewControllerUtils.fireEvent` now auto-observes lifecycle events (private event→phase mapping + `observeViewLifecycle`). Drivers that fire through it (`ViewSupport.fireViewInitEvent`, `DialogWindow`,
premium tabbedmode `Views`) no longer add explicit `observeViewLifecycle` wraps. Drivers that fire via `Composite.fireEvent`/`super.fireEvent` (`View.afterNavigation`/`beforeEnter`/`close`, `StandardDetailView`) wrap
with `observeViewLifecycle` directly — one observer per phase on that path.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: After API cleanup, four NOOP guards remain (one per metric domain): three `private buildXxxObservation` helpers + `createActionExecutionObservation`. Each returns a different `Observation` shape, the data-loader case has an extra `StringUtils.isBlank(loaderId)` precondition, and a shared wrapper would only move the per-method branch into a passed-in lambda. Left as one `if` per method. Open to a specific factoring proposal.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -41,109 +51,236 @@
...
+     * Sentinel for the low-cardinality {@code view.id} tag when the enclosing view cannot be resolved
+     * or has no explicit id.
+     */
+    protected static final String MISSING_VIEW_ID = "N/A";
+
+    /**
+     * Sentinel for the low-cardinality {@code fragment.id} tag when there is no enclosing fragment or
+     * the fragment has no explicit id. The tag is added unconditionally so the Prometheus meter for a
+     * given metric name always carries the same set of tag keys — otherwise conditionally-added tags
+     * make {@link io.micrometer.prometheusmetrics.PrometheusMeterRegistry} reject one of the
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/DataLoaderObservationInfo.java
```diff
@@ -0,0 +1,45 @@
...
+ * @param viewId     id of the enclosing {@link View}, or {@code null} if not resolvable
+ * @param loaderId   id of the target data loader
+ * @param fragmentId id of the enclosing {@link Fragment} when the loader lives inside a fragment,
+ *                   {@code null} for loaders attached directly to a view
+ */
+public record DataLoaderObservationInfo(@Nullable String viewId,
+                                        @Nullable String loaderId,
+                                        @Nullable String fragmentId) {
+
+    protected static final DataLoaderObservationInfo EMPTY = new DataLoaderObservationInfo(null, null, null);
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/FragmentObservationInfo.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public FragmentObservationInfo(FragmentOwner parent,
+                                            @Nullable String fragmentId,
+                                            String fragmentClass) {
+        this(fragmentId, fragmentClass,
+                parent instanceof Component component ? resolveViewId(component) : null);
+    }
+
+    @Nullable
+    protected static String resolveViewId(Component component) {
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -41,109 +51,236 @@
...
+     * registrations and silently drop a slice of the data.
+     */
+    protected static final String MISSING_FRAGMENT_ID = "N/A";
+
+    /** Same rationale as {@link #MISSING_FRAGMENT_ID}, for {@code target.id} of action observations. */
+    protected static final String MISSING_TARGET_ID = "N/A";
```
KremnevDmitry: The string declaration "N/A" occurs three times

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -41,109 +51,236 @@
...
+    protected static final String GENERATED_LOADER_ID_SENTINEL = "<generated>";
 
     @Autowired(required = false)
     protected ObservationRegistry observationRegistry;
 
+    @Autowired(required = false)
+    protected MeterRegistry meterRegistry;
+
+    @Autowired
+    protected LegacyUiTimerSupport legacyUiTimerSupport;
```
KremnevDmitry: MeterRegistry here is injected using `required=false` flag, but it is injected in the `LegacyUiTimerSupport` with `required=true` flag. So we need it or not?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/DataLoader.java
```diff
@@ -111,15 +112,44 @@ public interface DataLoader {
...
     /**
-     * Sets function that provides monitoring info about this data loader.
+     * No-op retained for binary compatibility. The legacy monitoring info is now derived from
+     * {@link #setObservationInfoProvider(Function)}; setting a legacy provider has no effect.
+     *
+     * @deprecated Use {@link #setObservationInfoProvider(Function)} instead.
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     default void setMonitoringInfoProvider(Function<DataLoader, DataLoaderMonitoringInfo> monitoringInfoProvider) {
     }
```
KremnevDmitry: Broken `deprecated` contract. We need to delegate the execution of this method to the new one, otherwise it will be breaking change. `Deprecated` marks that method should work as excepted, but is not recommended and will be removed later.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/FragmentLifecycleObservationInfo.java
```diff
@@ -1,33 +0,0 @@
...
-import io.jmix.flowui.fragment.Fragment;
-import org.jspecify.annotations.Nullable;
-
-/**
- * POJO class for information about {@link Fragment} that will be used for observation cardinalities.
- *
- * @param fragmentId    id of the fragment
- * @param fragmentClass FQN of the target fragment class
- */
-public record FragmentLifecycleObservationInfo(@Nullable String fragmentId, String fragmentClass) {
```
KremnevDmitry, line 28: Im not sure that we can just remove this public record

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/ViewLifecycleObservationInfo.java
```diff
@@ -1,32 +0,0 @@
...
-
-import io.jmix.flowui.view.View;
-
-/**
- * POJO class for information about {@link View} that will be used for observation cardinalities.
- *
- * @param viewId    id of the view
- * @param viewClass FQN of the target view class
- */
-public record ViewLifecycleObservationInfo(String viewId, String viewClass) {
```
KremnevDmitry, line 27: Im not sure that we can just remove this public record

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/ViewControllerUtils.java
```diff
@@ -207,14 +208,37 @@ public static <T extends Facet> T getViewFacet(View<?> view, Class<T> facetClass
...
+        if (phase == null) {
+            ComponentUtil.fireEvent(view, event);
+            return;
+        }
+        view.getUiObservationSupport().observeViewLifecycle(view, phase,
+                () -> ComponentUtil.fireEvent(view, event));
+    }
+
+    @Nullable
+    protected static ViewLifecycle mapLifecycleEvent(ComponentEvent<?> event) {
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
```diff
@@ -252,4 +275,14 @@ public boolean isExceptionDialogModal() {
...
+
+    /**
+     * @see #legacyMonitoringEnabled
+     * @deprecated See {@link #legacyMonitoringEnabled}.
+     */
+    @Experimental
+    @Deprecated(since = "3.0", forRemoval = true)
+    public boolean isLegacyMonitoringEnabled() {
+        return legacyMonitoringEnabled;
+    }
```
KremnevDmitry: `@Experimental` and `@Deprecated` in the same time?

What should the user do? They can't use the `property` yet because it's still `experimental`, but they can't use it anymore because it's already `deprecated`?

I think the property shouldn't be `@Experimental`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -41,109 +51,236 @@
  *
  * @see UiProperties#isUiObservationEnabled()
  */
+@Experimental
 @org.springframework.stereotype.Component("flowui_UiObservationSupport")
 public class UiObservationSupport {
```
KremnevDmitry, line 53: File is not formatted. Lines is not aligned

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
```diff
@@ -41,109 +51,236 @@
...
-        return observation;
+                .lowCardinalityKeyValue("fragment.class", info.fragmentClass())
+                .lowCardinalityKeyValue("view.id",
+                        Strings.isNullOrEmpty(info.viewId()) ? MISSING_VIEW_ID : info.viewId())
+                .lowCardinalityKeyValue("fragment.id",
+                        Strings.isNullOrEmpty(info.fragmentId()) ? MISSING_FRAGMENT_ID : info.fragmentId());
     }
 
-    public Observation createActionExecutionObservation(Action action) {
+    public Observation createActionExecutionObservation(Action action, @Nullable Component triggerComponent) {
```
KremnevDmitry: `triggerComponent` is a bad name. We are trigger action on it, but it should be target... But the target is computed for the actions who has a target. Anyway naming is strange. Maybe `invocationSource`? Or something else?

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/UiProperties.java
fractal3000: @Experimental was removed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/ViewControllerUtils.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/DataLoaderObservationInfo.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/FragmentObservationInfo.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed. Other text constants was extracted.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/DataLoader.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed.  The MeterFilter moved into LegacyUiTimerSupport, which already owns the registry for legacy Timer recording. UiObservationSupport no longer carries the field or the @PostConstruct — it's now purely modern Observation.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/ViewLifecycleObservationInfo.java
fractal3000: Package `io.jmix.flowui.observation` is marked `@Experimental` at the package level (`package-info.java`), which explicitly permits API churn. `ViewLifecycleObservationInfo` had no external callers — removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/FragmentLifecycleObservationInfo.java
fractal3000: Package `io.jmix.flowui.observation` is marked `@Experimental` at the package level (`package-info.java`), which explicitly permits API churn. `FragmentLifecycleObservationInfo` was renamed to `FragmentObservationInfo` for symmetry with `DataLoaderObservationInfo`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
KremnevDmitry: I still don't fully understand why we need three separate named constants for the same string literal. Why can't we just use `NOT_AVAILABLE` everywhere?

KremnevDmitry (approved):
Can be merged after resolving merge conflicts and fixing `N/A` comment.

Changes notes, breaking changes and usage cases is required in issue description.
Also, it is required to create a jmix-docs issue to describe it.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/observation/UiObservationSupport.java
fractal3000: Fixed

PR #5295, opened 2026-05-15
Stable order of properties in metadata

### LINKED ISSUE #5294: Stable order of properties in metadata
The order of MetaProperty objects returned by `MetaClass.getProperties()` and `getOwnProperties()` should be stable and mirror the order of fields in the entity class.

This is required for https://github.com/jmix-framework/jmix/issues/5226 and https://github.com/jmix-framework/jmix/issues/5272.

knstvk:
Implemented the ordered metadata snapshots in `MetaClassImpl`, with copy-on-write publication of ordered `getProperties()` / `getOwnProperties()` views and base-to-leaf inherited ordering. 

`MetaModelLoader` now loads declared fields in a single pass, while method-based `@JmixProperty` entries still append after field-backed properties.

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
@@ -17,21 +17,23 @@
 package io.jmix.core.metamodel.model.impl;
 
 import io.jmix.core.metamodel.model.*;
+import org.jspecify.annotations.Nullable;
 
 import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
 
 public class MetaClassImpl extends MetadataObjectImpl implements MetaClass {
 
-    private Map<String, MetaProperty> propertyByName = new ConcurrentHashMap<>();
-    private Map<String, MetaProperty> ownPropertyByName = new ConcurrentHashMap<>();
+    private volatile Map<String, MetaProperty> propertyByName = Collections.emptyMap();
+    private volatile Collection<MetaProperty> properties = Collections.emptyList();
+    private volatile Map<String, MetaProperty> ownPropertyByName = Collections.emptyMap();
+    private volatile Collection<MetaProperty> ownProperties = Collections.emptyList();
 
-	private final Session session;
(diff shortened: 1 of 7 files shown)


jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
```diff
@@ -67,13 +70,13 @@ public Session getSession() {
...
     public <T> Class<T> getJavaClass() {
-        return javaClass;
+        return (Class<T>) javaClass;
     }
 
     @Override
     public Collection<MetaProperty> getProperties() {
-		return new ArrayList<>(propertyByName.values());
-	}
+        return properties;
```
glebfox, line 78: Would it be better to return an immutable collection?

jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
```diff
@@ -128,52 +132,44 @@ public MetaPropertyPath getPropertyPath(String propertyPath) {
 
     @Override
     public Collection<MetaProperty> getOwnProperties() {
-        return new ArrayList<>(ownPropertyByName.values());
+        return ownProperties;
```
glebfox, line 135: Would it be better to return an immutable collection?

glebfox (changes requested):

jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
knstvk: It's already immutable, see `publishPropertySnapshots`

jmix-core/core/src/main/java/io/jmix/core/metamodel/model/impl/MetaClassImpl.java
knstvk: It's already immutable, see `publishOwnPropertySnapshots`

glebfox (approved):

PR #5334, opened 2026-05-31
Composition collections in runtime detail view template

### LINKED ISSUE #5226: Runtime view templates

knstvk:
## Motivation

Runtime view templates (#5225) generate list/detail views from `@DetailViewTemplate` metadata, but had a known limitation: **composition collections were unsupported** — a generated detail view only rendered single-value fields, with no way to view or edit a composition's child rows.

## What it does

When a `@DetailViewTemplate` entity has composition `*-to-many` properties, the generated detail view now renders a `TabSheet`:
- the first **General** tab holds the form of single-value fields;
- each composition collection gets its own tab with a `dataGrid` (create / edit / remove). Create/edit open the line entity's own detail view in a dialog, so line entities are expected to also carry `@DetailViewTemplate`.

Entities without composition collections render a plain `formLayout` exactly as before.

## Implementation

- **`ViewTemplateHelper#getCollectionProperties`** — returns composition `*-to-many` properties (associations and datatype collections excluded), honoring `excludeProperties`.
- **`detail-template.ftl`** — adds a fetch-plan entry and a nested `<collection>` container per composition property, and a conditional `<tabSheet>`; collection-tab dataGrids use `list_create`/`list_edit`/`list_remove` with `openMode=DIALOG`.
- **Hidden back-reference** — the inverse attribute of a composition collection (the child's parent reference) is excluded by default in both the master's columns and the line entity's own detail form (restorable via `includeProperties`).
- **Localized labels** — tab labels are emitted as `msg://` references so captions resolve per-user at view-load time (the "General" tab uses a new `viewTemplate.generalTab` key).

### DIFF
--- a/jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
+++ b/jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
@@ -29,9 +29,6 @@ actions.Read=Read
 actions.Add=Add
 actions.Exclude=Exclude
 actions.Refresh=Refresh
-
-viewTemplate.generalTab=General
-
 actions.Apply=Apply
 
 actions.logout.text=Log out
@@ -491,3 +488,5 @@ io.jmix.flowui.component.genericfilter.model/FilterConfigurationModel.defaultFor
 io.jmix.flowui.component.genericfilter.model/FilterConfigurationModel.rootCondition = Root condition
 io.jmix.flowui.component.genericfilter.model/FilterConfigurationModel.sysTenantId = Tenant id
 io.jmix.flowui.component.genericfilter.model/FilterConfigurationModel.defaultForMe = Default for me
+
+viewTemplate.generalTab=General
--- a/jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
+++ b/jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
@@ -29,6 +29,9 @@ actions.Read=Read
 actions.Add=Add
 actions.Exclude=Exclude
 actions.Refresh=Refresh
+
+viewTemplate.generalTab=General
+
 actions.Apply=Apply
 
 actions.logout.text=Log out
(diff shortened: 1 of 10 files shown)


jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
```diff
@@ -29,6 +29,9 @@ actions.Read=Read
 actions.Add=Add
 actions.Exclude=Exclude
 actions.Refresh=Refresh
+
+viewTemplate.generalTab=General
```
glebfox: Don't forget to add to translations module

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/view/template/detail-template.ftl
```diff
@@ -22,11 +36,51 @@
         <action id="closeAction" type="detail_close"/>
     </actions>
     <layout>
-        <formLayout id="form" dataContainer="entityDc">
-            <#list properties as property>
-            ${componentXmlFactory.createComponentXml(property, null)}
+        <#if collectionProperties?has_content>
+        <tabSheet id="contentTabSheet" width="100%">
```
glebfox: It is better to set `alignSelf="STRETCH" instead of `width="100%"` for `tabSheet`, because it has a border by default in **Aura**, and this results in a horizontal scrollbar.

glebfox (changes requested):

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/messages.properties
knstvk: Done

jmix-flowui/flowui/src/main/resources/io/jmix/flowui/view/template/detail-template.ftl
knstvk: Done

glebfox (approved):

PR #5338, opened 2026-06-01
Remove XStream from dependencies

### LINKED ISSUE #5337: Remove XStream from dependencies
It's now used only in legacy parts of Audit and Reports add-ons: XML-based storage of entity snapshots and reports.

knstvk:
XStream is now used only in legacy parts of Audit and Reports add-ons: XML-based storage of entity snapshots and reports. Not enough motivation to maintain a dependency, so let's drop these unused legacy storage format.


KremnevDmitry (commented):
1. Credits should be removed too, I guess

Image

2. Im not sure about naming of the following attributes of `EntitySnapshot`:
```java
    @Column(name = "FETCH_PLAN_XML")
    private String fetchPlanXml;

    @Column(name = "SNAPSHOT_XML")
    private String snapshotXml;
```
We actually store JSON values in these properties.
Is renaming them part of the current issue, or should we leave everything as is?

glebfox (approved):

knstvk:
> Credits should be removed too, I guess

Done

> Im not sure about naming of the following attributes of EntitySnapshot

Should be done in a separate task, perhaps when we resurrect this feature.

KremnevDmitry (approved):

PR #5361, opened 2026-06-08
JpqlFilter: incorrect behavior when parameter type is an entity #5355

### LINKED ISSUE #5355: JpqlFilter: incorrect behavior when parameter type is an entity
### Environment

Jmix version: 2.x

### Description

When `JpqlFilter` is configured with an entity class as the parameter type and a WHERE clause
that compares an entity reference directly (e.g. `{E}.customer = :param`), the filter does not
work correctly. Instead of passing the selected entity object as a query parameter, the filter
extracts and passes the entity's ID value.

This causes EclipseLink to receive a type mismatch — an ID (UUID) where an entity is expected —
resulting in no records returned or a runtime exception.

### Use case

A developer defines a `JpqlFilter` in XML with an entity class as `parameterClass`:

```xml
<jpqlFilter parameterClass="com.example.Customer"
            where="{E}.customer = :param"/>
```

The user selects a customer in the filter's value component and applies the filter.

### Current behavior

The filter query is executed with the entity's ID (UUID) as the parameter value instead of the
entity object. The filtered data grid shows no results even when matching records exist.

### Expected behavior

The entity object selected by the user is passed to the query as-is. The filter correctly
returns all records where the `customer` field matches the selected entity.

### Notes

The bug exists since the initial implementation of `JpqlFilter`. It is not visible when using
`parameterClass` of a scalar type (String, UUID, Integer, etc.) — only entity-typed parameters
are affected.

fractal3000:
#5355

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/jpqlfilter/JpqlFilter.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/jpqlfilter/JpqlFilter.java
@@ -19,6 +19,7 @@ package io.jmix.flowui.component.jpqlfilter;
 import com.google.common.base.Strings;
 import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
 import io.jmix.core.entity.EntityValues;
+import io.jmix.core.metamodel.annotation.JmixEntity;
 import io.jmix.core.querycondition.JpqlCondition;
 import io.jmix.flowui.component.filter.SingleFilterComponentBase;
 import io.jmix.flowui.model.DataLoader;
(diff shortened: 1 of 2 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/jpqlfilter/JpqlFilter.java
```diff
@@ -103,7 +104,14 @@ protected void updateQueryCondition(@Nullable V newValue) {
             Object parameterValue = null;
             if (newValue != null) {
                 if (EntityValues.isEntity(newValue)) {
-                    parameterValue = EntityValues.getIdOrEntity(newValue);
+                    // If parameterClass is itself an entity type the JPQL compares entities directly
+                    // (e.g. "{E}.customer = :param") — pass the entity object as-is.
+                    // Otherwise (e.g. UUID param with "{E}.customer.id = :param") extract the ID.
+                    if (parameterClass != null && parameterClass.isAnnotationPresent(JmixEntity.class)) {
```
KremnevDmitry, line 110: `JpqlFilterSupport` already detects entity parameters via `Entity.class.isAssignableFrom(parameterClass)` — this introduces a second, different predicate for the same check. `@JmixEntity` is not `@Inherited`, so an entity subclass without its own annotation would be misclassified and the ID extracted again.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/jpqlfilter/JpqlFilter.java
```diff
@@ -103,7 +104,14 @@ protected void updateQueryCondition(@Nullable V newValue) {
...
+                        parameterValue = newValue;
+                    } else {
+                        parameterValue = EntityValues.getIdOrEntity(newValue);
+                    }
```
KremnevDmitry, line 114: The fix inverts the required syntax for entity-typed parameters: `{E}.customer.id = :param` (the only previously-working form) breaks, while `{E}.customer = :param` now works.

Apps migrating with the `.id` form must update their conditions. Is this approved breaking changes? If so, it should be mentioned in the issue and marked as `breaking changes` tag.

KremnevDmitry (changes requested):

PR #5362, opened 2026-06-08
Several improvements to GenericFilter's programmatic API  #5083

### LINKED ISSUE #5083: Several improvements to `GenericFilter`'s programmatic API
## Status

This issue originally listed four improvements to `GenericFilter`'s programmatic API.
#5351 (PR #5363) delivered two of them and addressed a third differently. This issue now
tracks the **remaining** work.

| # | Improvement | Status |
|---|---|---|
| 1 | `MutableConfiguration` interface — fix `Configuration` LSP violation | ❌ **Remaining** |
| 2 | Atomic "register + activate" | ⚠️ Resolved differently (see below) |
| 3 | Stable public `refreshCurrentConfiguration()` | ✅ Done in #5351 |
| 4 | `protectedFromUserDeletion` flag | ✅ Done in #5351 (with a deviation, see note) |

---

### Problem (unchanged)

`Configuration` declares mutating methods — `setName`, `setRootLogicalFilterComponent`,
`setModified`, `setFilterComponentModified`, `setFilterComponentDefaultValue`,
`resetFilterComponentDefaultValue`, `resetAllDefaultValues` — that `DesignTimeConfiguration`
cannot fulfil: it throws `UnsupportedOperationException` at runtime. There is no compile-time
signal that these methods are unsafe to call on a design-time configuration (LSP violation).

### Expected behaviour

- `Configuration` becomes **read-only** (identity, getters, `getQueryCondition`, read-only flags).
- A new `MutableConfiguration extends Configuration` exposes the mutating methods.
- `RunTimeConfiguration implements MutableConfiguration`;
  `DesignTimeConfiguration implements Configuration` — no more `UnsupportedOperationException`
  overrides.
- APIs that mutate a configuration accept/return `MutableConfiguration`, so passing a
  design-time configuration into a mutating context becomes a **compile error**, not a
  runtime exception.

### Compatibility note

`Configuration` is public API. Moving the mutators to `MutableConfiguration` is a breaking
change for any caller that mutates a configuration through the `Configuration` type. The
migration must be completed within the same change (callers updated; no mutators left
deprecated-but-still-declared on `Configuration`).

---

## Notes on what was already delivered

**Point 4 — `protectedFromUserDeletion` (done, with deviations):**
The flag currently lives on the concrete `RunTimeConfiguration`
(`isProtectedFromUserDeletion()` / `setProtectedFromUserDeletion()`) — once
`MutableConfiguration` exists it should move there. By design it only **hides the Remove
action** in the filter toolbar (`GenericFilterRemoveAction#isApplicable`); programmatic
`removeConfiguration()` still removes the configuration (it is used internally, e.g. by the
rename flow). This intentionally deviates from the original wording ("prevents
`removeConfiguration()`").

**Point 2 — atomic register + activate (resolved differently):**
Instead of a dedicated `addAndSetCurrentConfiguration()` method, the wrong-order / silent
no-op problem is mitigated by `setCurrentConfiguration()` now **logging a warning** when the
configuration is not registered. The atomic method was prototyped during #5351 and
intentionally dropped to avoid growing the API surface. No further action planned unless a
concrete need arises.

fractal3000:
#5083

- MutableConfiguration: new interface extending Configuration with all  mutating methods. RunTimeConfiguration now implements MutableConfiguration instead of Configuration, giving compile-time safety against calling mutating methods on a DesignTimeConfiguration. Mutating methods on Configuration are deprecated in favour of MutableConfiguration.

- addAndSetCurrentConfiguration(): atomic alternative to calling  addConfiguration() followed by setCurrentConfiguration(). The two-step  sequence silently does nothing when the configuration has not yet been
  registered at the time setCurrentConfiguration is invoked.

- refreshCurrentConfiguration(): stable public method to re-render the filter UI after programmatic modification of the current configuration. Was previously only accessible as the protected refreshCurrentConfigurationLayout().

- MutableConfiguration.protectedFromUserDeletion: flag that prevents the user from deleting a programmatic configuration through the UI. When true, GenericFilterRemoveAction hides the Remove button for that configuration,
  and GenericFilter.removeConfiguration() silently skips it.

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/Configuration.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/Configuration.java
@@ -18,13 +18,18 @@ package io.jmix.flowui.component.genericfilter;
 
 import io.jmix.core.querycondition.LogicalCondition;
 import io.jmix.flowui.component.filter.FilterComponent;
+import io.jmix.flowui.component.genericfilter.configuration.DesignTimeConfiguration;
 import io.jmix.flowui.component.genericfilter.configuration.RunTimeConfiguration;
 import io.jmix.flowui.component.logicalfilter.LogicalFilterComponent;
 
-import org.jspecify.annotations.Nullable;
+import org.springframework.lang.Nullable;
 
 /**
  * A configuration is a set of filter components.
+ *
+ * <p><b>Note:</b> several mutating methods declared in this interface throw
+ * {@link UnsupportedOperationException} when called on a {@link DesignTimeConfiguration}.
+ * Those methods are deprecated — use {@link MutableConfiguration} when you need to call them.
  */
 public interface Configuration extends Comparable<Configuration> {
 
@@ -45,12 +50,15 @@ public interface Configuration extends Comparable<Configuration> {
     String getName();
 
     /**
-     * Sets the name of configuration. This method is only available for
-     * the {@link RunTimeConfiguration}.
+     * Sets the name of configuration.
      *
      * @param name a configuration name
+     * @throws UnsupportedOperationException when called on {@link DesignTimeConfiguration}
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      * @see RunTimeConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     void setName(@Nullable String name);
 
     /**
@@ -60,13 +68,16 @@ public interface Configuration extends Comparable<Configuration> {
     LogicalFilterComponent<?> getRootLogicalFilterComponent();
 
     /**
-     * Sets the root element of configuration. This method is only available for
-     * the {@link RunTimeConfiguration}.
+     * Sets the root element of configuration.
      *
      * @param rootLogicalFilterComponent a root element of configuration
+     * @throws UnsupportedOperationException when called on {@link DesignTimeConfiguration}
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      * @see LogicalFilterComponent
      * @see RunTimeConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     void setRootLogicalFilterComponent(LogicalFilterComponent<?> rootLogicalFilterComponent);
 
     /**
@@ -76,15 +87,22 @@ public interface Configuration extends Comparable<Configuration> {
 
     /**
      * @return true if the configuration is modified
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     boolean isModified();
 
     /**
      * Sets whether configuration is modified. If a filter component is modified,
      * then a remove button appears next to it.
      *
-     * @param modified whether configuration is modified.
+     * @param modified whether configuration is modified
+     * @throws UnsupportedOperationException when called on {@link DesignTimeConfiguration}
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     void setModified(boolean modified);
 
     /**
@@ -93,7 +111,10 @@ public interface Configuration extends Comparable<Configuration> {
      *
      * @param filterComponent the filter component to check
      * @return whether the filter component of configuration is modified
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     boolean isFilterComponentModified(FilterComponent filterComponent);
 
     /**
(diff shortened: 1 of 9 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/configuration/DesignTimeConfiguration.java
```diff
@@ -24,10 +24,14 @@
 import io.jmix.flowui.component.genericfilter.GenericFilter;
 import io.jmix.flowui.component.logicalfilter.LogicalFilterComponent;
 
-import org.jspecify.annotations.Nullable;
+import org.springframework.lang.Nullable;
```
KremnevDmitry, line 27: We are using jspecify nullability annotations, spring nullable is deprecated

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/MutableConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+ * only implements the read-only {@link Configuration}.
+ * <p>
+ * Use this interface as the declared type when you need to call mutating methods,
+ * so the compiler catches misuse of {@link DesignTimeConfiguration} at compile time:
+ * <pre>{@code
+ * MutableConfiguration config = filter.runtimeConfigurationBuilder()
+ *         .id("myConfig")
+ *         .buildAndRegister();
+ * config.setModified(true);
+ * }</pre>
```
KremnevDmitry, line 38: Looks like a useless information, especially about implementations. Moreover, there is no runtimeConfigurationBuilder and such API on this moment.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/MutableConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @see Configuration
+ * @see RunTimeConfiguration
+ */
+@SuppressWarnings({"deprecation", "removal"})
```
KremnevDmitry, line 43: I think we shouldn't suppress this warnings on whole class

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/MutableConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+public interface MutableConfiguration extends Configuration {
+
+    /**
+     * Returns whether the configuration is modified.
```
KremnevDmitry, line 47: `@returns` should be used, the same for all classes

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/Configuration.java
```diff
@@ -133,20 +170,33 @@ public interface Configuration extends Comparable<Configuration> {
...
+     * Returns whether the configuration is available for all users.
+     *
      * @return true if the configuration is available for all users, otherwise false.
+     * @deprecated use {@link MutableConfiguration} instead
+     * @see MutableConfiguration
      */
+    @Deprecated(since = "3.0", forRemoval = true)
     default boolean isAvailableForAllUsers() {
         return false;
     }
```
KremnevDmitry, line 191: This method shouldn't be deprecated because it used for calculating configuration name for all types of configuration

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/Configuration.java
```diff
@@ -45,12 +50,15 @@ public interface Configuration extends Comparable<Configuration> {
     String getName();
 
     /**
-     * Sets the name of configuration. This method is only available for
-     * the {@link RunTimeConfiguration}.
+     * Sets the name of configuration.
      *
      * @param name a configuration name
+     * @throws UnsupportedOperationException when called on {@link DesignTimeConfiguration}
+     * @deprecated use {@link MutableConfiguration} instead
```
KremnevDmitry, line 57: For use instead we should pass the appropriate methods in new interface e.g.:`{@link MutableConfiguration#setName}`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -891,7 +920,8 @@ public void addConfiguration(Configuration configuration) {
      */
     public void removeConfiguration(Configuration configuration) {
         if (configuration != getEmptyConfiguration()
-                && !(configuration instanceof DesignTimeConfiguration)) {
+                && !(configuration instanceof DesignTimeConfiguration)
+                && !(configuration instanceof MutableConfiguration mc && mc.isProtectedFromUserDeletion())) {
```
KremnevDmitry, line 924: We shouldn't re-check this condition here. The runtime configuration should be available for internal removing, e.g. for the renaming process in the `io.jmix.flowui.component.genericfilter.GenericFilterSupport#initFilterConfiguration`. Otherwise the configuration will be duplicated.

Availablility of user-deletion should be handled (and already handled) in the GenericFilterRemoveAction `isApplicable` method.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/Configuration.java
```diff
@@ -18,13 +18,18 @@
...
+import org.springframework.lang.Nullable;
 
 /**
  * A configuration is a set of filter components.
+ *
+ * <p><b>Note:</b> several mutating methods declared in this interface throw
+ * {@link UnsupportedOperationException} when called on a {@link DesignTimeConfiguration}.
+ * Those methods are deprecated — use {@link MutableConfiguration} when you need to call them.
  */
 public interface Configuration extends Comparable<Configuration> {
```
KremnevDmitry, line 34: We have a lot of calls of all deprecated method. I think we should rewrite out own API if it is possible. Otherwise why do we introduce a new API if we are not using it.

Don't forget about backward compatibility.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -603,6 +603,35 @@ public void setCurrentConfiguration(Configuration currentConfiguration) {
...
+     * followed by {@link #setCurrentConfiguration(Configuration)}, which fails silently when
+     * the configuration is not yet registered at the time {@code setCurrentConfiguration} is
+     * invoked.
+     *
+     * @param configuration the configuration to register and activate
+     */
+    public void addAndSetCurrentConfiguration(Configuration configuration) {
+        addConfiguration(configuration);
+        setCurrentConfiguration(configuration);
+    }
```
KremnevDmitry, line 620: Looks like a useless method. I think we should throw an exception or write a error log message when the user or programmer trying to set current configuration which does not exist.

KremnevDmitry (changes requested):

PR #5363, opened 2026-06-08
GenericFilter: add convenience API for programmatic filter configuration #5351

### LINKED ISSUE #5351: GenericFilter: add convenience API for programmatic filter configuration
It is the continue of the issue #5083 

## Description

Creating a filter configuration in Java code requires performing a sequence of initialization steps in a specific order. The correct order is not documented and differs from what happens automatically in XML. Missing a step produces no compile-time or runtime error, but the filter behaves incorrectly: for example, calling `PropertyFilter.setProperty()` before `PropertyFilter.setDataLoader()` causes the operation selector to silently use defaults; not calling `FilterComponent.setConditionModificationDelegated(true)` makes conditions non-removable at runtime.

## Use case

A developer needs to set up a `GenericFilter` configuration in Java — for example, to apply a default configuration with values that depend on runtime context not available at XML load time, or to dynamically add and remove conditions based on user role. Currently this requires reading the XML loader source to discover what initialization steps are needed.

## Expected behavior

Creating a filter component or configuration programmatically should be as reliable as declaring it in XML. The framework should enforce the required initialization order, not leave it to the caller.

## Impact

Without a safe programmatic API, `GenericFilter` is effectively XML-only for anything non-trivial. Developers who need runtime configurations are forced to work with low-level internals.

fractal3000:
# GenericFilter: fluent builder API for programmatic filter configuration

Closes #5351.

## Problem

Configuring a `GenericFilter` programmatically today requires manually reproducing
the initialisation the framework performs internally (delegating condition
modification, wiring the `DataLoader`, generating the value component and parameter
name, ordering `setProperty → setOperation → setValue`), plus manual wiring of
`RunTimeConfiguration`. Missing any step silently produces wrong behaviour.

## Solution

Two new `@Experimental` fluent builders. Single filter components are built by
assembling a condition model from the fluent parameters and delegating to the same
`FilterConverter` the framework uses for XML loading and the add-condition UI — the
converter is the single source of truth, so a programmatically built component is
initialised exactly like a declarative one.

The owning `GenericFilter` must have a `DataLoader` (as it always does in practice);
`build()` / `buildAndRegister()` fail fast otherwise.

### `FilterComponentBuilder`

Obtained via `filter.filterComponentBuilder()`.

```java
// PropertyFilter — auto-generates parameterName
PropertyFilter<String> nameFilter = filter.filterComponentBuilder()
        .propertyFilter()
        .property("name")
        .operation(PropertyFilter.Operation.CONTAINS)
        .defaultValue("Acme")
        .operationEditable(true)
        .build();

// JpqlFilter (void) — a checkbox that toggles a fixed condition on/off.
// Value type is Boolean; defaultValue(true) makes it active by default.
JpqlFilter<Boolean> activeFilter = filter.filterComponentBuilder()
        .jpqlFilter()
        .where("{E}.status = 'ACTIVE'")
        .defaultValue(true)
        .label("Active only")
        .build();

// JpqlFilter (typed) — a user-supplied value is bound to the query parameter.
// parameterName is optional (auto-generated); set it only for a named bind parameter.
JpqlFilter<String> codeFilter = filter.filterComponentBuilder()
        .jpqlFilter(String.class)
        .where("{E}.code = ?")
        .build();

// GroupFilter — copies autoApply from the owning GenericFilter; addAll() for many at once
GroupFilter group = filter.filterComponentBuilder()
        .groupFilter()
        .operation(LogicalFilterComponent.Operation.OR)
        .addAll(nameFilter, activeFilter)
        .build();
```

All three builders are **one-shot**: calling `build()` twice on the same instance
throws `IllegalStateException`.

### `RunTimeConfigurationBuilder`

Obtained via `filter.runtimeConfigurationBuilder()`.

```java
filter.runtimeConfigurationBuilder()
        .id("dynamicSearch")
        .name("Dynamic Search")
        .addAll(nameFilter, activeFilter)
        .add(codeFilter, "DEFAULT-CODE")   // typed default — compile-time safe
        .makeCurrent()
        .buildAndRegister();
```

Key behaviours:
- Configurations are **protected from user deletion by default**; opt out with
  `allowDeletion()` or `allowDeletion(boolean)`. This only hides the Remove action
  in the UI — programmatic `removeConfiguration()` still works (e.g. for renaming).
- `add(SingleFilterComponentBase<V>, @Nullable V)` — typed overload catches
  type mismatches between a filter and its default value at compile time.
  `addAll(FilterComponent...)` adds several components without defaults.
- A duplicate configuration `id` is rejected with `IllegalStateException`.
- `makeCurrent()` activates the configuration immediately (it is **not** the
  persistent default-configuration marker set by the `genericFilter_makeDefault`
  action).

### `GenericFilter` additions

| Method | Description |
|---|---|
| `filterComponentBuilder()` | Returns a new `FilterComponentBuilder` |
| `runtimeConfigurationBuilder()` | Returns a new `RunTimeConfigurationBuilder` |
| `refreshCurrentConfiguration()` | Public, stable equivalent of the protected `refreshCurrentConfigurationLayout()` — re-renders the current configuration after programmatic changes |
| `setCurrentConfiguration(Configuration)` | Now logs a warning when the configuration is not registered, instead of silently doing nothing |

## API design notes

- **`jpqlFilter()` returns `JpqlFilterBuilder<Boolean>`** — a void filter has no
  query parameter and is rendered as a checkbox that toggles its fixed WHERE clause.
  Its value type is therefore `Boolean`; `defaultValue(true)` makes the condition
  active by default. (Use `jpqlFilter(Boolean.class)` instead when a boolean value
  must be *bound* to a query parameter, e.g. `{E}.closed = :isClosed`.)

- **`parameterName` is optional** for `JpqlFilter` — auto-generated like the XML
  loader; set it explicitly only for a named bind parameter.

- **`addAll(FilterComponent...)`** is named distinctly rather than overloading
  `add` to avoid an ambiguous overload with `add(SingleFilterComponentBase, V)`.

- **`allowDeletion(boolean)`** — symmetrical with `hasInExpression(boolean)`;
  `allowDeletion()` is a no-arg convenience alias for `allowDeletion(true)`.

- **The builders and their nested types are extensible** — `protected`
  constructors/fields and overridable factory methods.

- **All new public API is annotated `@Experimental`** — may be changed or
  removed in a future minor version.

## What is not in scope

- XML loader changes — the builders are a purely programmatic API on top of
  the existing component model.
- `DesignTimeConfiguration` builders — design-time configs are managed via
  the Studio XML editor.

## Tests

`GenericFilterBuilderApiTest` — integration tests covering:

- `PropertyFilterBuilder`: property/operation, auto parameterName, defaultValue,
  operationEditable, operationTextVisible, one-shot guard
- `JpqlFilterBuilder`: void (checkbox) and typed variants, void WHERE applied only
  when active, hasInExpression, value-component generation, auto parameterName,
  one-shot guard
- DataLoader requirement (build fails fast without a loader)
- `GroupFilterBuilder`: operation, children, `addAll`, autoApply propagation,
  DataLoader propagation, one-shot guard
- `RunTimeConfigurationBuilder`: registration, `addAll`, `makeCurrent`, operation,
  id guard, duplicate-id guard, one-shot guard, typed add, default value map,
  setValue on component, modified state, allowDeletion variants,
  protectedFromUserDeletion (and that programmatic removal still works)
- One-shot guard: a build that fails validation does not consume the builder
- `GenericFilter` helpers: `refreshCurrentConfiguration`

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
@@ -242,7 +242,6 @@ public class FilterComponentBuilder {
                 throw new IllegalStateException(
                         "PropertyFilterBuilder.build() must not be called more than once; create a new instance per PropertyFilter");
             }
-            built = true;
             if (property == null) {
                 throw new IllegalStateException(
                         "PropertyFilterBuilder: 'property' is required — call .property(\"...\") before .build()");
@@ -269,6 +268,7 @@ public class FilterComponentBuilder {
             if (defaultValue != null) {
                 propertyFilter.setValue(defaultValue);
             }
+            built = true;
             return propertyFilter;
         }
     }
@@ -376,7 +376,6 @@ public class FilterComponentBuilder {
                 throw new IllegalStateException(
                         "JpqlFilterBuilder.build() must not be called more than once; create a new instance per JpqlFilter");
             }
-            built = true;
             if (where == null) {
                 throw new IllegalStateException(
                         "JpqlFilterBuilder: 'where' is required — call .where(\"...\") before .build()");
@@ -401,6 +400,7 @@ public class FilterComponentBuilder {
             if (defaultValue != null) {
                 jpqlFilter.setValue(defaultValue);
             }
+            built = true;
             return jpqlFilter;
         }
     }
@@ -447,9 +447,6 @@ public class FilterComponentBuilder {
 
         /**
          * Adds several filter components to this group at once.
-         * <p>
-         * Named {@code addAll} (rather than overloading {@code add}) for consistency with
-         * {@link RunTimeConfigurationBuilder#addAll(FilterComponent...)}.
          */
         public GroupFilterBuilder addAll(FilterComponent... filterComponents) {
             checkNotNullArgument(filterComponents, "filterComponents must not be null");
@@ -470,7 +467,6 @@ public class FilterComponentBuilder {
                 throw new IllegalStateException(
                         "GroupFilterBuilder.build() must not be called more than once; create a new instance per GroupFilter");
             }
-            built = true;
             checkDataLoaderPresent(filter);
 
             GroupFilter groupFilter = uiComponents.create(GroupFilter.class);
@@ -481,6 +477,7 @@ public class FilterComponentBuilder {
             for (FilterComponent component : components) {
                 groupFilter.add(component);
             }
+            built = true;
             return groupFilter;
         }
     }
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
@@ -1,0 +1,492 @@
+/*
+ * Copyright 2026 Haulmont.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.jmix.flowui.component.genericfilter;
+
+import io.jmix.core.Metadata;
+import io.jmix.core.annotation.Experimental;
+import io.jmix.core.querycondition.PropertyConditionUtils;
+import io.jmix.flowui.UiComponents;
+import io.jmix.flowui.component.filter.FilterComponent;
+import io.jmix.flowui.component.genericfilter.converter.FilterConverter;
+import io.jmix.flowui.component.genericfilter.registration.FilterComponents;
+import io.jmix.flowui.component.jpqlfilter.JpqlFilter;
+import io.jmix.flowui.component.jpqlfilter.JpqlFilterSupport;
+import io.jmix.flowui.component.logicalfilter.GroupFilter;
+import io.jmix.flowui.component.logicalfilter.LogicalFilterComponent;
+import io.jmix.flowui.component.propertyfilter.PropertyFilter;
+import io.jmix.flowui.entity.filter.FilterValueComponent;
+import io.jmix.flowui.entity.filter.JpqlFilterCondition;
+import io.jmix.flowui.entity.filter.PropertyFilterCondition;
+import org.jspecify.annotations.Nullable;
+import org.springframework.context.ApplicationContext;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static io.jmix.core.common.util.Preconditions.checkNotNullArgument;
+
+/**
+ * A factory for creating filter components ({@link PropertyFilter}, {@link JpqlFilter},
+ * {@link GroupFilter}) that are pre-configured for use inside a specific {@link GenericFilter}.
+ * <p>
+ * Single filter components are built by assembling a condition model from the fluent parameters
+ * and delegating to the same converter the framework uses when loading a filter from XML or
+ * adding a condition through the UI. The converter is the single source of truth for component
+ * initialisation, so a programmatically built component behaves exactly like a declarative one.
+ * <p>
+ * The owning {@link GenericFilter} must have a {@code DataLoader}; otherwise {@code build()}
+ * throws an {@link IllegalStateException}.
+ * <p>
+ * This factory and its nested builders are designed for extension: their constructors and
+ * fields are {@code protected}. To customise behaviour, subclass {@code FilterComponentBuilder}
+ * and override the relevant factory method ({@link #propertyFilter()}, {@link #jpqlFilter()},
+ * {@link #groupFilter()}) to return your own builder subclass, then return your factory from a
+ * {@link GenericFilter} subclass that overrides {@link GenericFilter#filterComponentBuilder()}.
+ * <p>
+ * Obtain an instance via {@link GenericFilter#filterComponentBuilder()}:
+ * <pre>{@code
+ * FilterComponentBuilder builder = filter.filterComponentBuilder();
+ *
+ * PropertyFilter<String> nameFilter = builder.propertyFilter()
+ *         .property("name")
+ *         .operation(PropertyFilter.Operation.CONTAINS)
+ *         .defaultValue("Acme")
+ *         .build();
+ *
+ * // void (checkbox) filter: toggles a fixed condition on/off
+ * JpqlFilter<Boolean> activeFilter = builder.jpqlFilter()
+ *         .where("{E}.status = 'ACTIVE'")
+ *         .defaultValue(true)
+ *         .build();
+ *
+ * // typed filter: a user-supplied value is bound to the query parameter
+ * JpqlFilter<String> codeFilter = builder.jpqlFilter(String.class)
+ *         .where("{E}.code = ?")
+ *         .build();
+ * }</pre>
+ */
+@Experimental
+public class FilterComponentBuilder {
+
+    protected final GenericFilter filter;
+    protected final UiComponents uiComponents;
+    protected final Metadata metadata;
+    protected final FilterComponents filterComponents;
+    protected final JpqlFilterSupport jpqlFilterSupport;
+
+    protected FilterComponentBuilder(GenericFilter filter, ApplicationContext applicationContext) {
+        this.filter = filter;
+        this.uiComponents = applicationContext.getBean(UiComponents.class);
+        this.metadata = applicationContext.getBean(Metadata.class);
+        this.filterComponents = applicationContext.getBean(FilterComponents.class);
+        this.jpqlFilterSupport = applicationContext.getBean(JpqlFilterSupport.class);
+    }
+
+    /**
+     * Returns a builder for a {@link PropertyFilter} bound to the owning filter.
+     *
+     * @param <V> the value type inferred from the entity property
+     * @return a new {@link PropertyFilterBuilder}
+     */
+    public <V> PropertyFilterBuilder<V> propertyFilter() {
+        return new PropertyFilterBuilder<>(filter, metadata, filterComponents);
+    }
+
+    /**
+     * Returns a builder for a <em>void</em> (checkbox) {@link JpqlFilter}.
+     * <p>
+     * Use this variant for a condition with a fixed WHERE clause that carries no user-supplied
+     * value, e.g. {@code {E}.status = 'ACTIVE'}. The filter is rendered as a checkbox: when
+     * checked the WHERE clause is applied, when unchecked it is omitted. The value type is
+     * therefore {@link Boolean}; use {@code defaultValue(true)} to make the condition active by
+     * default.
+     *
+     * @return a new {@link JpqlFilterBuilder} that produces a checkbox filter
+     */
+    public JpqlFilterBuilder<Boolean> jpqlFilter() {
+        return new JpqlFilterBuilder<>(filter, metadata, filterComponents, jpqlFilterSupport, Void.class);
+    }
+
+    /**
+     * Returns a builder for a <em>typed</em> {@link JpqlFilter} whose user-supplied value is
+     * bound to the query parameter.
+     *
+     * @param parameterClass the Java class of the query parameter
+     * @param <V>            the value type
+     * @return a new {@link JpqlFilterBuilder}
+     */
+    public <V> JpqlFilterBuilder<V> jpqlFilter(Class<V> parameterClass) {
+        checkNotNullArgument(parameterClass, "parameterClass must not be null");
+        return new JpqlFilterBuilder<>(filter, metadata, filterComponents, jpqlFilterSupport, parameterClass);
+    }
+
+    /**
+     * Returns a builder for a {@link GroupFilter} bound to the owning filter.
+     *
+     * @return a new {@link GroupFilterBuilder}
+     */
+    public GroupFilterBuilder groupFilter() {
+        return new GroupFilterBuilder(filter, uiComponents);
+    }
+
+    protected static void checkDataLoaderPresent(GenericFilter filter) {
+        if (filter.getDataLoader() == null) {
+            throw new IllegalStateException(
+                    "GenericFilter has no DataLoader. Filter components cannot be built without a DataLoader; " +
+                            "set it before using the builder (the 'dataLoader' attribute in XML, or " +
+                            "GenericFilter.setDataLoader(...)).");
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // PropertyFilterBuilder
+    // -------------------------------------------------------------------------
+
+    /**
+     * Fluent builder for {@link PropertyFilter}.
+     *
+     * @param <V> the value type
+     */
+    public static class PropertyFilterBuilder<V> {
+
+        protected final GenericFilter filter;
+        protected final Metadata metadata;
+        protected final FilterComponents filterComponents;
+
+        protected String property;
+        protected PropertyFilter.Operation operation;
+        protected V defaultValue;
+        protected String label;
+        protected boolean operationEditable = false;
+        protected boolean operationTextVisible = true;
+        protected boolean built = false;
+
+        protected PropertyFilterBuilder(GenericFilter filter, Metadata metadata, FilterComponents filterComponents) {
+            this.filter = filter;
+            this.metadata = metadata;
+            this.filterComponents = filterComponents;
+        }
+
+        /**
+         * Sets the entity property path (e.g. {@code "name"} or {@code "customer.city.name"}).
+         * Required.
+         */
+        public PropertyFilterBuilder<V> property(String property) {
+            checkNotNullArgument(property, "property must not be null");
+            this.property = property;
+            return this;
+        }
+
+        /**
+         * Sets the filtering operation (e.g. {@link PropertyFilter.Operation#EQUAL}).
+         * Required.
+         */
+        public PropertyFilterBuilder<V> operation(PropertyFilter.Operation operation) {
+            checkNotNullArgument(operation, "operation must not be null");
+            this.operation = operation;
+            return this;
+        }
+
+        /**
+         * Sets the default (pre-filled) value for the filter condition. Optional.
+         */
+        public PropertyFilterBuilder<V> defaultValue(@Nullable V defaultValue) {
+            this.defaultValue = defaultValue;
+            return this;
+        }
+
+        /**
+         * Overrides the auto-generated label text. Optional.
+         */
+        public PropertyFilterBuilder<V> label(@Nullable String label) {
+            this.label = label;
+            return this;
+        }
+
+        /**
+         * Controls whether the user can change the filter operation at runtime via a
+         * dropdown selector. Defaults to {@code false}.
+         */
+        public PropertyFilterBuilder<V> operationEditable(boolean operationEditable) {
+            this.operationEditable = operationEditable;
+            return this;
+        }
+
+        /**
+         * Controls whether the operation label (e.g. "contains", "=") is shown next to
+         * the value field. Defaults to {@code true}.
+         */
+        public PropertyFilterBuilder<V> operationTextVisible(boolean operationTextVisible) {
+            this.operationTextVisible = operationTextVisible;
+            return this;
+        }
+
+        /**
+         * Builds and returns the configured {@link PropertyFilter}.
+         *
+         * @throws IllegalStateException if {@code property} or {@code operation} was not set,
+         *         if the owning filter has no DataLoader, or if this builder instance has
+         *         already been used
+         */
+        @SuppressWarnings("unchecked")
+        public PropertyFilter<V> build() {
+            if (built) {
+                throw new IllegalStateException(
+                        "PropertyFilterBuilder.build() must not be called more than once; create a new instance per PropertyFilter");
+            }
+            built = true;
+            if (property == null) {
+                throw new IllegalStateException(
+                        "PropertyFilterBuilder: 'property' is required — call .property(\"...\") before .build()");
+            }
+            if (operation == null) {
+                throw new IllegalStateException(
+                        "PropertyFilterBuilder: 'operation' is required — call .operation(...) before .build()");
+            }
+            checkDataLoaderPresent(filter);
+
+            PropertyFilterCondition model = metadata.create(PropertyFilterCondition.class);
+            model.setProperty(property);
+            model.setOperation(operation);
+            model.setParameterName(PropertyConditionUtils.generateParameterName(property));
+            model.setOperationEditable(operationEditable);
+            model.setOperationTextVisible(operationTextVisible);
+            model.setLabel(label);
+            model.setValueComponent(metadata.create(FilterValueComponent.class));
+
+            FilterConverter<PropertyFilter<V>, PropertyFilterCondition> converter =
+                    (FilterConverter<PropertyFilter<V>, PropertyFilterCondition>) filterComponents
+                            .getConverterByModelClass(PropertyFilterCondition.class, filter);
+            PropertyFilter<V> propertyFilter = converter.convertToComponent(model);
+            if (defaultValue != null) {
+                propertyFilter.setValue(defaultValue);
+            }
+            return propertyFilter;
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // JpqlFilterBuilder
+    // -------------------------------------------------------------------------
+
+    /**
+     * Fluent builder for {@link JpqlFilter}.
+     *
+     * @param <V> the value type ({@link Boolean} for a void/checkbox filter, or the parameter type)
+     */
+    public static class JpqlFilterBuilder<V> {
+
+        protected final GenericFilter filter;
+        protected final Metadata metadata;
+        protected final FilterComponents filterComponents;
+        protected final JpqlFilterSupport jpqlFilterSupport;
+        protected final Class<?> parameterClass;
+
+        protected String where;
+        protected String join;
+        protected String parameterName;
+        protected V defaultValue;
+        protected String label;
+        protected boolean hasInExpression;
+        protected boolean built = false;
+
+        protected JpqlFilterBuilder(GenericFilter filter, Metadata metadata, FilterComponents filterComponents,
+                          JpqlFilterSupport jpqlFilterSupport, Class<?> parameterClass) {
+            this.filter = filter;
+            this.metadata = metadata;
+            this.filterComponents = filterComponents;
+            this.jpqlFilterSupport = jpqlFilterSupport;
+            this.parameterClass = parameterClass;
+        }
+
+        /**
+         * Sets the JPQL WHERE clause fragment (use {@code {E}} as the entity alias and
+         * {@code ?} as the parameter placeholder). Required.
+         */
+        public JpqlFilterBuilder<V> where(String where) {
+            checkNotNullArgument(where, "where must not be null");
+            this.where = where;
+            return this;
+        }
+
+        /**
+         * Sets the optional JPQL JOIN clause fragment (e.g. {@code "join {E}.tags t"}).
+         */
+        public JpqlFilterBuilder<V> join(@Nullable String join) {
+            this.join = join;
+            return this;
+        }
+
+        /**
+         * Sets the query parameter name. Optional — if not set, a name is generated
+         * automatically (matching the behaviour of the XML loader). Set it explicitly only
+         * when the WHERE clause uses a named bind parameter ({@code :paramName}) that must
+         * match a specific name.
+         */
+        public JpqlFilterBuilder<V> parameterName(String parameterName) {
+            checkNotNullArgument(parameterName, "parameterName must not be null");
+            this.parameterName = parameterName;
+            return this;
+        }
+
+        /**
+         * Sets the default (pre-filled) value. Optional.
+         * <p>
+         * For a void (checkbox) filter the value type is {@link Boolean}: {@code defaultValue(true)}
+         * makes the condition active by default.
+         */
+        public JpqlFilterBuilder<V> defaultValue(@Nullable V defaultValue) {
+            this.defaultValue = defaultValue;
+            return this;
+        }
+
+        /**
+         * Overrides the auto-generated label text. Optional.
+         */
+        public JpqlFilterBuilder<V> label(@Nullable String label) {
+            this.label = label;
+            return this;
+        }
+
+        /**
+         * Marks the condition as having an IN expression (the value is a collection).
+         */
+        public JpqlFilterBuilder<V> hasInExpression(boolean hasInExpression) {
+            this.hasInExpression = hasInExpression;
+            return this;
+        }
+
+        /**
+         * Builds and returns the configured {@link JpqlFilter}.
+         *
+         * @throws IllegalStateException if {@code where} was not set, if the owning filter has
+         *         no DataLoader, or if this builder instance has already been used
+         */
+        @SuppressWarnings("unchecked")
+        public JpqlFilter<V> build() {
+            if (built) {
+                throw new IllegalStateException(
+                        "JpqlFilterBuilder.build() must not be called more than once; create a new instance per JpqlFilter");
+            }
+            built = true;
+            if (where == null) {
+                throw new IllegalStateException(
+                        "JpqlFilterBuilder: 'where' is required — call .where(\"...\") before .build()");
+            }
+            checkDataLoaderPresent(filter);
+
+            JpqlFilterCondition model = metadata.create(JpqlFilterCondition.class);
+            model.setParameterClass(parameterClass.getName());
+            model.setParameterName(parameterName != null
+                    ? parameterName
+                    : jpqlFilterSupport.generateParameterName(null, parameterClass.getSimpleName()));
+            model.setWhere(where);
+            model.setJoin(join);
+            model.setHasInExpression(hasInExpression);
+            model.setLabel(label);
+            model.setValueComponent(metadata.create(FilterValueComponent.class));
+
+            FilterConverter<JpqlFilter<V>, JpqlFilterCondition> converter =
+                    (FilterConverter<JpqlFilter<V>, JpqlFilterCondition>) filterComponents
+                            .getConverterByModelClass(JpqlFilterCondition.class, filter);
+            JpqlFilter<V> jpqlFilter = converter.convertToComponent(model);
+            if (defaultValue != null) {
+                jpqlFilter.setValue(defaultValue);
+            }
+            return jpqlFilter;
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // GroupFilterBuilder
+    // -------------------------------------------------------------------------
+
+    /**
+     * Fluent builder for {@link GroupFilter}.
+     */
+    public static class GroupFilterBuilder {
+
+        protected final GenericFilter filter;
+        protected final UiComponents uiComponents;
+
+        protected LogicalFilterComponent.Operation operation = LogicalFilterComponent.Operation.AND;
+        protected final List<FilterComponent> components = new ArrayList<>();
+        protected boolean built = false;
+
+        protected GroupFilterBuilder(GenericFilter filter, UiComponents uiComponents) {
+            this.filter = filter;
+            this.uiComponents = uiComponents;
+        }
+
+        /**
+         * Sets the logical operation ({@code AND} or {@code OR}). Defaults to {@code AND}.
+         */
+        public GroupFilterBuilder operation(LogicalFilterComponent.Operation operation) {
+            checkNotNullArgument(operation, "operation must not be null");
+            this.operation = operation;
+            return this;
+        }
+
+        /**
+         * Adds a filter component to this group.
+         */
+        public GroupFilterBuilder add(FilterComponent filterComponent) {
+            checkNotNullArgument(filterComponent, "filterComponent must not be null");
+            components.add(filterComponent);
+            return this;
+        }
+
+        /**
+         * Adds several filter components to this group at once.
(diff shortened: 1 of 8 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+    protected final GenericFilter filter;
+    protected final UiComponents uiComponents;
+    protected final SingleFilterSupport singleFilterSupport;
+
+    FilterComponentBuilder(GenericFilter filter, UiComponents uiComponents,
+                           SingleFilterSupport singleFilterSupport) {
+        this.filter = filter;
+        this.uiComponents = uiComponents;
+        this.singleFilterSupport = singleFilterSupport;
+    }
```
KremnevDmitry: What should I do if I want to override the builder and create my own subclass?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -632,6 +636,31 @@ public void refreshCurrentConfiguration() {
...
+     * The builder handles all mandatory initialisation steps that the XML loader performs
+     * automatically: {@code setConditionModificationDelegated(true)}, {@code setDataLoader(...)},
+     * and the correct ordering of property / operation / value assignments.
+     *
+     * @return a new {@code FilterComponentBuilder} instance
+     */
+    @Experimental
+    public FilterComponentBuilder componentBuilder() {
+        return new FilterComponentBuilder(this, uiComponents, singleFilterSupport);
+    }
```
KremnevDmitry: `componentBuilder()` is a too generic name, I guess. Can I create a button?)

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+import java.util.List;
+
+import static io.jmix.core.common.util.Preconditions.checkNotNullArgument;
+
+/**
+ * A factory for creating filter components ({@link PropertyFilter}, {@link JpqlFilter},
+ * {@link GroupFilter}) that are pre-configured for use inside a specific {@link GenericFilter}.
+ * <p>
+ * All builders returned by this factory automatically perform the initialisation steps
+ * that the XML loader ({@code GenericFilterLoader}) does implicitly:
```
KremnevDmitry: # Important
All JavaDocs says that the builder is preforming the initialization steps that the XML load does implicitly. But what is we change the loading logic? We should to rewrite this builder too?

This has already happened: for a void `JpqlFilter` the converter always creates a
value component (Yes/No) and sets a `parameterName`, while the builder skips both
for `Void` — so such a filter never applies its WHERE condition.

For example the reference path (`GenericFilterAddConditionAction`) doesn't duplicate anything:
it resolves the converter via `FilterComponents.getConverterByModelClass()` and
calls `convertToComponent(model)`. The builders should do the same — build the
condition model from the fluent parameters and delegate to the converter (single
source of truth).

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+     *
+     * @param parameterClass the Java class of the query parameter
+     * @param <V>            the value type
+     * @return a new {@link JpqlFilterBuilder}
+     */
+    public <V> JpqlFilterBuilder<V> jpqlFilter(Class<V> parameterClass) {
+        checkNotNullArgument(parameterClass, "parameterClass must not be null");
+        //noinspection unchecked
+        return new JpqlFilterBuilder<>(filter, uiComponents, (Class<?>) parameterClass, singleFilterSupport);
+    }
```
KremnevDmitry, line 129: Image

Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+     *
+     * @return a new {@link GroupFilterBuilder}
+     */
+    public GroupFilterBuilder groupFilter() {
+        return new GroupFilterBuilder(filter, uiComponents);
+    }
+
+    // -------------------------------------------------------------------------
+    // PropertyFilterBuilder
+    // -------------------------------------------------------------------------
```
KremnevDmitry: Why do we need this comment?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+
+    /**
+     * Fluent builder for {@link PropertyFilter}.
+     * <p>
+     * All mandatory initialisation is handled inside {@link #build()}: the correct call
+     * order ({@code setDataLoader → setProperty → setOperation → setValue}) is enforced
+     * internally, so the caller is free to set these in any order via the fluent methods.
```
KremnevDmitry: > the correct call
     * order ({@code setDataLoader → setProperty → setOperation → setValue}) is enforced
     * internally, so the caller is free to set these in any order via the fluent methods.

Im not sure that this is a mandatory information for the end-API-user. Its looks like implementation details, moreover it was mentioned in the builder JavaDoc for `FilterComponentBuilder` class

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+
+        protected String property;
+        protected PropertyFilter.Operation operation;
+        protected V defaultValue;
+        protected String label;
+        protected boolean operationEditable = false;
+        protected boolean operationTextVisible = true;
+        protected boolean built = false;
+
+        PropertyFilterBuilder(GenericFilter filter, UiComponents uiComponents) {
```
KremnevDmitry: package-private again?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+         * <p>
+         * Enforces the required call order:
+         * <ol>
+         *   <li>{@code setConditionModificationDelegated(true)}</li>
+         *   <li>{@code setDataLoader(...)}</li>
+         *   <li>{@code setProperty(...)}</li>
+         *   <li>{@code setParameterName(...)} — auto-generated via {@link PropertyConditionUtils}</li>
+         *   <li>{@code setOperation(...)}</li>
+         *   <li>{@code setValue(...)} (only if a default value was provided)</li>
+         * </ol>
```
KremnevDmitry: Again about internal implementation and API call order?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+            }
+            // 7. generate and attach value component (mirrors JpqlFilterConverter.convertToComponent())
+            if (parameterClass != Void.class && filter.getDataLoader() != null) {
+                //noinspection unchecked
+                HasValueAndElement<?, V> vc = singleFilterSupport.generateValueComponent(
+                        filter.getDataLoader().getContainer().getEntityMetaClass(),
+                        hasInExpression,
+                        (Class<?>) parameterClass);
+                jf.setValueComponent(vc);
+            }
```
KremnevDmitry: By this code, the value component will be never created for the `Void` parameterClass. 
For `Void` parameter class parameterName and valueComponent is required.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+            if (defaultValue != null) {
+                pf.setValue(defaultValue);
+            }
+            return pf;
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // JpqlFilterBuilder
+    // -------------------------------------------------------------------------
```
KremnevDmitry: Again

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+                            parameterName, e);
+                }
+            }
+            return jf;
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // GroupFilterBuilder
+    // -------------------------------------------------------------------------
```
KremnevDmitry: Again

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+            // 2. data loader
+            if (filter.getDataLoader() != null) {
+                jf.setDataLoader(filter.getDataLoader());
+            }
+            // 3. parameterName before parameterClass (affects where substitution)
+            if (parameterName != null) {
+                jf.setParameterName(parameterName);
+            }
+            // 4. parameterClass — one-time setter, must come before setCondition
+            jf.setParameterClass((Class) parameterClass);
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+            if (hasInExpression) {
+                jf.setHasInExpression(true);
+            }
+            // 7. generate and attach value component (mirrors JpqlFilterConverter.convertToComponent())
+            if (parameterClass != Void.class && filter.getDataLoader() != null) {
+                //noinspection unchecked
+                HasValueAndElement<?, V> vc = singleFilterSupport.generateValueComponent(
+                        filter.getDataLoader().getContainer().getEntityMetaClass(),
+                        hasInExpression,
+                        (Class<?>) parameterClass);
```
KremnevDmitry: Image

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+                        "PropertyFilterBuilder: 'operation' is required — call .operation(...) before .build()");
+            }
+
+            PropertyFilter<V> pf = uiComponents.create(PropertyFilter.class);
+            // 1. delegate condition modification — must be first
+            pf.setConditionModificationDelegated(true);
+            // 2. data loader before property so that initOperationSelectorActions runs correctly
+            if (filter.getDataLoader() != null) {
+                pf.setDataLoader(filter.getDataLoader());
+            }
```
KremnevDmitry: Why we are testing on `null` our `dataLoader`? As I know, the filter component can't be existed without `dataLoader` as soon as a `genericFilter`. In the loaders the `dataLoader` of nested filter components of `genericFilter` is set without any null-checking.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+
+        protected String where;
+        protected String join;
+        protected String parameterName;
+        protected V defaultValue;
+        protected String label;
+        protected boolean hasInExpression;
+        protected boolean built = false;
+
+        JpqlFilterBuilder(GenericFilter filter, UiComponents uiComponents, Class<?> parameterClass,
```
KremnevDmitry: package-private

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+    public static class GroupFilterBuilder {
+
+        protected final GenericFilter filter;
+        protected final UiComponents uiComponents;
+
+        protected LogicalFilterComponent.Operation operation = LogicalFilterComponent.Operation.AND;
+        protected final List<FilterComponent> components = new ArrayList<>();
+        protected boolean built = false;
+
+        GroupFilterBuilder(GenericFilter filter, UiComponents uiComponents) {
```
KremnevDmitry: package-private

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+     *
+     * @return the newly created and registered {@link RunTimeConfiguration}
+     * @throws IllegalStateException if {@code id} was not set
+     */
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    public RunTimeConfiguration buildAndRegister() {
+        if (id == null) {
+            throw new IllegalStateException(
+                    "RunTimeConfigurationBuilder: 'id' is required — call .id(\"...\") before .buildAndRegister()");
+        }
```
KremnevDmitry, line 216: What if the configuration with the same id is already exist?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
KremnevDmitry: P.S. we should cover this case with tests

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,515 @@
...
+            // the Vaadin tree. Use RunTimeConfigurationBuilder.add(fc, defaultValue) to persist
+            // the default for reset support even when setValue fails.
+            if (defaultValue != null) {
+                try {
+                    jf.setValue(defaultValue);
+                } catch (RuntimeException e) {
+                    log.debug("Could not pre-set default value on JpqlFilter '{}'",
+                            parameterName, e);
+                }
+            }
```
KremnevDmitry: I don't understand the point about unavailability to set the default value....
setValue can't be failed when the component is not yet in the Vaadin tree (!)
This is a server-side UI, it should be available.

Think about it one more time.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+
+    protected String id;
+    protected String name;
+    protected LogicalFilterComponent.Operation operation = LogicalFilterComponent.Operation.AND;
+    protected boolean makeDefault = false;
+    protected boolean allowDeletion = false;
+
+    protected final List<ComponentEntry> entries = new ArrayList<>();
+
+    RunTimeConfigurationBuilder(GenericFilter filter, UiComponents uiComponents) {
```
KremnevDmitry: package-private

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * Adds a filter component using the component's current value (if any) as the default.
+     *
+     * @param filterComponent filter component to add
+     */
+    public RunTimeConfigurationBuilder add(FilterComponent filterComponent) {
+        checkNotNullArgument(filterComponent, "filterComponent must not be null");
+        entries.add(new ComponentEntry(filterComponent, null, false));
+        return this;
+    }
```
KremnevDmitry, line 121: Maybe it is make sense to add vararg overloading... for GroupFilterBuilder too

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Marks this configuration as the default (currently active) configuration.
+     */
+    public RunTimeConfigurationBuilder asDefault() {
+        this.makeDefault = true;
+        return this;
+    }
```
KremnevDmitry, line 163: default != currently active. Default configuration is a configuration what will be set for the `genericFilter` by default after opening view. `default` is a persistent marker. See other usage of `default` word in the `genericFilter`

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Made constructors/fields protected and factory methods overridable — subclassing works now.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
fractal3000: Renamed to filterComponentBuilder()

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Done. The single-component builders now build a condition model and delegate to the FilterConverter (resolved via FilterComponents.getConverterByModelClass), exactly like GenericFilterAddConditionAction — no duplicated init. This fixes the void filter: the converter creates the checkbox value component and parameter name, so the WHERE is applied.

jpqlFilter() now returns JpqlFilterBuilder<Boolean> (a checkbox toggle; defaultValue(true) activates it), and parameterName is optional/auto-generated. Note both jpqlFilter() and jpqlFilter(Boolean.class) yield JpqlFilterBuilder<Boolean> — the type parameter is just the value type; they differ by the internal parameterClass (Void = toggle the WHERE, Boolean = bind the value), chosen by which method you call.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — removed the redundant suppression and cast.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed — there is no manual init order anymore; the converter handles initialisation.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — protected now

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed via delegation — the converter now creates the value component and parameter name for the void filter. Covered by tests (void filter applies its WHERE when active, omits it when inactive).

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — this manual init is gone; the converter sets the parameter class.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — value-component generation is now delegated to the converter; this code is gone.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — the builder now requires the filter to have a DataLoader (fail-fast) instead of null-checking it. Same applied to RunTimeConfigurationBuilder.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — protected now.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Fixed — protected now.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Fixed — buildAndRegister() now throws IllegalStateException if a configuration with the same id is already registered. Covered by a test.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Agreed — removed the defensive try/catch; setValue is called directly.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Fixed — protected now.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Added a vararg to both RunTimeConfigurationBuilder and GroupFilterBuilder. I named it addAll(FilterComponent...) rather than overloading add: in idiomatic Java an add(FilterComponent...) overload actually resolves fine, but a two-argument call would mis-dispatch under Groovy's dynamic dispatch (erasure makes it pick add(SingleFilterComponentBase, V)), and having two same-arity-matching add overloads is confusing regardless (Effective Java Item 52). Can switch to an add overload if you prefer.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Agreed — "default" is the persistent per-user marker (set via the genericFilter_makeDefault action) for the configuration activated on view open. The builder method only activates the configuration immediately, so it was misnamed. Renamed it to makeCurrent() and clarified the distinction in the Javadoc.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,492 @@
...
+    protected final Metadata metadata;
+    protected final FilterComponents filterComponents;
+    protected final JpqlFilterSupport jpqlFilterSupport;
+
+    protected FilterComponentBuilder(GenericFilter filter, ApplicationContext applicationContext) {
+        this.filter = filter;
+        this.uiComponents = applicationContext.getBean(UiComponents.class);
+        this.metadata = applicationContext.getBean(Metadata.class);
+        this.filterComponents = applicationContext.getBean(FilterComponents.class);
+        this.jpqlFilterSupport = applicationContext.getBean(JpqlFilterSupport.class);
```
KremnevDmitry: It make sense to extract dependencies initialization in a separate method.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
KremnevDmitry: Not removed, as I can see

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,492 @@
...
+    public GroupFilterBuilder groupFilter() {
+        return new GroupFilterBuilder(filter, uiComponents);
+    }
+
+    protected static void checkDataLoaderPresent(GenericFilter filter) {
+        if (filter.getDataLoader() == null) {
+            throw new IllegalStateException(
+                    "GenericFilter has no DataLoader. Filter components cannot be built without a DataLoader; " +
+                            "set it before using the builder (the 'dataLoader' attribute in XML, or " +
+                            "GenericFilter.setDataLoader(...)).");
```
KremnevDmitry: To avoid hadcoding components name, class, interfaces names we use the following template:

E.g.:
```java
            String message = String.format(
                    "%s doesn't have data binding. Set either dataContainer and property or metaClass attribute.",
                    resultComponent.getClass().getSimpleName()
            );
```

This approach is convenient when renaming classes, interfaces, and other types.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,492 @@
...
+        protected V defaultValue;
+        protected String label;
+        protected boolean operationEditable = false;
+        protected boolean operationTextVisible = true;
+        protected boolean built = false;
+
+        protected PropertyFilterBuilder(GenericFilter filter, Metadata metadata, FilterComponents filterComponents) {
+            this.filter = filter;
+            this.metadata = metadata;
+            this.filterComponents = filterComponents;
```
KremnevDmitry: Whats about passing `applicationContext` and creating a separate method to autowire dependencies? It will be consistent, I guess

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
KremnevDmitry: I don't think so :)

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
KremnevDmitry: And here :)

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -594,15 +599,82 @@ public Configuration getCurrentConfiguration() {
...
+     * followed by {@link #setCurrentConfiguration(Configuration)}, which fails silently when
+     * the configuration is not yet registered at the time {@code setCurrentConfiguration} is
+     * invoked.
+     *
+     * @param configuration the configuration to register and activate
+     */
+    public void addAndSetCurrentConfiguration(Configuration configuration) {
+        addConfiguration(configuration);
+        setCurrentConfiguration(configuration);
+    }
```
KremnevDmitry: This method doesn't add any new semantics, but merely combines two existing public methods. This increases the API surface area and creates a precedent for other combinations of operations. I'd prefer to leave the explicit call to `addConfiguration()` and `setCurrentConfiguration()` to client code, as per their contracts.

We are not interested in creating a contract larger than we already have, as we will subsequently have to bear responsibility for it.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,280 @@
...
+        if (makeCurrent) {
+            filter.addAndSetCurrentConfiguration(config);
+        } else {
+            filter.addConfiguration(config);
+        }
+
+        return config;
+    }
+
+    // -------------------------------------------------------------------------
```
KremnevDmitry: no comments

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Sorry, removed.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Removed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Agree. The initialization was extracted.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Done — the message now uses filter.getClass().getSimpleName() instead of the hardcoded name (and is shortened to a single line).

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: Done — the nested builders now receive the ApplicationContext and resolve their dependencies in an autowireDependencies(ApplicationContext) method, consistent with the outer factory.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
fractal3000: Agreed — removed addAndSetCurrentConfiguration(). buildAndRegister() now calls addConfiguration() and, when makeCurrent() was requested, setCurrentConfiguration() directly.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+        checkNotNullArgument(filterComponent, "filterComponent must not be null");
+        entries.add(new ComponentEntry(filterComponent, null, false));
+        return this;
+    }
+
+    /**
+     * Adds several filter components at once, each using its current value (if any) as the default.
+     * <p>
+     * Named {@code addAll} rather than overloading {@code add} to avoid ambiguity with
+     * {@link #add(SingleFilterComponentBase, Object)} when two filter components are passed.
```
KremnevDmitry: This is a redundant information, I guess

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,487 @@
...
+            checkNotNullArgument(filterComponent, "filterComponent must not be null");
+            components.add(filterComponent);
+            return this;
+        }
+
+        /**
+         * Adds several filter components to this group at once.
+         * <p>
+         * Named {@code addAll} (rather than overloading {@code add}) for consistency with
+         * {@link RunTimeConfigurationBuilder#addAll(FilterComponent...)}.
```
KremnevDmitry: And this

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+
+        RunTimeConfiguration config = new RunTimeConfiguration(id, root, filter);
+        config.setName(name);
+
+        for (ComponentEntry entry : entries) {
+            FilterComponent fc = entry.filterComponent;
+
+            if (entry.overrideDefault && fc instanceof SingleFilterComponentBase<?> sfc) {
+                ((SingleFilterComponentBase) sfc).setValue(entry.defaultValue);
+            }
```
KremnevDmitry: Redundant casting after pattern matching, if we remove `<?>` for it

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
```diff
@@ -0,0 +1,487 @@
...
+         *         if the owning filter has no DataLoader, or if this builder instance has
+         *         already been used
+         */
+        @SuppressWarnings("unchecked")
+        public PropertyFilter<V> build() {
+            if (built) {
+                throw new IllegalStateException(
+                        "PropertyFilterBuilder.build() must not be called more than once; create a new instance per PropertyFilter");
+            }
+            built = true;
```
KremnevDmitry: `built` looks redundant

BTW, if an API user makes a mistake and forgets to set a property and then tries to build a component, the build will be corrupted. If we want to keep `built` frag it should be set after generation success

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Removed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/RunTimeConfigurationBuilder.java
fractal3000: Done

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/FilterComponentBuilder.java
fractal3000: *built* is now set only after a successful build, so a failed-validation build no longer burns the builder. Kept the guard (it blocks reusing a builder — for GroupFilterBuilder that would re-parent the same children into a second group) and added it to RunTimeConfigurationBuilder for consistency.

KremnevDmitry (approved):
Don't forget to create an issue for Docs

Issue #5396, opened 2026-06-23
Make Jasperreports dependency optional


Gavrilov-Ivan:
- Make jasperreports-related dependencies compileOnly.
- Jasperreports dependencies should be added explicitly on the project side if necessary

Gavrilov-Ivan:
To enable support of jasperreports templates porject should explicitly declare the following dependencies:

- `implementation 'net.sf.jasperreports:jasperreports'`
- `implementation 'net.sf.jasperreports:jasperreports-fonts'`
- `implementation 'net.sf.jasperreports:jasperreports-functions'`

PR #5397, opened 2026-06-23
Make Jasperreports dependency optional

Gavrilov-Ivan:
See #5396

### DIFF
--- a/jmix-reports/reports/src/main/java/io/jmix/reports/yarg/formatters/factory/DefaultFormatterFactory.java
+++ b/jmix-reports/reports/src/main/java/io/jmix/reports/yarg/formatters/factory/DefaultFormatterFactory.java
@@ -29,6 +29,8 @@ import io.jmix.reports.yarg.structure.BandData;
 import io.jmix.reports.yarg.structure.ReportTemplate;
 import io.jmix.reports.yarg.util.groovy.Scripting;
 import org.jspecify.annotations.NullMarked;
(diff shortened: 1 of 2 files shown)


jmix-reports/reports/src/main/java/io/jmix/reports/yarg/formatters/factory/DefaultFormatterFactory.java
```diff
@@ -185,6 +194,28 @@ protected void setDefaultInlinersProvider() {
         inlinersProvider = new DefaultInlinersProvider();
     }
 
+    /**
+     * Checks whether the optional JasperReports library is present on the classpath.
+     */
+    protected boolean isJasperReportsAvailable() {
+        try {
+            Class.forName(JASPER_REPORTS_MARKER_CLASS, false, DefaultFormatterFactory.class.getClassLoader());
```
knstvk: What about using `org.springframework.util.ClassUtils#isPresent`?
I believe it's safer.

knstvk (changes requested):

jmix-reports/reports/src/main/java/io/jmix/reports/yarg/formatters/factory/DefaultFormatterFactory.java
Gavrilov-Ivan: Done

knstvk (approved):

PR #5403, opened 2026-06-24
GenericFilter: activating a configuration in onInit breaks subsequent configuration switching  #5400

### LINKED ISSUE #5400: GenericFilter: activating a configuration in `onInit` breaks subsequent configuration switching
## Description

A `GenericFilter` preserves the data loader's own condition and combines it with the active
configuration. If a configuration is made current **during the view's `onInit`**, that own condition
is recorded incorrectly — it already includes the activated configuration. Switching to another
configuration afterwards then combines both configurations instead of replacing, so the loader gets a
contradictory condition and the grid usually comes back empty.

This happens with the plain base API (`GenericFilter.setCurrentConfiguration(...)` in `onInit`) and
with the new builder's `makeCurrent()`; a standalone `GroupFilter` is affected the same way. It is a
regression from jmix#2406 and was surfaced by jmix#5351, but it is independent of that API.
Declarative `default="true"` configurations are not affected.

## Reproduction

A list view with a `GenericFilter` and two configurations on the same attribute (e.g. `status` =
`OPEN` / `status` = `CLOSED`). In `onInit`, make the first one current
(`setCurrentConfiguration(...)`). Open the view, then switch the filter to the second configuration.

- **Expected:** the grid shows rows of the second configuration.
- **Actual:** the grid is empty; the loader condition is `status = OPEN AND status = CLOSED`.

## Impact

Any view that activates a filter configuration programmatically in `onInit` (directly or via the
builder) shows wrong/empty data after the user switches configurations — a silent data-correctness
issue in a core, widely-used component.

## Expected behavior

Activating a configuration in `onInit` must not affect later switching: switching always applies only
the target configuration's condition (on top of the loader's own condition). The filtered result on
open must also be correct.

fractal3000:
#5400

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
@@ -115,13 +115,9 @@ public class GenericFilter extends Composite<JmixDetails>
     protected String applyShortcut;
     protected int propertyHierarchyDepth;
     protected DataLoader dataLoader;
-    /**
-     * The data loader's own condition (without any filter contribution), captured lazily and once
-     * before the filter first updates the loader condition, then combined (AND) with the current
-     * configuration on every change.
-     */
     protected Condition initialDataLoaderCondition;
     protected boolean initialDataLoaderConditionInitialized;
+    protected Condition lastConditionSetByFilter;
     protected Predicate<MetaPropertyPath> propertyFiltersPredicate;
 
     protected VerticalLayout contentWrapper;
@@ -368,11 +364,6 @@ public class GenericFilter extends Composite<JmixDetails>
 
     /**
      * Sets a {@link DataLoader} related to the filter.
-     * <p>
-     * The data loader's own condition is preserved and combined (AND) with the current
-     * configuration's condition. It is captured once, before the filter first applies a
-     * configuration; therefore any base loader condition must be set (in XML or in {@code onInit})
-     * before a configuration is activated.
      *
      * @param dataLoader a {@link DataLoader} to set
      */
@@ -392,7 +383,7 @@ public class GenericFilter extends Composite<JmixDetails>
      * in {@link #updateDataLoaderCondition()} before the first filter contribution. Retained for
      * backward compatibility.
      */
-    @Deprecated
+    @Deprecated(since = "3.0", forRemoval = true)
     protected void updateDataLoaderInitialCondition(@Nullable Condition condition) {
         this.initialDataLoaderCondition = copy(condition);
         this.initialDataLoaderConditionInitialized = true;
@@ -847,10 +838,12 @@ public class GenericFilter extends Composite<JmixDetails>
 
     protected void updateDataLoaderCondition() {
         if (dataLoader != null) {
-            if (!initialDataLoaderConditionInitialized) {
-                // Capture the data loader's own condition once, before the first filter contribution,
-                // so it is never polluted by a configuration activated during onInit.
-                initialDataLoaderCondition = copy(dataLoader.getCondition());
+            Condition currentCondition = dataLoader.getCondition();
+            // Re-capture the loader's own condition only when it was replaced externally (a different
+            // object than the filter's last output); the filter never adopts its own output.
+            if (!initialDataLoaderConditionInitialized
+                    || (lastConditionSetByFilter != null && currentCondition != lastConditionSetByFilter)) {
+                initialDataLoaderCondition = copy(currentCondition);
                 initialDataLoaderConditionInitialized = true;
             }
             LogicalFilterComponent<?> logicalFilterComponent = getCurrentConfiguration().getRootLogicalFilterComponent();
@@ -869,6 +862,7 @@ public class GenericFilter extends Composite<JmixDetails>
             }
 
             dataLoader.setCondition(resultCondition);
+            lastConditionSetByFilter = resultCondition;
         }
     }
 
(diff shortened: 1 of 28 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -369,15 +381,21 @@ public void setDataLoader(DataLoader dataLoader) {
...
         rootLogicalFilterComponent.setDataLoader(dataLoader);
         rootLogicalFilterComponent.setAutoApply(autoApply);
     }
 
+    /**
+     * @deprecated no longer used internally; the initial data loader condition is now captured lazily
+     * in {@link #updateDataLoaderCondition()} before the first filter contribution. Retained for
+     * backward compatibility.
+     */
+    @Deprecated
```
glebfox: since?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
```diff
@@ -195,15 +195,28 @@ public void setDataLoader(DataLoader dataLoader) {
         updateSummaryText();
     }
 
+    /**
+     * @deprecated no longer used internally; the initial data loader condition is now captured lazily
+     * in {@link #updateDataLoaderCondition()} before the first filter contribution. Retained for
+     * backward compatibility.
+     */
+    @Deprecated
```
glebfox: since?

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilterUtils.java
```diff
@@ -23,6 +23,11 @@
 @Internal
 public class GroupFilterUtils {
 
+    /**
+     * @deprecated no longer used internally; {@link GroupFilter} now captures the initial data loader
+     * condition lazily. Retained for backward compatibility.
+     */
+    @Deprecated
```
glebfox: since?

glebfox (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/GenericFilterLoader.java
```diff
@@ -87,14 +87,6 @@ protected void loadDataLoader(GenericFilter component, Element element) {
...
                     component.setDataLoader(dataLoader);
-
-                    getContext().addInitTask(new AbstractInitTask() {
-                        @Override
-                        public void execute(Context context) {
-                            FilterUtils.updateDataLoaderInitialCondition(resultComponent,
-                                    dataLoader.getCondition());
-                        }
-                    });
                 });
```
KremnevDmitry, line 90: Breaking changes is not documented. Was this approved?

The base loader condition must now be set *before* a configuration is activated. Previously the loader `init` task ran *after* `onInit`, so this captured `myBaseCondition`:
```java
// in onInit
genericFilter.setCurrentConfiguration(c1);
ordersDl.setCondition(myBaseCondition);   // set AFTER activation
```

Now the first `updateDataLoaderCondition()` snapshots the baseline without it, and it's lost on the next switch — a silent regression for anyone relying on the old timing.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilterUtils.java
fractal3000: Fixed

jmix-flowui/flowui/src/main/java/io/jmix/flowui/xml/layout/loader/component/GenericFilterLoader.java
fractal3000: Thanks, valid point. Removed the regression: the filter now adopts the loader's condition as its baseline if it was replaced externally (a different object, not the filter's own output). So setting the base condition after activating a configuration in onInit works again (and the same for a standalone GroupFilter), while configuration switching stays fixed. Covered by tests.

fractal3000: Separately: we need to remember to add a clarification to the documentation. To change the base condition, set a new object via DataLoader.setCondition(). If instead, after the first updateDataLoaderCondition (for GenericFilter or GroupFilter), you obtain a reference to the Condition from DataLoader and modify it in place without changing the reference in DataLoader, the filter ignores it: the change is not picked up and is overwritten on the next condition rebuild (the exception being a direct DataLoader.load() before the rebuild, which reads the mutated object, but the filter does not preserve it). This behavior is not changed by this PR — it was the case before as well.

KremnevDmitry (approved):

glebfox (approved):

Issue #5331, opened 2026-05-28
DataContext merge can duplicate owner collection on bidirectional graph with multiple java instances of same id


Flaurite:
### Environment
Jmix version: 2.8+

### Bug Description
Reproduces only with not saved entities.

When we have association with bidirectional reference it causes recursion to reach the same managed entity again, `DataContext#mergeList()` re-enters and writes to the same `dstList` as the outer call, producing duplicate elements in the owner's collection.
**For instance:**
- `ProductSet` with `Product` collection
 - `Product` with `ProductSet` link

### Steps To Reproduce
1. Download demo: [datacontext-demo.zip](https://github.com/user-attachments/files/28349672/datacontext-demo.zip)
2. Launch the application
3. Open "Product sets"
4. Create one, do no save
5. Create Product 1, Product 2
6. Edit Product 1, edit Product 2

### Current Behavior
After saving editions in Product 2 there are 4 rows with duplications instead of 2 rows.

### Expected Behavior
2 rows with changes we've made in editor.

knstvk:
Fixed in https://github.com/jmix-framework/jmix/pull/5446

Issue #4907, opened 2025-11-24
Entity changes are not preserved in the UI when there is a deep composition relationships


OLODiman11:
### Environment

Jmix version: 2.7.0

### Bug Description

Entity changes are not preserved in the UI when there is a deep composition relationships

### Steps To Reproduce

1) Run sample project
2) Navigate to `As`
3) Create entity `A` with name `A1`
4) Create entity `B` with name `B1`
5) Create entity `C` with name `C1`
6) Create entity `D` with name `D1`
7) Save everything
Now you should have the following relationships:
A1 > B1 > C1 > D1
8) Open edit view for `A1`
9) Open edit view for `B1`
10) Open edit view for `C1`
11) Open edit view for `D1`
12) Change its name to `D2`
13) Save all but `A1`, keep it open
14) Open edit view for `B1` again
15) Open edit view for `C1` again
16) The name of the `D` entity remains `D1` in the UI

### Current Behavior

The name of the `D` entity remains `D1` in the UI

### Expected Behavior

The name of the `D` changes to `D2` in the UI

### Notes

After step 13, only `C1` and `D1` are marked as modified in the `a-detail-view` context, but not `B1`. So when opening the `B1` view again it provides the original `B1` which does not have any changes in `C1` and `D1` .

### Sample Project

[nested-composition-bug.zip](https://github.com/user-attachments/files/23724146/nested-composition-bug.zip)

knstvk:
There is a workaround: in each intermediate detail view (except the root and the last ones), add the listener for the next level entity changes to mark the current entity instance modified.

This is shown in UI Samples: https://demo.jmix.io/ui-samples/sample/composition-2-levels?tab=TerminalDetailView.java

For the attached project, this means adding the following handlers:

```java
public class BDetailView extends StandardDetailView<B> {

    @ViewComponent
    private DataContext dataContext;

    @Subscribe(id = "csDc", target = Target.DATA_CONTAINER)
    public void onCsDcItemPropertyChange(final InstanceContainer.ItemPropertyChangeEvent<C> event) {
        B b = getEditedEntity();
        dataContext.setModified(b, true);
    }
}
```

```java
public class CDetailView extends StandardDetailView<C> {

    @ViewComponent
    private DataContext dataContext;

    @Subscribe(id = "dsDc", target = Target.DATA_CONTAINER)
    public void onDsDcItemPropertyChange(final InstanceContainer.ItemPropertyChangeEvent<D> event) {
[...]

Issue #4906, opened 2025-11-24
Entity changes are not preserved in the UI when reopening its detail view


OLODiman11:
### Environment

Jmix version: 2.7.0

### Bug Description

Entity changes are not preserved in the UI when reopening its detail view

### Steps To Reproduce

1) Run sample project
2) Navigate to `As`
3) Create entity `A` with name `A1`
4) Create entity `B` with name `B1`
5) Create entity `C` with name `C1`
6) Create entity `D` with name `D1`
7) Save everything
Now you should have the following relation:
A1 > B1 > C1 > D1
8) Open edit view for `A1`
9) Edit `B1`, change its name to `B2`
10) Click `Ok` in the `B` view (do not close `A1` view)
11) Edit `B2` again (that was previously `B1`)
12) The name remains `B1` in the UI

### Current Behavior

The name of the `B` entity remains `B1` in the UI

### Expected Behavior

The name of the `B` changes to `B2` in the UI

### Notes

As i understand it happens because of specific fetch plans

`a-detail-view` fetch plan:
```
<fetchPlan extends="_base">
	<property name="bs" fetchPlan="_base">
		<property name="cs" fetchPlan="_base"/>
	</property>
</fetchPlan>
```

`b-detail-view` fetch plan:
```
<fetchPlan extends="_base">
	<property name="cs" fetchPlan="_base">
		<property name="d" fetchPlan="_base"/>
	</property>
</fetchPlan>
```

When you open `a-detail-view` it loads all `cs` and does not load `d`
When you open `b-detail-view` it loads all `cs` and `d`
When clicking `Ok` in `b-detail-view` it merges its context to the `a-detail-view` context, but it does not mark `d` as loaded (`DataContextimpl#mergeList` calls `internalMerge` with `isRoot` set to `false` and eventially it gets to `mergeLoadedPropertiesInfo`)
Then when opening `b-detail-view` again it checks if the provided `B` (from the `a-detail-view` context) is loaded with a proper fetch plan and fails because it thinks that `d` is not loaded, so it reloads the entity loosing all the changes made to it.

### Sample Project

[fetch-plan-bug.zip](https://github.com/user-attachments/files/23723582/fetch-plan-bug.zip)

knstvk:
Fixed in https://github.com/jmix-framework/jmix/pull/5446

SergeiAksenov2:
Tested on:
Jmix version: 3.1.999-SNAPSHOT
[...]

Issue #4071, opened 2025-01-16
Editing child entity of composition entity does not fire DataContext.ChangeEvent


alexbudarov:
### Environment

Jmix version: 2.4.2

### Bug Description

DataContext.ChangeEvent isn't fired in a screen with composition collection, after the child entity is edited in a nested dialog.

### Steps To Reproduce

- Launch attached modified jmix-petclinic sample project
There are changes in Owner details screen - a yellow alert panel is shown in the screen layout as soon as data context change is detected, e.g. when user modifies something in this editor screen.
A listener for DataContext.ChangeEvent is used for detecting changes.
![image](https://github.com/user-attachments/assets/ed7af65d-6382-4ae5-9a88-7eef63e2d9e9)

- Login as joy
- Open any Owner (e.g. James) for editing
- Press Enable editing
Now test various cases of modifying the entity

1 - modify First name. 
Yellow alert about unsaved changes appears - OK

2 - reload page. Switch to Pets tab.
Create new Pet.
Yellow alert about unsaved changes appears - OK

3 - reload page. Switch to Pets tab. 
Delete one of pets.
Yellow alert about unsaved changes appears - OK

4 - reload page. Switch to Pets tab. 
Open one of Pets (e.g. Drowzee) for editing.
Modify Name, Birthdate. Press OK.

### Current Behavior
The alert does not appear (e.g. DataContext.ChangeEvent is not fired) = KO.

### Expected Behavior
The DataContext.ChangeEvent should fire, and alert should appear after editing one of pets.

### Sample Project

Modified jmix-petcilnic sample:
[jmix-petclinic.zip](https://github.com/user-attachments/files/18436663/jmix-petclinic.zip)

knstvk:
Cannot reproduce on 2.5.1. Need additional investigation why.

knstvk:
Reproduced on 2.5.2 because of fixing #4429.
A workaround is demonstrated in UI Samples: https://demo.jmix.io/ui-samples/sample/composition-info-panel

knstvk:
Fixed in https://github.com/jmix-framework/jmix/pull/5446

SergeiAksenov2:
Tested on:
Jmix version: 3.1.999-SNAPSHOT
Jmix Studio Plugin Version: 3.1.SNAPSHOT8660-261
IntelliJ version: IntelliJ IDEA 2026.1.5 RC
Successful.

PR #5446, opened 2026-07-12
DataContext: attribute-level change tracking and merge hardening

knstvk:
## Summary

This branch reworks `DataContextImpl` change tracking and merge so that unsaved
user edits are no longer lost or spuriously reported. Tracking becomes
per-attribute (with baselines) instead of per-entity, the merge now protects
dirty attributes from being overwritten, and several long-standing merge and
loaded-state defects are fixed. Targets the 3.1 line.

## What changed

- **Attribute-level change tracking.** A new internal `DataContextChangeTracker`
  records, per managed instance, which attributes the user changed and the
  baseline each change is measured against (scalar value, reference id, embedded
  dotted path, or an order-insensitive membership bag for collections).
  Reverting an attribute to its baseline un-dirties it; when the last dirty
  attribute of an entity reverts, the entity leaves `getModified()` (unless kept
  via `setModified(entity, true)`).

- **Merge conflict rule: user edits win over incoming state.** For a dirty
  attribute on an already-managed instance, a non-fresh merge skips the copy and
  a fresh merge keeps the user value but rebaselines it (un-dirtying only if the
  incoming value is now equal). Clean attributes copy as before. Reloads and
  stale copies no longer clobber unsaved edits.

- **Loaded-state correctness and durability.** The merge resets a pre-existing
  managed instance's loaded-state cache before its copy loops (recomputed from
  the unioned fetch group) instead of copying the source's cache, so a save no
  longer writes null over a loaded value. A value that becomes present because it
  was set or merge-installed is now marked loaded, and set/fetched loaded-state
  stays durable across a fresh narrower merge.

- **Event model split.** Merge-triggered `ChangeEvent`s are deferred and fired
  when the outermost merge unwinds; an edit injected by a listener during a merge
  is now recognized and tracked instead of being swallowed. Replaces the old
  `disableListeners` boolean.

- **Composition and child→parent saves.** `DataContextInternal.mergeFromChild`
  lets a child context's edits win over the parent and unions the child's dirty
  attributes into the parent's tracker. A child save marks the composition owner
  chain modified (reopen protection only, not persisted).

- **Identity map on merge-transplant.** A merge that brings a
  reference/collection loaded onto an instance the context tracked unfetched now
  merges the reached nodes into the context instead of transplanting source
  copies.

## New public API

- `DataContext.getModifiedAttributes(Object entity)` — default method returning
  the names of the entity's modified, not-yet-saved attributes.
- Diagnostics log category `io.jmix.flowui.datacontext.diagnostics` (DEBUG):
  reports every dirty/revert/rebaseline transition, merge-skips of dirty
  attributes, and a once-per-instance notice when a read-only `NoopDataContext`
  receives a `merge()` call.

`isModified`, `getModified`, `setModified`, `clearChanges`, and `hasChanges`
keep their signatures; `isModified` membership is now exact.

## Fixed issues

- #5331 — merge duplicates the owner collection on a bidirectional graph
- #4906 — changes lost when reopening a nested detail view
- #4907 — changes not preserved with deep composition
- #4071 — composition child edit fires no `ChangeEvent`
- #1258 — changes made inside a listener during merge are dropped
- #1409 — detail view always suggests to save after an edit-and-revert
- #532 — adds diagnostics for unwanted "unsaved changes"

## Tests

New `data_components` suites cover the tracker, the merge conflict policy,
diagnostics, and loaded-state behavior, plus `StandardDetailView` cases for the
unsaved-changes prompt. A `DataContextInvariantFuzzTest` generates random
load/merge/edit scenarios and enforces a tiered regression gate over the merge
invariants (slow test, run with `-PincludeSlowTests=true`).

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
@@ -643,13 +643,15 @@ public class DataContextImpl implements DataContextInternal {
     }
 
     /**
-     * Whether this is a root non-fresh merge onto a pre-existing managed instance whose loaded-state
-     * info is the caching kind. Such an instance carries source-relative negatives cached at earlier
-     * merges (its creation merge caches every attribute outside its fetch plan as unloaded), and its
-     * fetch group was just unioned with the source's in {@link #mergeSystemState}.
+     * Whether this is a root non-fresh or a fresh merge onto a pre-existing managed instance whose
+     * loaded-state info is the caching kind. Such an instance carries source-relative negatives cached
+     * at earlier merges, and its fetch group was just unioned with the source's in {@link #mergeSystemState}.
+     * In both cases the loaded-state cache is recomputed from that unioned fetch group (blanked before the
+     * copy loops in {@link #resetLoadedInfoBeforeCopy}) instead of being replaced with the source's, so a
+     * fetched attribute the narrower source omits is not reverted to unloaded by a copied source negative.
      */
     protected boolean isColdResetTarget(Object dstEntity, boolean isRoot, MergeOptions options, boolean dstExisted) {
-        return isRoot && !options.isFresh() && dstExisted
+        return (isRoot || options.isFresh()) && dstExisted
                 && EntitySystemAccess.getEntityEntry(dstEntity).getLoadedPropertiesInfo() instanceof CachingLoadedPropertiesInfo;
     }
 
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
@@ -102,7 +102,18 @@ public class DataContextImpl implements DataContextInternal {
 
     protected PropertyChangeListener propertyChangeListener = new PropertyChangeListener();
 
-    protected boolean disableListeners;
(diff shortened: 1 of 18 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
```diff
@@ -799,10 +1222,158 @@ public Collection<Object> isolate(List<Object> entities) {
...
+            }
+        }
+        overridingAttributes = childDirtyAttributes; // consulted by the merge-rule checks
+        Object result;
+        try {
+            result = merge(entity);
+        } finally {
+            overridingAttributes = Set.of();
+        }
+        for (String attribute : childDirtyAttributes) {
```
Gavrilov-Ivan, line 1317: Naming: incoming set is 'childDirtyAttributes' but later there is a check if each attribute is actually dirty. So original 'childDirtyAttributes' is not necessarily 'dirty' in general.
Or\and add javadoc about it.

Gavrilov-Ivan (changes requested):
Add more javadocs to `DataContextImpl`

jmix-flowui/flowui/src/main/java/io/jmix/flowui/model/impl/DataContextImpl.java
knstvk: Not really. The check below tests if the attribute with this name is dirty in the instance tracked by th current context.

Gavrilov-Ivan (approved):

PR #5459, opened 2026-07-15
LookupField annotation

### LINKED ISSUE #5476: Entity annotation to control reference lookup strategy in UI
Create a declarative, data-model-level control over the UI component Jmix generates for entity reference attributes: an EntityComboBox (dropdown) or an EntityPicker (view lookup). 

It should be used in generic mechanisms like genericFilter, input dialog, dataGrid inline editor, runtime view templates.

knstvk:
Declarative, data-model-level control over the UI component Jmix generates for entity **reference** attributes: an `EntityComboBox` (dropdown) or an `EntityPicker` (view lookup), the actions on it, and optional opt-in lazy loading of dropdown items. Restores the CUBA `@Lookup` capability dropped in the Jmix migration and extends it with lazy loading that works for any data store, including non-persistent DTO entities. Applications using no annotation are unaffected — the no-annotation code paths are unchanged.

### What's included

- **Core API** (`jmix-core`, `io.jmix.core.entity.annotation`): `@LookupField` (targetable at a class or a reference field), `@LookupItemsQuery` (explicit JPQL `query` or `byInstanceName`), and the `LookupType` enum (`DROPDOWN` / `VIEW`).
- **Shared resolution** (`jmix-flowui`): `LookupFieldSupport` + `EffectiveLookupConfig` — one source of truth for precedence (field annotation → `entity-field-*` properties → class annotation) and validation. Consumed by both generators.
- **Live component generation**: `EntityFieldCreationSupport` builds the picker/combobox and its lazy items callbacks; the items-fetch logic lives in the shared `ItemsFetchCallbackSupport`.
- **Runtime view templates** (builds on #5225 / #5226): template-generated detail views now honor the annotation via `ComponentXmlFactory`. Eager dropdowns render as true eager comboboxes backed by a generated data container; explicit-query and `byInstanceName` dropdowns render as lazy `<itemsQuery>`.
- **XML**: `byInstanceName` is now expressible directly in a hand-written `<entityComboBox><itemsQuery>`, with design-time Studio metadata for the attribute.

### Behavior

- **Items loading**: eager (small dictionaries, client-side filtering), explicit paged JPQL `query` with `:searchString`, or `byInstanceName` (query conditions over the instance-name attributes; works for any store). Items load through `DataManager`, so row-level security applies.
- **Never throws**: invalid configuration logs a warning and degrades (lazy → eager → view lookup) so a field always renders.

### Compatibility

Breaking on the protected/internal surface only (all within `*Support` extension points): `EntityFieldCreationSupport` precedence internals moved to `LookupFieldSupport`; `DataLoaderSupport.getSearchString` / `VALUE_PARAMETER` removed; instance-name helpers moved to `ItemsFetchCallbackSupport`; several bean constructors gained parameters. No released public API is affected.

### Testing

Spock integration tests (in-memory HSQL) over resolver precedence/validation, live-component generation, the XML `byInstanceName` loader, template XML generation with degradation, and an end-to-end template render.

### Follow-ups (non-blocking)

- Create a Studio issue to use `@LookupField` when generating views.
- Support for `@LookupField` meta-annotation in Dynamic Model. May require changes in the way the metadata store nested annotations like `@LookupItemsQuery` (currently they are stored as Java annotation objects).

### DIFF
--- a/jmix-core/core/src/main/java/io/jmix/core/entity/annotation/LookupItemsQuery.java
+++ b/jmix-core/core/src/main/java/io/jmix/core/entity/annotation/LookupItemsQuery.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2026 Haulmont.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.jmix.core.entity.annotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures lazy loading of drop-down items for {@link LookupField} with
+ * {@link LookupType#DROPDOWN}. Usable only as the {@link LookupField#itemsQuery()} member.
+ * <p>
+ * Either set an explicit {@link #query()}, or set {@link #byInstanceName()} to match items by
+ * query conditions built from the instance-name metadata of the referenced entity. If neither
+ * is set, items are loaded eagerly.
+ */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LookupItemsQuery {
+
+    /**
+     * If true, drop-down items are matched by query conditions
+     * ({@link io.jmix.core.querycondition.Condition}) built from the instance-name metadata of
+     * the referenced entity: string-typed instance-name-related attributes are matched with a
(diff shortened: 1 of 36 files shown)


jmix-core/core/src/main/java/io/jmix/core/entity/annotation/LookupItemsQuery.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Configures lazy loading of drop-down items for {@link LookupField} with
+ * {@link LookupType#DROPDOWN}. Usable only as the {@link LookupField#itemsQuery()} member.
+ * <p>
+ * Either set an explicit {@link #query()}, or set {@link #byInstanceName()} to match items by
+ * query conditions built from the instance-name metadata of the referenced entity. If neither
+ * is set, items are loaded eagerly.
+ */
+@Target({})
```
glebfox, line 31: Should it be empty?

glebfox (approved):

jmix-core/core/src/main/java/io/jmix/core/entity/annotation/LookupItemsQuery.java
knstvk: Yes, it should be used only in LookupField#itemsQuery().

PR #5503, opened 2026-07-17
Programmatic GenericFilter prompts for a configuration name when opened for editing  #5451

### LINKED ISSUE #5451: Programmatic GenericFilter prompts for a configuration name when opened for editing
### Environment

Jmix Studio Plugin Version: 3.0.1-262
IntelliJ version: IntelliJ IDEA 2026.2 RC

### Bug Description

Programmatic GenericFilter is created by new API (#5351) with some name. But filter prompts for name when opened for editing or saving

````
        PropertyFilter<OrderStatus> statusFilter = genericFilter.filterComponentBuilder()
                .<OrderStatus>propertyFilter()
                .property("status")
                .operation(PropertyFilter.Operation.EQUAL)
                .build();

        genericFilter.runtimeConfigurationBuilder()
                .id("openOrders")
                .name("Open Orders")
                .add(statusFilter, OrderStatus.OPEN)
                .makeCurrent()
                .buildAndRegister();
````

### Steps To Reproduce

1.Launch demo project
2.Navigate to scenario 1
3.Click Edit or Save in menu that looks like COG

### Current Behavior

GenericFilter prompts for a configuration name

### Expected Behavior

GenericFilter should use configuration name that was defined by code

### Sample Project

Image

[filter-test.zip](https://github.com/user-attachments/files/30003772/filter-test.zip)

### PR DESCRIPTION
**Pull request by fractal3000**
Fixes #5451

### Problem

Opening the configuration editor for a named run-time configuration that exists only in memory (created programmatically, or added via `addConfiguration`, and never persisted) shows an empty "Configuration name" field and the "Configuration name required" error, even though the name is visible in the filter header. The conditions are transferred correctly; only the name is lost.

### Root cause

`GenericFilterSupport.loadFilterConfigurationModel(boolean isNewConfiguration, Configuration currentConfiguration)` loads the model from the database by id whenever `isNewConfiguration` is `false` (i.e. the id is not the empty-configuration id). For an in-memory configuration there is no database row, so the load returns `null` and the fallback creates an empty `FilterConfigurationModel` carrying only the username — the name (and id) are not copied. The editor binds to that empty model.

### Fix

In the fallback, when the configuration is not new, copy its `id` and `name` into the model. The id is required as well: on save, `saveCurrentFilterConfiguration` builds the result configuration from `configurationModel.getConfigurationId()`, and `initFilterConfiguration` replaces the existing configuration when the ids differ — so without the id the save would drop the original and create a broken/duplicate configuration. The `isNewConfiguration` guard leaves both empty for a genuinely new configuration.

### Notes

- Pre-existing editor bug, independent of the programmatic-configuration API (#5351): it reproduces for any in-memory named run-time configuration not present in the database, regardless of how it was created.
- The persistence path lives in `flowui-data`, so the regression test lives there too.

### Tests

Added `GenericFilterConfigurationNamePrefillTest` (`flowui-data`) covering all three branches of `loadFilterConfigurationModel(boolean, Configuration)`:

- **Existing in-memory configuration, not in the DB** (`isNewConfiguration == false`, load returns `null`): the fallback model is pre-filled with the configuration's id and name. This is the fixed case — it fails on the unpatched code (`expected: <Open Projects> but was: <null>`) and passes with the fix.
- **Existing persisted configuration** (`isNewConfiguration == false`, `configurationModel != null`): the stored model is returned unchanged, the fallback is not entered.
- **New configuration** (`isNewConfiguration == true`): the fallback model stays empty (no id/name), so creating a new configuration still asks the user for a name — this guards the `!isNewConfiguration` guard.

The branch actually taken is asserted via `EntityStates.isNew` (a model loaded from the database is not new; a `metadata.create` fallback model is). The existing `FilterConfigurationPersistenceTest` stays green.

### DIFF
--- a/jmix-flowui/flowui-data/src/test/java/filter_configuration_persistence/GenericFilterConfigurationNamePrefillTest.java
+++ b/jmix-flowui/flowui-data/src/test/java/filter_configuration_persistence/GenericFilterConfigurationNamePrefillTest.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2026 Haulmont.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package filter_configuration_persistence;
+
+import filter_configuration_persistence.view.FilterConfigurationPersistenceTestView;
+import io.jmix.core.DataManager;
+import io.jmix.core.EntityStates;
+import io.jmix.core.querycondition.PropertyConditionUtils;
+import io.jmix.flowui.component.genericfilter.Configuration;
+import io.jmix.flowui.component.genericfilter.FilterConfigurationPersistence;
+import io.jmix.flowui.component.genericfilter.GenericFilter;
+import io.jmix.flowui.component.genericfilter.GenericFilterSupport;
+import io.jmix.flowui.component.genericfilter.model.FilterConfigurationModel;
+import io.jmix.flowui.component.logicalfilter.LogicalFilterComponent;
+import io.jmix.flowui.component.propertyfilter.PropertyFilter;
+import io.jmix.flowui.component.propertyfilter.SingleFilterSupport;
+import io.jmix.flowui.component.textfield.TypedTextField;
+import io.jmix.flowui.entity.filter.FilterValueComponent;
+import io.jmix.flowui.entity.filter.GroupFilterCondition;
+import io.jmix.flowui.entity.filter.PropertyFilterCondition;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: jmix-flowui/flowui-data/src/test/java/filter_configuration_persistence/GenericFilterConfigurationNamePrefillTest.java [hunk 1]
```diff
@@ -0,0 +1,178 @@
...
+        return view.genericFilter;
+    }
+
+    protected FilterConfigurationModel loadEditorModel(boolean isNewConfiguration, Configuration configuration)
+            throws Exception {
+        Method method = GenericFilterSupport.class.getDeclaredMethod(
+                "loadFilterConfigurationModel", boolean.class, Configuration.class);
+        method.setAccessible(true);
+        return (FilterConfigurationModel) method.invoke(genericFilterSupport, isNewConfiguration, configuration);
+    }
```
KremnevDmitry: Reflection is a bad practice

KremnevDmitry: Also, the tests miss the core case: none of them performs a save to verify that editing an unsaved configuration updates it by id instead of creating a duplicate. That anti-duplicate outcome is the essence of #5451 (the code comment even claims it), so it should be covered.

KremnevDmitry (changes requested):

Issue #5487, opened 2026-07-16
GenericFilter: a design-time configuration value changed via the URL is not reset on re-navigation


fractal3000:
## Environment

- Jmix: reproduces on current `release_3_0` and earlier; not version-specific.
- UI: Flow UI, `genericFilter` + `urlQueryParameters` facet, a design-time configuration.

## Bug Description

When a `genericFilter` uses a design-time configuration (defined in the view XML) and a filter value is changed via URL query parameters, the value is not reset to its default on a clean re-navigation — it stays applied as if it were the view default. For a runtime configuration the same value is reset.

Cause: `GenericFilterUrlQueryParametersBinder.applyInitialState` restores component state only inside the `instanceof RunTimeConfiguration` branch, while the URL mutation path (`updatePropertyCondition`) changes components regardless of configuration type.

## Steps To Reproduce

On `filter-test`:

1. Open **Scenario 22** (design-time configuration `byNumber`; grid filtered by `number CONTAINS "1"`).
2. In the address bar:
   `.../orders/scenario22?genericFilter_genericFilterConfiguration=byNumber&genericFilter_genericFilterCondition=property:number_contains_5` → grid filtered by `number CONTAINS "5"`.
3. Re-open **Scenario 22** via the menu (no query parameters).

## Current Behavior

The value stays `5` (grid `CONTAINS "5"`), as if it were the default. For a runtime configuration (**Scenario 23**) the same value resets to `1`.

## Expected Behavior

On a clean re-navigation the value resets to the design-time default (`1`), consistent with a runtime configuration.

## How to Fix

Run `restoreInitialStates()` regardless of configuration type (the state is already captured for design-time configurations too); keep only the structural add/remove reconciliation runtime-specific.

[filter-test.zip](https://github.com/user-attachments/files/30103684/filter-test.zip)

fractal3000:
Will be fixed in https://github.com/jmix-framework/jmix/pull/5504

mig-sam:
Tested on:
Jmix version: 3.0.999-SNAPSHOT
Jmix Studio Plugin Version: 3.0.SNAPSHOT8610-261
IntelliJ version: IntelliJ IDEA 2026.1

Issue #5486, opened 2026-07-16
GenericFilter: the remove (×) button on a condition inside a nested group does nothing


fractal3000:
## Environment

- Jmix: reproduces on current `release_3_0` and earlier; not version-specific.
- UI: Flow UI, `genericFilter`, a run-time configuration containing a nested group (group condition) with conditions inside.

## Bug Description

For a condition nested inside a group (`root → groupFilter → [condition, ...]`), the **×** remove button is shown but clicking it does not remove the condition.

Cause: `GenericFilter.removeFilterComponent` always removes from the configuration root, and `GroupFilter.remove` is not recursive (it only removes its direct children). A nested condition is not a direct child of the root, so it is not found and nothing is removed.

## Steps To Reproduce

1. Open a view with a `genericFilter` (non-design-time configuration).
2. Create a run-time configuration and add a group condition with two conditions inside it.
3. Click **×** on one of the nested conditions.

## Current Behavior

The nested condition is not removed and keeps filtering.

## Expected Behavior

The **×** button removes the condition it belongs to, regardless of nesting.

## How to Fix

Remove the component from its **actual parent** instead of always from the root: in `GenericFilter.removeFilterComponent`, locate the parent `LogicalFilterComponent` that contains the component and call `remove` on it (or make `LogicalFilterComponent.remove` recursive), consistently with other `remove` callers.

fractal3000:
Is not reproduced.

PR #5504, opened 2026-07-17
A generic filter created by the programmatic API becomes broken after navigating to the view a second time  #5425

### LINKED ISSUE #5425: A generic filter created by the programmatic API becomes broken after navigating to the view a second time
### Environment

Jmix Studio Plugin Version: 3.0.1-261
IntelliJ version: IntelliJ IDEA 2026.1

### Bug Description

A generic filter created by the programmatic API works fine after navigate to view, but stops to work in case user navigate one more time on this view. Generic filter disappears and user can't to reset filter or choose it from list of configurated filters. Only page refresh helps to restore filter.

### Steps To Reproduce

1.Launch demo project
2.Navigate to Scenarios -> Scenario 1 — Simple PropertyFilter
3.Navigate to Scenarios -> Scenario 1 — Simple PropertyFilter second time

### Current Behavior

A generic filter becomes broken after navigating to the view a second time

### Expected Behavior

A generic filter works the same as filter created by UI (filter disappears but user can choose it from list)

### Sample Project

https://github.com/user-attachments/assets/44cfee2b-a523-4ede-9c5b-44c9123327e8

[filter-test.zip](https://github.com/user-attachments/files/29598441/filter-test.zip)

fractal3000:
Fixes #5425 and #5487

### Problem

A `GenericFilter` bound via the `urlQueryParameters` facet broke on a clean re-navigation to the same view. When Vaadin reuses the view instance, `onInit` is not called again, and `applyInitialState()` wiped the current configuration instead of restoring it:

- a programmatically created/activated configuration disappeared and could not be reselected (#5425);
- a value or operation changed via the URL was not reset on re-navigation — for a design-time configuration it stuck as if it were the default (#5487).

### Fix

On re-navigation the binder now restores the captured initial state instead of clearing it:

- **Structure** — the configuration tree is reconciled recursively: conditions removed by the user are re-added, conditions added afterwards (e.g. from the URL) are dropped, order and `modified` flags are restored, component instances are preserved.
- **Values/operations** — the captured operation and value of each condition are restored for **every** configuration type, including design-time (#5487); structural and default-value restore stay runtime-specific.
- **Event storm** — the reconciliation unbinds the components-change listener and rebinds once, so it no longer pushes URL parameters per add/remove.

Supporting changes: the duplicated base-condition capture/compose logic in `GenericFilter` and `GroupFilter` was extracted into `@Internal BaseConditionSupport` (behaviour unchanged); the internal `InitialState`/`ComponentNode` records and the single-filter state helper are marked `@Internal`; generic-filter test views were renamed to the `*TestView` convention.

### Tests

Facet-level re-navigation tests driving the real event chain (`RestoreComponentsStateEvent` → `QueryParametersChangeEvent`): programmatic-baseline survival, empty-configuration clearing, URL value/operation/condition reset, design-time value reset (#5487), nested-group restore, pure-user-action scope, and an event-storm regression. Also a regression test for removing a condition nested in a group (#5486).

### Commits

- `restore configuration state on same-view re-navigation (5425, 5487)` — the fix and its tests.
- `add a regression test for removing a condition nested in a group (5486)` — the regression test.

### Backward compatibility

No `public` API contract is removed or changed: the `HasInitialState` interface is untouched, `GenericFilter.updateDataLoaderInitialCondition` is kept as `@Deprecated(forRemoval = true)` rather than removed, and the only new public types (`BaseConditionSupport`, `SingleFilterComponentStateSupport`) are `@Internal` and purely additive. The behavioural changes are the bug fixes themselves — re-navigation now restores the configuration instead of clearing it (#5425) and resets a URL-changed design-time value (#5487); both correct broken behaviour, so no application migration is required.

However, the following **`protected` members of the URL query-parameter binders changed shape**. They are now marked `@Internal`, and all usages are confined to the `flowui` module (no references in `jmix-premium` or sample apps), but a subclass that referenced them would break at source/binary level. Worth a release-note line for anyone who extended these binders:

- **BC-1 — `PropertyFilterUrlQueryParametersBinder`** (sharpest break): the `protected record InitialState(Operation, Object)` is **removed**, and the type of the `protected` field `initialState` changed from `InitialState` to `SingleFilterComponentStateSupport.State`. A subclass reading that field or the record no longer compiles.
- **BC-2 — `DataGridFilterUrlQueryParametersBinder.InitialState`**: record signature changed from `(String key, String property, PropertyFilter.Operation operation, Object value)` to `(String key, String property, SingleFilterComponentStateSupport.State state)`.
- **BC-3 — `GenericFilterUrlQueryParametersBinder.InitialState`**: record signature changed from `(Configuration configuration)` to `(Configuration configuration, List<ComponentNode> structure, Map<…> states, Map<…> defaultValues)`; a new `ComponentNode` record was added.

Mitigations already in place: all three records carry `@Internal`, the public `HasInitialState` contract does not expose them, and the new `SingleFilterComponentStateSupport` bean is resolved from the standard `io.jmix.flowui` component scan (no manual wiring needed).

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/DataGridFilterUrlQueryParametersBinder.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/DataGridFilterUrlQueryParametersBinder.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableMap;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.router.QueryParameters;
+import io.jmix.core.annotation.Internal;
 import io.jmix.flowui.component.UiComponentUtils;
 import io.jmix.flowui.component.grid.DataGrid;
 import io.jmix.flowui.component.grid.DataGridColumn;
@@ -62,6 +63,7 @@ public class DataGridFilterUrlQueryParametersBinder extends AbstractUrlQueryPara
     protected UrlParamSerializer urlParamSerializer;
     protected FilterUrlQueryParametersSupport filterUrlQueryParametersSupport;
     protected RouteSupport routeSupport;
+    protected SingleFilterComponentStateSupport singleFilterComponentStateSupport;
 
     protected List<InitialState> initialStates = new ArrayList<>();
 
@@ -79,6 +81,7 @@ public class DataGridFilterUrlQueryParametersBinder extends AbstractUrlQueryPara
     protected void autowireDependencies() {
         filterUrlQueryParametersSupport = applicationContext.getBean(FilterUrlQueryParametersSupport.class);
         routeSupport = applicationContext.getBean(RouteSupport.class);
+        singleFilterComponentStateSupport = applicationContext.getBean(SingleFilterComponentStateSupport.class);
     }
 
     protected void initComponent(Grid<?> grid) {
@@ -102,8 +105,7 @@ public class DataGridFilterUrlQueryParametersBinder extends AbstractUrlQueryPara
                         new InitialState(
                                 dataGridColumn.getKey(),
                                 Objects.requireNonNull(propertyFilter.getProperty()),
-                                propertyFilter.getOperation(),
-                                propertyFilter.getValue()
+                                singleFilterComponentStateSupport.capture(propertyFilter)
                         )
                 );
             }
@@ -163,8 +165,7 @@ public class DataGridFilterUrlQueryParametersBinder extends AbstractUrlQueryPara
                     && initialState.property.equals(headerFilter.getPropertyFilter().getProperty())) {
                 PropertyFilter propertyFilter = headerFilter.getPropertyFilter();
 
-                propertyFilter.setOperation(initialState.operation);
-                propertyFilter.setValue(initialState.value);
+                singleFilterComponentStateSupport.restore(propertyFilter, initialState.state());
                 headerFilter.apply();
             }
         }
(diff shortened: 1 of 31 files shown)


jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/DataGridFilterUrlQueryParametersBinder.java
KremnevDmitry: unchecked suppression no needed anymore

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/PropertyFilterUrlQueryParametersBinder.java
```diff
@@ -118,8 +120,7 @@ protected void updateQueryParameters() {
     @SuppressWarnings("unchecked")
```
KremnevDmitry: redundant

jmix-flowui/flowui/src/main/java/io/jmix/flowui/facet/urlqueryparameters/GenericFilterUrlQueryParametersBinder.java
```diff
@@ -205,13 +263,66 @@ protected String serializePropertyCondition(PropertyCondition condition) {
...
+            configuration.setFilterComponentModified(component, false);
+        }
+        for (ComponentNode node : structure) {
+            logicalFilterComponent.add(node.component());
+            configuration.setFilterComponentModified(node.component(), node.modified());
+            if (node.component() instanceof LogicalFilterComponent<?> nestedComponent) {
+                restoreStructure(configuration, nestedComponent, node.children());
+            }
+        }
+    }
```
KremnevDmitry, line 327: `applyInitialState()` runs on every same-view re-navigation (`View.java:228` → `UrlQueryParametersFacetImpl.java:129`), and `restoreStructure` (`:289`) unconditionally removes (`:298`) and re-adds (`:302`) every baseline component even with no user edits. 


`GroupFilter.add()` registers a fresh `operationChange → apply()` listener (`GroupFilter.java:283`) whose `Registration` is discarded, and `remove()` (`:306`) never unregisters it. So a baseline `PropertyFilter` accumulates K+1 `apply()` listeners after K re-navigations, retained for the view's lifetime. Changing its operation then fires K+1 `dataLoader.load()` calls (K+1 DB queries + URL pushes) — a resource/perf regression, though the restored state stays correct. 

This is newly introduced: the old `removeAll()` + `setCurrentConfiguration()` path never re-added baseline instances via `add()`.

KremnevDmitry (changes requested):

PR #5532, opened 2026-07-27
Support for Firebird database

makssent:
Firebird keeps coming up as an additional/secondary database around the CUBA and Jmix ecosystem. For example, it's already used as an additional data store in jmix-db-tests. There seems to be some interest in it, so I decided to try adding it as a supported main database type.

I don't have access to add Firebird support to Studio, but this should be a reasonable starting point for the open-source part of the framework.


knstvk:
Unfortunately, we cannot accept your PR. Support for a new database requires ongoing maintenance effort from the team, and we cannot afford it.

Please consider making an add-on instead.

PR #5634, opened 2026-08-28
One condition is shown but another is applied in GenericFilter

### LINKED ISSUE #5516: One condition is shown but another is applied in GenericFilter
### Environment

Jmix Studio Plugin Version: 3.0.1-261
IntelliJ version: IntelliJ IDEA 2026.1

### Bug Description

The user sees a GenericFilter with the condition from the `openOrders` configuration, but the rows are fetched using a different condition -`PropertyCondition.greaterOrEqual("amount", BigDecimal.valueOf(100))`

### Steps To Reproduce

1.Launch demo project
2.Navigate to Scenario 16 view
3.Click button `Set base AFTER activation (amount ≥ 100)`
Following code will be executed:
````
        // Set the base AFTER a configuration is active; it is adopted on the next configuration switch.
        genericFilter.setCurrentConfiguration(genericFilter.getConfiguration("openOrders"));
        genericFilter.getDataLoader().setCondition(PropertyCondition.greaterOrEqual("amount", BigDecimal.valueOf(100)));
        genericFilter.apply();
````

### Current Behavior

One condition is shown but another is applied in GenericFilter

### Expected Behavior

Should be shown condition that is applied

### Sample Project

Image

[filter-test.zip](https://github.com/user-attachments/files/30307181/filter-test.zip)

### PR DESCRIPTION
**Pull request by fractal3000**
See #5516

`GenericFilter` composes the data loader condition as `base AND current configuration`, where the base is whatever the application set on the loader itself. Until now that composition ran only when the configuration was switched; every other applying gesture just loaded. When the application replaced the loader condition after a configuration was already active (allowed since #5400), the loader kept loading by the new base alone while the filter showed the configuration's condition.

Now every applying gesture recomposes before loading — but **only when the loader condition actually diverged**: the filter remembers the condition object it set last (`lastConditionSetByFilter`, the rule introduced by #5400) and recomposes only if the application has replaced it since. An untouched loader condition is not rewritten, so applications that never replace the base see byte-for-byte the previous behavior. The covered gestures:

- `GenericFilter.apply()`, the Apply button and its shortcut; a standalone `GroupFilter.apply()` — recompose in place;
- a condition **value** change (Enter, a dropdown selection) and an **operation** change, which load the data loader directly from the condition component or from the delegated group: before such a load the component invokes a recomposition delegate — a callback to the owning `GenericFilter` or standalone `GroupFilter`, maintained by the owner itself (a group hands its chain to children in `add()` and clears it in `remove()`, the filter sets it on a configuration's root group, the converters set it on components they create). Works at any nesting level, including a group nested in a standalone `groupFilter`; `DataGridHeaderFilter` stays outside this wiring, which is pinned by a test.

Tests: `GenericFilterApplyAfterBaseChangeTest` (flowui, 12 cases) and `DataGridHeaderFilterApplyTest` (the header filter is outside the recomposition wiring) — apply()/Apply button/value change/operation change after the base was replaced, for `GenericFilter` and standalone `GroupFilter` incl. a nested group, `autoApply=false` variants, repeated apply() does not accumulate conditions, exactly one load per gesture, the empty configuration without a base, and the delegated root group leaving an untouched loader condition alone. Six of the original eight fail without the apply-path fix; the four gesture cases fail without the gesture fix. Full `:flowui:test` (1122) and `:flowui-data:test` (31) pass; the `filter-test` scenario app was swept with Playwright on 3.0.1 and on this branch — 47 views, the only differences are the scenarios that reproduce this issue.

Notes for review and release notes:

- **Behavior change (the point of the fix).** After `dataLoader.setCondition(...)` while a configuration is shown, any applying gesture composes `base AND configuration`. An application that replaced the loader condition specifically to override the filter will see the configuration re-applied. Repro: replace the loader condition after a configuration is active, then press Enter in a condition field.
- **Known characteristic.** Two filters on one `DataLoader`: alternating applying gestures adopt the neighbour's composition as the base and re-add their own output — the condition tree grows with duplicate (idempotent) AND entries; query results stay correct. Repro: `GenericFilter` + standalone `groupFilter` on one loader, alternate gestures between them. Root cause is the shared loader condition, out of scope here.
- **Remaining gap (unchanged by this PR).** A standalone `propertyFilter`/`jpqlFilter` that is not part of a composite filter, and a configuration whose condition components are all invisible, still load by a replaced base. Both are reachable only programmatically.
- **Pre-existing, filed as #5655.** During URL restore, an operation change fires an intermediate load (`GroupFilter.add` registers an unconditional operation-change listener). This PR does not change the number of loads; the fix is #5659, stacked on this PR.

QA (`filter-test` from #5586, Scenarios 16, 25, 26 — the attached archive of #5586 predates 25/26; the current project has them):
- Scenario 16, **Set base AFTER activation (amount ≥ 100)**: the label must show `AND[amount greater_or_equal 100, AND[status equal OPEN]]` and only Open orders — not `amount greater_or_equal 100` alone.
- Scenario 25, **Replace base only**, then the filter's **Refresh**, or a change of the Status **value**, or of its **operation**: both conditions, not the base alone.
- Scenario 26 (standalone `GroupFilter`), **Replace base, then apply()**, and **Replace base only** + change the Status value or operation: both conditions.
- A view with a filter and no configuration selected: Refresh still loads all rows; no errors.
- `autoApply="false"`: pressing Apply after the application replaced the loader condition loads by both conditions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### CODE REVIEW

File: jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java [hunk 1]
```diff
@@ -242,8 +242,16 @@ public void setAutoApply(boolean autoApply) {
...
+        if (dataLoader != null) {
+            // A standalone group recomposes its output onto a base condition the application may have
+            // replaced since the last contribution, regardless of autoApply, so that a later load by
+            // the application uses the right condition. A delegated group leaves this to its owner.
+            if (!isConditionModificationDelegated()) {
+                updateDataLoaderCondition();
+            }
+            if (autoApply) {
+                dataLoader.load();
+            }
```
KremnevDmitry, line 253: **1. Value/operation edits still load by the replaced base (`autoApply=true`, the default).**
Neither gesture reaches the patched entry points: a value change goes through `SingleFilterComponentBase.apply()` (loads directly, no recompose), an operation change goes through the delegated root group's `apply()` (skips recompose, still loads). So after `dataLoader.setCondition(newBase)`, both gestures load by `newBase` alone while the UI shows the configuration — the most common reproduction path of #5516. Acknowledged in the description, but worth a tracked follow-up issue. Also, the comment “A delegated group leaves this to its owner” overstates: the owner is never invoked on this path — better phrased as a known limitation.

**2. Same gap for a nested group inside a *standalone* `groupFilter`** — not covered by the disclosed “inside a GenericFilter” limitation.
The operation-change listener calls the *nested* (delegated) group's `apply()`, which skips recompose and loads the bare base; the owning group is never notified. A direct child of the standalone group recomposes correctly (tested), so behavior differs one nesting level deeper. Reachable via plain XML.

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
fractal3000: **1.** Fixed rather than deferred: commit `0e4a1ce268` covers both gestures. Before its direct load, a delegated condition component (or a delegated group) walks up the component tree to the nearest non-delegated owner — the `GenericFilter` or a standalone `GroupFilter` — and lets it recompose, under the same divergence guard as `apply()`: an untouched loader condition is left exactly as before. The code comment you flagged is reworded accordingly ("a delegated group does not compose itself: before loading it lets the nearest owner recompose an outdated base"). The remaining gap is narrowed to programmatic-only paths (a standalone `propertyFilter` outside any composite filter; a configuration whose condition components are all invisible) and is disclosed in the description.

**2.** Confirmed, and covered by the same commit: the walker skips delegated groups, so a nested group inside a standalone `groupFilter` reaches the owning group — there is a dedicated test for exactly this case ("an operation change in a group nested in a standalone GroupFilter recomposes onto a replaced base").

Verification: 4 new gesture tests are red without the commit and green with it; full `:flowui:test` (1121) and `:flowui-data:test` (31) pass. The description also lists two behavioral notes worth a reviewer's eye: the gesture semantics change for applications that replace the loader condition to override the filter (the point of #5516), and the duplicate-entry growth when two filters share one loader (same identity rule as the apply paths; root cause is the shared loader condition, out of scope here).

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -318,6 +321,12 @@ protected void updateApplyButtonText(boolean autoApply) {
     }
 
     protected void onApplyButtonClick(ClickEvent<MenuItem> clickEvent) {
+        // Same recomposition rule as apply(): recompose only when the application replaced the
+        // loader condition since the filter's last contribution. Unlike apply(), the button always
+        // loads and keeps the current page.
+        if (isLoaderConditionOutdated()) {
+            updateDataLoaderCondition();
+        }
```
KremnevDmitry: Copy-paste of `recomposeLoaderConditionIfOutdated` method body

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
```diff
@@ -465,11 +474,36 @@ protected void updateCurrentConfigurationAutoApply(boolean autoApply) {
      */
     public void apply() {
         if (dataLoader != null) {
+            // The application may have replaced the loader condition since the filter's last
+            // contribution (a new base condition); compose that base AND the shown configuration
+            // before loading, not the base alone. An untouched loader condition is left as is,
+            // so applications that never replace it see exactly the previous behavior.
+            if (isLoaderConditionOutdated()) {
+                updateDataLoaderCondition();
+            }
```
KremnevDmitry: Copy-paste of `recomposeLoaderConditionIfOutdated` method body

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
```diff
@@ -242,11 +243,41 @@ public void setAutoApply(boolean autoApply) {
...
-        if (dataLoader != null && autoApply) {
-            dataLoader.load();
+        if (dataLoader != null) {
+            // A standalone group recomposes its output onto a base condition only when the
+            // application replaced the loader condition since the group's last contribution;
+            // an untouched loader condition is left as is. A delegated group does not compose
+            // itself: before loading it lets the nearest owner recompose an outdated base.
+            if (!isConditionModificationDelegated() && isLoaderConditionOutdated()) {
+                updateDataLoaderCondition();
+            }
```
KremnevDmitry: Copy-paste of `recomposeLoaderConditionIfOutdated` method body

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SupportsLoaderConditionRecompose.java
```diff
@@ -0,0 +1,60 @@
...
+ * ({@code base AND own output}) and can refresh it when the application has
+ * replaced the loader condition since the filter's last contribution.
+ * <p>
+ * A condition component whose modifications are delegated loads the data loader
+ * directly, bypassing the owner's composition. Before such a load it locates the
+ * nearest owner up the component tree and lets it recompose, so the load never
+ * runs by a stale base alone.
+ */
+@Internal
+public interface SupportsLoaderConditionRecompose {
```
KremnevDmitry: Why do we need an interface for all this? It looks like overengineering. 

It exists only for `instanceof` in the component tree traversal, and the traversal exists only because the delegated component doesn't know its owner. We can get rid of it.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SingleFilterComponentBase.java
```diff
@@ -210,6 +210,13 @@ public void apply() {
         if (isAttached() && dataLoader != null) {
             setupLoaderFirstResult();
             if (autoApply) {
+                // A delegated condition loads directly, bypassing the owning filter's composition;
+                // if the application replaced the loader condition since the owner composed it
+                // last, this condition's query condition is no longer part of it. Let the owner
+                // recompose first, so the load never runs by the replaced base alone.
+                if (isConditionModificationDelegated()) {
+                    SupportsLoaderConditionRecompose.recomposeNearestOwner(this);
+                }
```
KremnevDmitry: Is the `DataGridHeaderFilter` affected?

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
fractal3000: Fixed in `12890bbb48` — all three inlined copies are replaced with the method call. The follow-up `4155278db6` completes the dedup: the "condition was replaced" predicate now lives once, in `BaseConditionSupport.isReplacedExternally()`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/GenericFilter.java
fractal3000: Fixed in `12890bbb48`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
fractal3000: Fixed in `12890bbb48`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SupportsLoaderConditionRecompose.java
fractal3000: Agreed — reworked in `12890bbb48`: the interface and the tree traversal are removed. The owner sets a `Runnable` delegate on its children itself (the root gets it from `GenericFilter`, nested ones from `GroupFilter.add()`), the same wiring as already done for `conditionModificationDelegated`.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/filter/SingleFilterComponentBase.java
fractal3000: No: its `PropertyFilter` is not delegated and receives no delegate, so its `apply()` path is unchanged. Pinned by the new `DataGridHeaderFilterApplyTest`.

fractal3000:
Rework pushed: `12890bbb48` (a delegate instead of the interface + traversal, dedup, a header-filter test) and `4155278db6` (a quality pass over the final diff). `:flowui:test` is green, both gestures (value / operation change after the loader condition was replaced) verified on the demo project, and the UI sweep shows no behavior change.

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/converter/AbstractFilterComponentConverter.java
```diff
@@ -37,6 +38,12 @@ public AbstractFilterComponentConverter(GenericFilter filter) {
     public C convertToComponent(M model) {
         C filterComponent = createComponent();
         filterComponent.setConditionModificationDelegated(true);
+        // A root group created by a converter belongs to the filter directly and is never added to
+        // another group, so it gets the filter's delegate here; any other component is re-wired to
+        // its owning group's chain when it is added to that group.
+        if (filterComponent instanceof GroupFilter groupFilter) {
+            groupFilter.setLoaderConditionRecomposeDelegate(filter::recomposeLoaderConditionIfOutdated);
+        }
```
KremnevDmitry: Wiring the delegate at creation sites misses any root built outside them (e.g. a hand-built `RunTimeConfiguration` with a `filterComponentBuilder().groupFilter()` root), which then never recomposes. Set it once in `refreshCurrentConfigurationLayout()`, where the filter adopts the current root, and drop the converter branch.

KremnevDmitry (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/genericfilter/converter/AbstractFilterComponentConverter.java
fractal3000: Agreed — done in `129fa8dc90`: the delegate is set in `refreshCurrentConfigurationLayout()` when the filter adopts the root — every configuration activation passes through it, whoever built the root. The converter branch and the factory wiring are removed, `recomposeLoaderConditionIfOutdated` is back to `protected`. Added a test with a hand-built `RunTimeConfiguration` — red before the rework, green after.

KremnevDmitry (approved):

PR #5659, opened 2026-09-04
GenericFilter: restoring filter state from the URL fires a redundant data load when a condition operation differs

### LINKED ISSUE #5655: GenericFilter: restoring filter state from the URL fires a redundant data load when a condition operation differs
### Environment

Jmix version: 3.0.1

### Bug Description

Performance problem. When the filter state is restored from URL query parameters and the operation of a condition in the URL differs from the one stored in the current configuration, the view fires a redundant intermediate data load before the final one — two loads per navigation instead of one.

Mechanics: `GenericFilterUrlQueryParametersBinder.updatePropertyCondition` applies the operation from the URL to the live condition component (`configurationComponent.setOperation(...)`). The operation-change listener registered by `GroupFilter.add` does not distinguish this programmatic call from a user gesture and unconditionally invokes `GroupFilter.apply()`; for the delegated root group its `apply()` skips recomposition but still calls `dataLoader.load()` (there is no `isAttached()` check, unlike `SingleFilterComponentBase.apply()`). The restore then finishes with its own composition and load. Value changes do not add loads (`UiComponentUtils.setValue` is not a client-side change).

The intermediate load runs the full pipeline — the SELECT, the pagination count query, entity materialization by the fetch plan, row-level security — and its result is immediately discarded. For a view whose query takes ~400 ms, opening a deep link pays roughly double. The condition of the discarded load also differs from the final one, so on large tables it may hit a worse query plan than the final load itself.

The gap is reached with default settings whenever a user changes a condition operation (conditions added by the user and saved configurations have `operationEditable = true` by default) and the URL is then reopened — F5, browser Back, a bookmark, a link shared with a colleague.

### Steps To Reproduce

1. A list view with a `GenericFilter` bound to the data loader and an `urlQueryParameters` facet with the `genericFilter` binder; a configuration whose condition has an editable operation, e.g.:

   ```xml
   <configurations>
       <configuration id="byStatus" name="By status">
           <propertyFilter property="status" operation="EQUAL" operationEditable="true"/>
       </configuration>
   </configurations>
   ```

2. Count data loads, e.g. with a `PostLoadEvent` listener on the loader, or enable SQL logging.
3. Select the configuration, pick a value, change the operation (`=` → `<>`): the URL now carries `..._genericFilterCondition=property:status_not-equal_...`.
4. Open the resulting URL in a fresh tab (or press F5).

### Current Behavior

Two data loads. With a load journal on the view:

```
Loads: 2 — load #1: AND[status not_equal OPEN]  |  load #2: AND[status not_equal OPEN, AND[status not_equal OPEN]]
```

The same URL with the matching operation (`..._equal_...`) produces one load — the only difference between the two URLs is the operation token.

### Expected Behavior

One data load per navigation: the restore applies the URL state as a single composition and loads once.

### Notes

- Same family as #5623 (the URL restore path eagerly does work it may not need). A proper implementation of #5623 — parse the URL into condition models first, validate on them, then update components and apply the state as a single composition with a single load — resolves this issue as well; suggested acceptance criterion for #5623: exactly one data load per navigation.
- A local fix alternative: suppress the operation-change listener during restore (or make it respect the gesture origin).

fractal3000:
See #5655

**Stacked on #5634** (base branch = its branch; GitHub will retarget this PR to `release_3_0` when #5634 merges). Only the last commit belongs to this PR.

Restoring the filter state from URL query parameters fires a redundant intermediate data load. Reproduction (Scenario 27 of the `filter-test` project attached to #5655): a configuration with an editable-operation condition, the user changes the operation (`=` → `<>`), then reloads the page or opens the bookmarked link — two data loads instead of one, the first by a half-restored state and discarded. For a 400 ms query a deep-link open pays roughly double.

Cause: `GenericFilterUrlQueryParametersBinder.updatePropertyCondition` applies the operation from the URL programmatically, and the operation-change listener registered by `GroupFilter.add` invoked `apply()` — and thus `dataLoader.load()` — for a programmatic `setOperation` as well, unlike the value path, which has always been gated by `isFromClient`.

Fix: the listener applies the filter on a client-driven change only. A programmatic `setOperation` no longer fires a load of its own — consistent with a programmatic `setValue`, which never did.

Tests: `GenericFilterUrlRestoreLoadTest` — a restore with a changed operation fires no load (fails without the fix); the tests that used the automatic load as an instrument (`GenericFilterReNavigationTest`, `GenericFilterApplyAfterBaseChangeTest`) now emulate the client-driven gesture via `setOperationInternal(op, true)`. Full `facet.url_query_parameters.*` and `component.genericfilter.*` pass on top of #5634.

Release note: an application that relied on a programmatic `PropertyFilter.setOperation(...)` reloading the data must call `apply()` or `dataLoader.load()` explicitly.

On `master` the same fix arrives with the forward-port of this PR (it was deliberately kept out of #5658).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
+++ b/jmix-flowui/flowui/src/main/java/io/jmix/flowui/component/logicalfilter/GroupFilter.java
@@ -314,8 +314,15 @@ public class GroupFilter extends Composite<VerticalLayout>
         if (filterComponent instanceof PropertyFilter) {
             // Keep the registration so remove() can detach it; otherwise re-adding a component
             // (e.g. on a filter re-navigation restore) would accumulate stale apply() listeners.
+            // Apply on the user's gesture only: a programmatic operation change (e.g. the URL binder
+            // restoring the filter state) must not fire a load of its own, consistently with the
+            // value path, which is gated by isFromClient in SingleFilterComponentBase.
             Registration operationChangeRegistration = ((PropertyFilter<?>) filterComponent)
-                    .addOperationChangeListener(operationChangeEvent -> apply());
+                    .addOperationChangeListener(operationChangeEvent -> {
+                        if (operationChangeEvent.isFromClient()) {
+                            apply();
+                        }
+                    });
             operationChangeRegistrations.put(filterComponent, operationChangeRegistration);
         }
 
--- a/jmix-flowui/flowui/src/test/groovy/component/genericfilter/GenericFilterApplyAfterBaseChangeTest.groovy
+++ b/jmix-flowui/flowui/src/test/groovy/component/genericfilter/GenericFilterApplyAfterBaseChangeTest.groovy
@@ -165,7 +165,7 @@ class GenericFilterApplyAfterBaseChangeTest extends FlowuiTestSpecification {
 
         when: "the base is replaced again and the user changes the operation of a condition, which applies the group through its listener"
         groupFilter.dataLoader.setCondition(PropertyCondition.greater("amount", 0))
-        number.setOperation(PropertyFilter.Operation.CONTAINS)
+        number.setOperationInternal(PropertyFilter.Operation.CONTAINS, true)
 
         then:
         hasPropertyConditionOn(groupFilter.dataLoader.condition, "amount")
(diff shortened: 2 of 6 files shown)


KremnevDmitry (approved):
Related issue should be marked as `breaking changes` by behaviour

PR #5662, opened 2026-09-07
Base view class and studio view template for read only show page of an entity

### LINKED ISSUE #3833: Base view class and Studio view template for read-only show page of an entity
It would be great to have a base view class for cases when we need a read-only details of an entity.

Like https://marmelab.com/react-admin/ShowTutorial.html in React Admin.

For example:
- for views that only show data, no editing at all
- or for entities that are modified in the screen, but in a custom way. E.g. an entity linked to a business process:
  -  We can move it to another status, change some attributes. But it is all done programmatically, and changes are saved via custom `@Service` beans.
  - There is no form with input fields to edit.
  - There are no OK / Cancel buttons. All changes are saved instantly.

Unexperienced devs are afraid to create such views from scratch. They are using StandardDetailView as base class, and after they get some problems such as:
- "unsaved changes" with no reason
- problems with "refreshing" changed entity in the data context (because InstanceLoader.load() calls DataContext.merge() -- which in turn has a very complex logic of merging entity state).

Read-only show page most likely needs to have read-only data context.

### LINKED ISSUE #5628: Add `entity_read` action for EntityPicker
### Description

Once an entity can have a dedicated read-only view (see jmix-framework/jmix#3833), there is no picker action that opens a referenced entity for viewing. `EntityPicker` has only `entity_open`, which opens the detail view, i.e. the editing view.

The list data components solve the same problem with two separate actions: `list_edit` (requires the UPDATE permission, disabled when the view is read-only) and `list_read` (requires only READ, stays enabled). `EntityPicker` has no counterpart of the second one.

### Use case

A read-only view shows an entity whose reference attributes are rendered as entity pickers. Following such a reference is a natural drill-down: the user wants to look at the referenced entity, not to edit it. The same applies to users who are only permitted to read the referenced entity — today the only available action sends them to an editing view.

### Expected behavior

`EntityPicker` should have a read action that opens the picker value in a read-only view, resolving the view the same way `list_read` does after #3833: the view marked as primary read view, otherwise the view with the `<entity>.read` id, otherwise the detail view opened in read-only mode. It should require only the READ permission and stay enabled on a read-only view. The existing `entity_open` action keeps its current behavior.

Worth checking as part of this issue: which picker actions remain reachable once the picker itself is read-only — if they are not, drill-down from a read-only view is impossible today regardless of the action used.

### Impact

Drill-down from a read-only view stays read-only along the whole path, and applications no longer have to replace a picker action to achieve it.

KremnevDmitry:
# Summary

Added the read view — a view that shows one entity instance and cannot edit it, a sibling of `StandardDetailView` rather than its subclass. It is reachable the same ways a detail view is: by resolution from an entity, from a list or a picker action, by navigation and in a dialog, in plain and tabbed-mode applications, and it is generated by Studio.

### What was done

- Base class and contract in module `flowui`: `ReadView<E>` with `setEntityToRead` / `getEntity`, `StandardReadView<E>` and the `@ReadEntityContainer` annotation naming the instance container. The class guarantees a read-only `DataContext` and read-only data-bound components by construction — descriptors of read views carry no `readOnly` attribute.
- View resolution: the `@PrimaryReadView` annotation and the `<entity>.read` id convention, with a fallback to the entity's detail view opened in read-only mode. `ViewRegistry` gained `getReadViewInfo(...)` and `getAvailableReadViewId(...)`; every entry point shares this single chain.
- Entry points: `ViewNavigators.readView(...)` and `DialogWindows.read(...)`, each taking an explicit entity class, a list component or a picker, plus `ReadViewNavigator` / `ReadWindowBuilder` and their processors.
- Actions: `list_read` (`ReadAction`) now opens the entity's read view when one exists and keeps its previous behavior otherwise; the new `entity_read` (`EntityReadAction`) does the same for `EntityPicker`, so a drill-down from a read view stays read-only. Its shortcut comes from the new `jmix.ui.component.picker-read-shortcut` property.
- Tabbed mode support in the commercial `tabbedmode-flowui` addon: its own read view builder with an adapter over the framework navigator and dialog builder, and replacements for the two framework processors, so all entry points open a read view in a tab instead of navigating to its route.
- Studio: the `Entity read view` and `DTO entity read view` templates in `jmix-templates`, generating the controller, the descriptor with a form over the chosen fetch plan, the `<entity>.read` id and the title message.

### How it works

Opening a read view for an entity resolves the view in one order everywhere: a view annotated `@PrimaryReadView`, otherwise the view registered as `<entity>.read`, otherwise the entity's detail view — which is then opened read-only, so an application that has no read view behaves exactly as before.

The view takes the entity id from its route parameter (or from `setEntityToRead` when opened in a dialog) and loads the instance through its own `InstanceLoader` and fetch plan, so a passed instance is always re-read rather than shown as is. Read-only is enforced in two places the developer cannot forget: the data context created for a read view is a `NoopDataContext`, and on `ReadyEvent` all components bound to a value source become read-only, which also disables actions that adjust themselves on a read-only view. Unbound components and filters are untouched.

Generated views take the route `<plural>/:id/read`, so they do not collide with the detail view's `<plural>/:id`.

### How to use

Generate the view from Studio (`Entity read view`), or write it by hand:

```java
@Route(value = "customers/:id/read", layout = MainView.class)
@ViewController(id = "Customer.read")
@ViewDescriptor(path = "customer-read-view.xml")
@ReadEntityContainer("customerDc")
public class CustomerReadView extends StandardReadView<Customer> {
}
```

Then open it declaratively — `<action id="read" type="list_read"/>` on a data grid, `<action id="read" type="entity_read"/>` on an entity picker (both are offered by the Studio view wizard) — or programmatically with `viewNavigators.readView(this, Customer.class).readEntity(customer).navigate()` or `dialogWindows.read(this, Customer.class).readEntity(customer).open()`.

### Breaking changes

None in signatures. One behavior change worth a release note: `list_read` used to always open the detail view in read-only mode, and now opens a view registered under `<entity>.read` if the application has one. Applications without such a view are unaffected. `UiComponentProperties` gained a constructor parameter for the new shortcut property, which affects only code instantiating that class directly.

### Compatibility

Opt-in: an application sees no change until it adds a read view or a `<entity>.read` view id.

### DIFF
--- a/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/action/StudioActionPropertyGroups.java
+++ b/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/action/StudioActionPropertyGroups.java
@@ -567,6 +567,30 @@ final class StudioActionPropertyGroups {
                             xmlAttribute = StudioXmlAttributes.ICON,
                             type = StudioPropertyType.ICON,
                             category = StudioProperty.Category.LOOK_AND_FEEL,
+                            defaultValue = "EYE",
+                            setParameterFqn = "com.vaadin.flow.component.Component"),
+                    @StudioProperty(
+                            xmlAttribute = StudioXmlAttributes.ID,
+                            type = StudioPropertyType.COMPONENT_ID,
+                            category = StudioProperty.Category.GENERAL,
+                            required = true,
+                            initialValue = "entityRead"),
+                    @StudioProperty(
+                            xmlAttribute = StudioXmlAttributes.TEXT,
+                            type = StudioPropertyType.LOCALIZED_STRING,
+                            category = StudioProperty.Category.GENERAL,
+                            defaultValue = "msg:///actions.entityPicker.read.description")
+            }
+    )
+    public interface EntityReadActionComponent extends BaseAction {
+    }
+
+    @StudioPropertyGroup(
+            properties = {
+                    @StudioProperty(
+                            xmlAttribute = StudioXmlAttributes.ICON,
+                            type = StudioPropertyType.ICON,
+                            category = StudioProperty.Category.LOOK_AND_FEEL,
                             defaultValue = "SEARCH",
                             setParameterFqn = "com.vaadin.flow.component.Component"),
                     @StudioProperty(
--- a/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/action/StudioListDataComponentActions.java
+++ b/jmix-flowui/flowui-kit/src/main/java/io/jmix/flowui/kit/meta/action/StudioListDataComponentActions.java
@@ -108,7 +108,7 @@ interface StudioListDataComponentActions {
 
     @StudioAction(
             type = "list_read",
-            description = "Opens a detail view for an entity instance in read-only mode",
+            description = "Shows an entity instance in its read view, or in its detail view opened in read-only mode",
             classFqn = "io.jmix.flowui.action.list.ReadAction",
             documentationLink = "%VERSION%/flow-ui/actions/list-actions.html#list_read",
             availableInViewWizard = true,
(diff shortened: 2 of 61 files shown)


glebfox (approved):

glebfox:
@KremnevDmitry support Tabbed Mode add-on

KremnevDmitry:
Support is included in the jmix-prem repo branch

PR #5674, opened 2026-09-09
Show dynamic model entities and attributes in Data Tools views

### LINKED ISSUE #5449: Show dynamic model entities and attributes in Data Tools views
### Description

The Data model view and the Entity inspector of the Data Tools add-on are built from JPA metadata. Entities and attributes created at runtime by the Dynamic Model add-on are invisible in both: a dynamic entity has no JPA table mapping, and a dynamic attribute of a static entity has no JPA column mapping.

As a result, a user who extends the data model at runtime cannot see the resulting schema in the Data model view, and cannot browse or edit the data through the Entity inspector — the two tools they would normally reach for.

### Expected behavior

**Data model view**

- Dynamic entities are listed like any other entity, with their data store and physical table.
- Dynamic attributes of a static entity are listed on that entity. Because they are stored in a side table, their column is qualified by the table name so the physical layout stays visible.
- Attributes that have no column of their own, such as calculated and collection attributes, are listed without a column.
- The identifier and the columns generated for collections are listed like any other column.
- The diagram indicates what is dynamic: the entity itself for a dynamic entity, the individual attributes for a static entity extended at runtime.
- The indication of what is dynamic appears only when the Dynamic Model add-on is present, so it does not add an always-empty column to other applications.

**Entity inspector**

- Dynamic entities appear in the entity dropdown and can be browsed, created, edited and deleted.
- Dynamic attributes of static entities appear in both the list and the detail view, and can be edited.
- References to dynamic entities are editable, not read-only.
- Filtering and sorting by dynamic attributes work through the generic filter.

The Entity inspector should decide what it can work with from an entity's **data store** rather than from JPA annotations. Deciding by store rather than by dynamic-model-specific rules means entities of any non-JPA store — REST DTO stores and other custom stores — become browsable in the inspector as well.

Applications that do not use the Dynamic Model add-on must be unaffected, and the open-source Data Tools add-on must not depend on the commercial Dynamic Model add-on.

### Impact

Runtime model extension becomes inspectable and manageable with the standard tools instead of requiring direct database access. The same change makes the Entity inspector usable for any custom non-JPA data store, which previously had no generic browsing UI at all.

knstvk:
### Summary

Made the Data Tools views work with entities and attributes that JPA metadata cannot describe, so add-ons and custom data stores can appear in them. Adds one extension point for the Data model view and makes the Entity inspector decide what it can handle from an entity's data store rather than from JPA annotations.

### What was done

- Added `DataModelContributor` (module `datatools`), the extension point a module implements to describe entities and attributes the Data model view cannot derive from JPA metadata, with the `EntityDescriptor` and `AttributeDescriptor` records it returns. `DataModelRegistry` consults contributors alongside its JPA scan; a contributor that throws is logged and skipped rather than breaking the whole view.
- Added `EntityInspectorSupport` (module `datatools`), which answers three questions from an entity's data store: which entities the inspector can browse, whether the store accepts a JPQL query string, and whether a property is stored. `EntityInspectorListView`, `EntityInspectorDetailView` and `InspectorDataGridBuilder` now use it instead of their JPA checks — the loader sets a JPQL string only for stores that accept one, and the detail view loads a non-JPA root by id.
- `EntityModel` and `AttributeModel` carry a `dynamic` flag, shown as a grid column only when a contributor is registered. The diagram marks a contributed entity with `<<dynamic>>` on the entity, and a contributed attribute of a static entity on the attribute.
- The reference field of a non-JPA property is no longer forced read-only: the owning-side rule reads JPA mapping annotations, which such a property does not have.

### How it works

`DataModelRegistry` builds the data model from JPA entities as before, then adds the entities contributors offer and routes any property JPA metadata does not describe to them. A contributed column is qualified as `TABLE.COLUMN` when it lives in a table other than the entity's own; an attribute with no column of its own shows no column. The Entity inspector offers every entity of a real data store, so entities of non-JPA stores — dynamic entities, REST DTO stores, custom stores — are browsable and editable without any store-specific code.

### How to use

Implement `DataModelContributor` and register it as a bean to make a module's entities and attributes appear in the Data model view. Nothing is needed for the Entity inspector — an entity of any registered data store is offered automatically.

### Compatibility

Opt-in for the Data model view: with no contributor registered, the data model is built exactly as before and the `Dynamic` columns stay hidden. The Entity inspector's entity list widens by design — entities of non-JPA stores now appear where previously only JPA entities did.

### Test plan

- `./gradlew :datatools:test :datatools-flowui:test`
- Manually: open the Entity inspector on an entity of a non-JPA store — the list shows its fields, not only the id; edit a row and check reference fields are editable.

### DIFF
--- a/jmix-datatools/datatools/src/main/resources/io/jmix/datatools/messages.properties
+++ b/jmix-datatools/datatools/src/main/resources/io/jmix/datatools/messages.properties
@@ -25,12 +25,14 @@ io.jmix.datatools.datamodel.entity/AttributeModel=Attribute model
 io.jmix.datatools.datamodel.entity/AttributeModel.attributeName=Attribute name
 io.jmix.datatools.datamodel.entity/AttributeModel.columnName=Column name
 io.jmix.datatools.datamodel.entity/AttributeModel.dbType=DB type
+io.jmix.datatools.datamodel.entity/AttributeModel.dynamic=Dynamic
 io.jmix.datatools.datamodel.entity/AttributeModel.id=Id
 io.jmix.datatools.datamodel.entity/AttributeModel.isMandatory=Is mandatory
 io.jmix.datatools.datamodel.entity/AttributeModel.javaType=Java type
 
 io.jmix.datatools.datamodel.entity/EntityModel=Entity model
 io.jmix.datatools.datamodel.entity/EntityModel.attributeModels=Attribute models
+io.jmix.datatools.datamodel.entity/EntityModel.dynamic=Dynamic
 io.jmix.datatools.datamodel.entity/EntityModel.id=Id
 io.jmix.datatools.datamodel.entity/EntityModel.name=Entity name
 io.jmix.datatools.datamodel.entity/EntityModel.tableName=Table name
--- a/jmix-datatools/datatools-flowui/src/main/java/io/jmix/datatoolsflowui/view/datamodel/DataModelListView.java
+++ b/jmix-datatools/datatools-flowui/src/main/java/io/jmix/datatoolsflowui/view/datamodel/DataModelListView.java
@@ -75,6 +75,8 @@ public class DataModelListView extends StandardView {
     @ViewComponent
     protected DataGrid<EntityModel> entityModelsDataGrid;
     @ViewComponent
+    protected DataGrid<AttributeModel> attributeModelsDataGrid;
+    @ViewComponent
     protected JmixCheckbox showSystemCheckBox;
     @ViewComponent
     protected TypedTextField<String> entityFilter;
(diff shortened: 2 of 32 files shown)


jmix-datatools/datatools/src/main/resources/io/jmix/datatools/messages.properties
```diff
@@ -25,12 +25,14 @@ io.jmix.datatools.datamodel.entity/AttributeModel=Attribute model
 io.jmix.datatools.datamodel.entity/AttributeModel.attributeName=Attribute name
 io.jmix.datatools.datamodel.entity/AttributeModel.columnName=Column name
 io.jmix.datatools.datamodel.entity/AttributeModel.dbType=DB type
+io.jmix.datatools.datamodel.entity/AttributeModel.dynamic=Dynamic
```
glebfox, line 28: add message keys to the translation module

glebfox (approved):

jmix-datatools/datatools/src/main/resources/io/jmix/datatools/messages.properties
knstvk: done

glebfox (approved):

PR #5675, opened 2026-09-10
Runtime view templates: generate a read-only view for an entity

### LINKED ISSUE #5627: Runtime view templates: generate a read-only view for an entity
### Description

Runtime view templates generate list and detail views for an entity from its metadata, without any source code in the application. There is no read-only counterpart: an application that relies on runtime templates has to hand-write a read view, or keep opening the generated detail view in read-only mode. Read views themselves are introduced in jmix-framework/jmix#3833.

### Use case

Applications that expose entities through runtime templates — admin-style views, entities coming from add-ons, entities whose UI is not worth hand-writing — need a viewing view for the same entities. Their editing views are generated, so the viewing view being the only hand-written one is an odd gap.

### Expected behavior

Runtime view templates should be able to generate a read-only view for an entity, registered under the same id and route conventions as the read views from #3833, so that it is picked up by view resolution, by the read action and by the menu exactly like a hand-written one.

### Impact

Entities served entirely by runtime templates get a viewing view with no source code, and the read-view conventions hold uniformly for generated and hand-written views.

KremnevDmitry:
See also: https://github.com/jmix-framework/jmix-premium/pull/316

# Summary

Read-only views can now be produced at runtime: from a `@ReadViewTemplate` annotation on an entity class, and from a Dynamic Model view declaration with `type: read`. Both paths reuse the read view infrastructure from #3833.

### What was done

**Framework (`flowui`)**

- Added the `@ReadViewTemplate` meta-annotation for entity classes.
- Added the stock `read-view.ftl` descriptor template and the `TemplateReadView` base class.
- `ViewRegistry` now owns the primary read view registry.

**Dynamic Model add-on (`dynmodel`, `dynmodel-flowui`)**

- `DynamicViewType` gained `READ` in both the meta enum and the Flow UI model enum.
- Deployment, validation, menu building, behavior handlers, the security policy store and the entity-views provider all handle the read type.

### How it works

For an annotated entity the framework renders a descriptor from the template, generates a controller extending `TemplateReadView`, and registers it under `<entity>.read` on `<slug-of-view-id>/:id/read` as the entity's primary read view. 

Read-only state comes from `ReadOnlyViewsSupport`, so fields are read-only – including those in tabs opened later – and modifying list actions are disabled.

For Dynamic Model, a dynamic entity's read declaration goes through the same template.

### How to use

Annotate the entity class:

```java
@ReadViewTemplate
@JmixEntity
@Table(name = "CUSTOMER")
@Entity
public class Customer { … }
```

Or declare it in the model – in the settings view (Views → Create → Type: Read), in YAML, or through the API:

```yaml
views:
  - type: "read"
    viewId: "Customer.read"
    viewRoute: "shown-customers"      # /:id/read is appended
    readEntityContainerId: "shownDc"  # needs a descriptor that declares it
```

```java
DynamicModelDefinitions.readView()
        .viewId("Customer.read")
        .readEntityContainerId("shownDc");
```

`readEntityContainerId` only takes effect together with a custom `descriptor` / `path` that declares the container under that id – the stock template declares the default one, exactly as with the detail template's edited-container id.

### Breaking changes

- `DynamicViewDefinition` is a record and its canonical constructor gained the `readEntityContainerId` component before `handlers`. Code constructing it positionally must be updated.
- Both `DynamicViewType` enums (`io.jmix.dynmodel.meta` and `io.jmix.dynmodelflowui.model`) gained a `READ` constant. Exhaustive `switch` expressions over them in add-on or application code stop compiling until the new branch is added.


jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/template/ReadViewTemplate.java
```diff
@@ -0,0 +1,77 @@
...
+ * The generated view shows the entity and cannot edit it. It is registered as the entity's primary read
+ * view, so read view resolution finds it: the read view navigator, a {@code list_read} action and an
+ * {@code entity_read} action all open it. Its route requires an entity id, so the view always opens for an
+ * existing instance. Note that the generated list view carries no {@code list_read} action of its own.
+ */
+@Experimental
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@MetaAnnotation
+public @interface ReadViewTemplate {
```
knstvk, line 39: `@ReadViewTemplate` should support `parentMenu()` the same as `@DetailViewTemplate`. The entity ID can be assigned in a route parameter, see for example https://docs.jmix.io/jmix/flow-ui/menu-config.html#route-parameters

knstvk (changes requested):

jmix-flowui/flowui/src/main/java/io/jmix/flowui/view/template/ReadViewTemplate.java
KremnevDmitry: It seems that, in this case, supporting `parentMenu()` doesn't make sense. The reason is that we can only pass the entity ID for display via route parameters, and defining that is the responsibility of `menu.xml`. 

Consequently, the order is also determined by the entry in `menu.xml`.

If we were to allow selecting `parentMenu()` here – within the annotation itself, we would need to implement a mechanism to pass the ID route parameter directly using that annotation.

That doesn't seem like the optimal approach to me. What do you think?

knstvk (approved):
You are right, `parentMenu()` doesn't make sense in ReadViewTemplate

glebfox (approved):

